PR #11, opened 2018-11-20
add support for timeouts on instance, image, and volume resources

displague:
It is not possible to override `timeouts` on a resource, unless that resource supports `timeouts`.

This PR adds some default timeout values to resources that interact with disks, node, and job scheduling which can take longer periods of time based on size and demand.

* `resource/linode_instance`: Add `timeouts` support for `create`, `update`, and `delete` (defaults 10, 20, 10)
* `resource/linode_image`: Add `timeouts` support for `create` (defaults 20)
* `resource/linode_volume`: Add `timeouts` support for `create`, `update`, and `delete` (defaults 10, 20, 10)

https://www.terraform.io/docs/configuration/resources.html#timeouts

```hcl
    ...
    timeouts {
        create = "10m"
        update = "20m"
        delete = "10m"
    }
}

resource "linode_image" "foo" {
    ...
    timeouts {
        create = "10m"
    }
}

resource "linode_volume" "foo" {
    ...
    timeouts {
        create = "10m"
        update = "20m"
        delete = "10m"
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
 ## 1.2.1 (Unreleased)
+
+ENHANCEMENTS:
+
+* resource/linode\_instance: Add `timeouts` support for `create`, `update`, and `delete` (defaults 10, 20, 10)
+* resource/linode\_image: Add `timeouts` support for `create` (defaults 20)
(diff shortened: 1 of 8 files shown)


displague:
✅ 52 tests passed

PR #14, opened 2018-12-03
add support for tags in domains, nodebalancers, and volume resources

displague:
TODO
-------

* [x] Fix diff detection when tags are returned in a different order #16 
* [x] Fix suspected linode delete + volume attach + volume update race #15 (introduced by this branch)

Enhancements
---------------

* resource/linode\_domain: Add `tags` field
* resource/linode\_nodebalancer: Add `tags` field
* resource/linode\_volume: Add `tags` field


displague:
Currently blocked by #15

displague:
Team city builds passing

image

PR #22, opened 2019-01-01
Fix arguments in nodebalancer_node docs

gerhard:
* region argument does not exist
* label is optional
* nodebalancer_id & config_id are required

I believe that the list of exported attributes isn't quiet right, but I was focusing on the incorrect parts.

### DIFF
--- a/website/docs/r/nodebalancer_node.html.md
+++ b/website/docs/r/nodebalancer_node.html.md
@@ -59,12 +59,15 @@ resource "linode_nodebalancer_node" "foonode" {
 
 The following arguments are supported:
 
-* `label` - (Required) The label of the Linode NodeBalancer Node. This is for display purposes only.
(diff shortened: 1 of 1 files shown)


displague:
Thanks again, @gerhard!

https://developers.linode.com/api/v4#operation/createNodeBalancerNode

As your PR suggests, region should not be there, and the two supporting IDs definitely should.  I imagine I may have missed documentation for other required fields that are referenced in the URL path.

Label is documented as required, did you find this not to be the case (it could be a docs bug)?

displague:
I see.  That is incorrect.
```
$ linode-cli nodebalancers node-create 41123 36123 --address="192.168.128.123:8080"
Request failed: 400
┌errors─┬───────────────────┐
│ field │ reason            │
├───────┼───────────────────┤
│ label │ label is required │
└───────┴───────────────────┘
```

gerhard:
- [x] link in previous comment fixed
- [x] label change reverted

PR #25, opened 2019-01-14
[AUTOMATED] Upgrade to Go 1.11

appilon:


appilon:
travis failuires are from `govendor` project seems to be using `dep` though

PR #26, opened 2019-01-26
[NEW] Add a domain data source

nZac:
This add a new domain data source to access details about a domain from the API. 

This is my first foray into Go, so I might have messed up the [integer conversions](https://github.com/terraform-providers/terraform-provider-linode/pull/26/files#diff-7c8f040ebdf5de9108e879df8e1e0ae5R118).

Also, this is almost a direct copy-paste from the `data.linode_instance` source... I think I got all the references to "image" but you might want to make sure I didn't mess anything else up.

If this is satisfactory, but needs a small tweak, by all means, feel free to add commits.

### DIFF
--- a/examples/main.tf
+++ b/examples/main.tf
@@ -16,6 +16,10 @@ data "linode_image" "ubuntu" {
   id = "linode/ubuntu18.04"
 }
 
+data "linode_domain" "linode_com" {
+  id = "1234566"
+}
+
 resource "linode_nodebalancer" "foo-nb" {
   label                = "${random_pet.project.id}"
   region               = "${data.linode_region.main.id}"
--- a/linode/data_source_linode_domain.go
+++ b/linode/data_source_linode_domain.go
@@ -1,0 +1,138 @@
+package linode
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/linode/linodego"
+)
+
+func dataSourceLinodeDomain() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceLinodeDomainRead,
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"domain": {
+				Type:        schema.TypeString,
+				Description: "The domain this Domain represents. These must be unique in our system; you cannot have two Domains representing the same domain.",
+				Computed:    true,
+			},
+			"type": {
+				Type:        schema.TypeString,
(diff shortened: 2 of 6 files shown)


examples/main.tf
```diff
@@ -16,6 +16,10 @@ data "linode_image" "ubuntu" {
   id = "linode/ubuntu18.04"
 }
 
+data "linode_domain" "linode_com" {
+  id = "1234566"
```
displague: `id` could come in handy too, but I think the most valuable lookup key will be the `domain` field.  

Would this simplify your use case?

displague:
`website/linode.erb` will also need an entry for this data source.

examples/main.tf
nZac: I would love to be able to use either.

However, while reading in [linodego](https://github.com/linode/linodego/blob/master/domains.go#L219) and the [apidocs](https://developers.linode.com/api/v4#operation/getDomain), I couldn't find a clean way to use `linode.com` or something similar to lookup the domain without getting a list and then doing a string match or something.

nZac:
@displague thanks for the review, happy to contribute back.

I see what you mean about the domain record. If I can get an hour this week to add that I would, it was surprising how easy it was to create a data source.  

An issue I found and submitted an issue (https://github.com/linode/linodego/issues/69) is for 429 responses when you get a lot of resources. This data resource made it especially worse because I used it in a lot of submodules. 

Is it possible to cache these?

examples/main.tf
displague: https://developers.linode.com/api/v4#operation/getDomains

`domain` and `type` are filterable. https://github.com/linode/linodego#filtering

using `client.ListDomains(...)` will result in a filtered list of domains matching that domain.  It would still be good to use a string comparison against the domains that are returned (there _should_ be 0 or 1) and verify that we got what we expected before using it.

displague: Here's a similar snippet from the Linode CSI driver: https://github.com/linode/linode-blockstorage-csi-driver/blob/e312ce00cf75ed243c4e545d73403c395ff1489d/pkg/linode-bs/controllerserver.go#L78-L106. 

In this case, it is only checking that exactly 1 result was returned.

displague: Linode enforces `domain` uniqueness, so we really should only ever get 0 or 1 returned, I just don't trust computers. 😃

displague:
Thanks, @nZac!

This hasn't been released yet, but it has been merged into master.  I added a few changes in https://github.com/terraform-providers/terraform-provider-linode/commit/3b35d7e2983778338b30657ac1dbe2c659ea0cb8 to support for `domain` lookups (in addition to the `id` lookups).

I'd like to include `domain_record` support for the `1.5.0` release.

PR #34, opened 2019-03-09
Add missing required argument to example

synical:
Started using this provider today and noticed that the domain resource example in the docs didn't work as it's missing the required `type` field.

Added this to the example with the value `master`.


displague:
Thanks, @synical!  I've encountered this myself and should have updated the docs sooner.

Documentation: code that is rarely tested.

PR #41, opened 2019-04-09
Accept "0" in Domain and Domain Record time-in-seconds related fields

displague:
Accept 0 in domain and domain_record ttl_sec, retry_sec, expire_sec, refresh_sec fields

The API permits and returns `0` for these fields even though the documented behavior suggests that this will be treated as `300`s.

Closes https://github.com/terraform-providers/terraform-provider-linode/issues/35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ ENHANCEMENTS:
 BUG FIXES:
 
 * Documentation and examples for `linode_domain` resource were missing required `type` field
+* `linode_domain_record` field `ttl_sec` accepts `0`, as the API does (#35)
+* `linode_domain` fields `ttl_sec`, `retry_sec`, `expire_sec`, and `refresh_sec` now accept `0`, as the API does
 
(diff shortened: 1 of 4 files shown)


ctreatma:
What is the TTL on the resulting DNS record in these cases?

- If I don't specify a TTL, the API returns `0` as the default, but per documentation this will be treated as `300`
- If I explicitly specify a TTL of `0`, the API returns `0`, but this will be treated as `<something>`

In each of these cases, if I `dig <hostname>`, what TTL will I see?

displague:
@ctreatma I verified that with a value of `0`, you will see `300` in your DNS query responses.

While this is arguably an upstream issue, the validation that denied `0` prevented users from importing existing domains and domain records.

ctreatma:
I agree we shouldn't mark things as invalid if the API will accept them.  We may want to document this behavior in the provider, so that terraform users understand that specifying a value of `0` for the TTL in the resource is valid but does not actually set a TTL of `0` on the resulting DNS record.

PR #53, opened 2019-06-17
resize disks only against powered-off instances.

ellisbenjamin:
Since disks can only be resized on powered-off Linodes, we need to issue a shutdown job prior to resizing. This issues a shutdown job, waits for the InstanceStatus to be "Offline," then resizes the disk and boots the instance.

### DIFF
--- a/linode/linode_instance_helpers.go
+++ b/linode/linode_instance_helpers.go
@@ -724,7 +724,9 @@ func changeInstanceDiskSize(client *linodego.Client, instance linodego.Instance,
 		if _, err := client.WaitForInstanceStatus(context.Background(), instance.ID, linodego.InstanceOffline, int(d.Timeout(schema.TimeoutUpdate).Seconds())); err != nil {
 			return fmt.Errorf("Error waiting for instance %d to go offline: %s", instance.ID, err)
 		} else {
-			client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize)
+			if err := client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize); err != nil {
+				return fmt.Errorf("Error resizing Disk %d for Instance %d: %s", disk.ID, instance.ID, err)
+			}
 		}
 
 		// Wait for the Disk Resize Operation to Complete, and boot instance.
--- a/linode/linode_instance_helpers.go
+++ b/linode/linode_instance_helpers.go
@@ -717,10 +717,17 @@ func changeInstanceType(client *linodego.Client, instance *linodego.Instance, ta
 
 func changeInstanceDiskSize(client *linodego.Client, instance linodego.Instance, disk linodego.InstanceDisk, targetSize int, d *schema.ResourceData) error {
 	if instance.Specs.Disk > targetSize {
-		client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize)
+		// This is idempotent, so we don't need to check the status of the instance.
+		client.ShutdownInstance(context.Background(), instance.ID)
 
-		// Wait for the Disk Resize Operation to Complete
-		// waitForEventComplete(client, instance.ID, "linode_resize", waitMinutes)
+		// Wait for instance to go offline. Resize the disk once Linode is shut down.
+		if _, err := client.WaitForInstanceStatus(context.Background(), instance.ID, linodego.InstanceOffline, int(d.Timeout(schema.TimeoutUpdate).Seconds())); err != nil {
+			return fmt.Errorf("Error waiting for instance %d to go offline: %s", instance.ID, err)
+		} else {
+			client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize)
+		}
+
+		// Wait for the Disk Resize Operation to Complete, and boot instance.
 		_, err := client.WaitForEventFinished(context.Background(), instance.ID, linodego.EntityLinode, linodego.ActionDiskResize, disk.Updated, int(d.Timeout(schema.TimeoutUpdate).Seconds()))
 		if err != nil {
 			return fmt.Errorf("Error waiting for resize of Instance %d Disk %d: %s", instance.ID, disk.ID, err)
--- a/linode/resource_linode_instance_test.go
+++ b/linode/resource_linode_instance_test.go
@@ -933,8 +933,6 @@ func TestAccLinodeInstance_diskResizeAndExpanded(t *testing.T) {
 	})
 }
 
-
-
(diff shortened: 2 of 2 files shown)


linode/linode_instance_helpers.go
```diff
@@ -717,10 +717,17 @@ func changeInstanceType(client *linodego.Client, instance *linodego.Instance, ta
...
+		// This is idempotent, so we don't need to check the status of the instance.
+		client.ShutdownInstance(context.Background(), instance.ID)
 
-		// Wait for the Disk Resize Operation to Complete
-		// waitForEventComplete(client, instance.ID, "linode_resize", waitMinutes)
+		// Wait for instance to go offline. Resize the disk once Linode is shut down.
+		if _, err := client.WaitForInstanceStatus(context.Background(), instance.ID, linodego.InstanceOffline, int(d.Timeout(schema.TimeoutUpdate).Seconds())); err != nil {
+			return fmt.Errorf("Error waiting for instance %d to go offline: %s", instance.ID, err)
+		} else {
+			client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize)
```
displague: The potential error here should be captured and this should return early in that case.  Not catching this error is what led to this issue.

linode/linode_instance_helpers.go
```diff
@@ -717,10 +717,17 @@ func changeInstanceType(client *linodego.Client, instance *linodego.Instance, ta
 
 func changeInstanceDiskSize(client *linodego.Client, instance linodego.Instance, disk linodego.InstanceDisk, targetSize int, d *schema.ResourceData) error {
 	if instance.Specs.Disk > targetSize {
-		client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize)
+		// This is idempotent, so we don't need to check the status of the instance.
+		client.ShutdownInstance(context.Background(), instance.ID)
```
displague: Resizing a disk that is not connected to the loaded config shouldn't require a shutdown.  Does the API distinguish between the two states or does it always return an error?

displague: This was left in the air back in https://github.com/displague/deprecated-terraform-provider-linode/pull/23#issuecomment-423197413 :-(

> Additional tests may be needed to verify that a resize on a live / running disk results in a shutdown and before the resize and a reboot afterwards.

ellisbenjamin:
Thanks a ton, @displague -- your insights are always appreciated. I'll update this per your comments and see if I can include the necessary changes to address https://github.com/terraform-providers/terraform-provider-linode/issues/13#issuecomment-443192604 as well

linode/linode_instance_helpers.go
ellisbenjamin: > Resizing a disk that is not connected to the loaded config shouldn't require a shutdown. Does the API distinguish between the two states or does it always return an error?

The API does not distinguish between disks that are attached / detached to the running instance via a config and will not permit resizing of an unattached disk associated with a running Linode.

linode/linode_instance_helpers.go
ellisbenjamin: I was mistaken in my previous understanding here.. Calling the disk resize endpoint against a running Linode returns an empty JSON object & HTTP 200, the same as it would for a successful call:
```
$  curl -iH "Content-Type: application/json"     -H "Authorization: Bearer $TOKEN"     -X POST -d '{
      "size": 222
    }'     https://api.linode.com/v4/linode/instances/14510321/disks/30138982/resize
HTTP/1.1 200 OK
Server: nginx
Date: Tue, 18 Jun 2019 15:11:17 GMT
Content-Type: application/json
Content-Length: 2
Connection: keep-alive
X-OAuth-Scopes: *
X-Accepted-OAuth-Scopes: linodes:read_write
X-Frame-Options: DENY
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: HEAD, GET, OPTIONS, POST, PUT, DELETE
Access-Control-Allow-Headers: Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Filter
X-Spec-Version: 4.0.24
X-RateLimit-Limit: 400
X-RateLimit-Remaining: 399
X-RateLimit-Reset: 1560870797
Retry-After: 119
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Status
Cache-Control: private, max-age=60, s-maxage=60
Content-Security-Policy: default-src 'none'
Vary: Authorization, X-Filter
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block
Strict-Transport-Security: max-age=31536000

{}
```

 An event is then created, which quickly comes back with a failed status, but does not provide any additional information about the failure:

```
{
  "page": 1,
  "data": [
    {
      "created": "2019-06-18T15:11:18",
      "id": 32006501,
      "rate": null,
      "seen": false,
      "status": "failed",
      "time_remaining": null,
      "username": "bellis",
      "entity": {
        "id": 14510321,
        "label": "example2",
        "url": "/v4/linode/instances/14510321",
        "type": "linode"
      },
      "action": "disk_resize",
      "percent_complete": 100,
      "read": false
    }
  ],
  "results": 1,
  "pages": 1
}
```

This endpoint does produce errors, eg, if you ask it to resize beyond the Linode's specifications, so I agree that we should still catch errors here:

```
$  curl -iH "Content-Type: application/json"     -H "Authorization: Bearer $TOKEN"     -X POST -d '{
      "size": 2222222222222222222
    }'     https://api.linode.com/v4/linode/instances/14510321/disks/30138982/resize
HTTP/1.1 400 BAD REQUEST
Server: nginx
Date: Tue, 18 Jun 2019 15:14:38 GMT
Content-Type: application/json
Content-Length: 67
Connection: keep-alive
X-OAuth-Scopes: *
X-Accepted-OAuth-Scopes: linodes:read_write
X-Frame-Options: DENY
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: HEAD, GET, OPTIONS, POST, PUT, DELETE
Access-Control-Allow-Headers: Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Filter
X-Spec-Version: 4.0.24
X-RateLimit-Limit: 400
X-RateLimit-Remaining: 399
X-RateLimit-Reset: 1560870998
Retry-After: 119

{"errors": [{"reason": "Insufficient space for resize operation"}]}
```

I'll add the appropriate code

ellisbenjamin:
This was resolved in PR 55--

https://github.com/terraform-providers/terraform-provider-linode/blob/master/linode/linode_instance_helpers.go#L704

PR #55, opened 2019-06-26
v0.12-updates

ellisbenjamin:
This PR includes necessary changes for the Linode Provider to work with TF v0.12.0+


mdawar:
Great, I hope this pull request gets merged soon, I really need this plugin to work with 0.12, downgrading is really not an option.

linode/resource_linode_instance.go
```diff
@@ -1160,16 +1160,44 @@ func resourceLinodeInstanceUpdate(d *schema.ResourceData, meta interface{}) erro
...
-		}
-		d.Set("type", d.Get("type").(string))
-	}
+	var rebootInstance bool
+	var diskIDLabelMap map[string]int
 
 	tfDisksOld, tfDisksNew := d.GetChange("disk")
 
-	rebootInstance, diskIDLabelMap, err := updateInstanceDisks(client, d, *instance, tfDisksOld, tfDisksNew)
+	tfDisksOldInterface := tfDisksOld.([]interface{})
```
jnschaeffer: This adds complexity to a function that is already very large; is it possible to put this into its own function?

linode/resource_linode_instance.go
```diff
@@ -1160,16 +1160,44 @@ func resourceLinodeInstanceUpdate(d *schema.ResourceData, meta interface{}) erro
...
+	for _, disk := range tfDisksNewInterface {
+		newDiskSize += disk.(map[string]interface{})["size"].(int)
+	}
+
+	if newDiskSize > oldDiskSize {
+		if d.HasChange("type") {
+			if err = changeInstanceType(&client, instance, d.Get("type").(string), d); err != nil {
+				return err
+			}
+			d.Set("type", d.Get("type").(string))
```
jnschaeffer: Rather than call `d.Get("type").(string)` twice, why not call it once and store the value?

linode/resource_linode_instance.go
```diff
@@ -1160,16 +1160,44 @@ func resourceLinodeInstanceUpdate(d *schema.ResourceData, meta interface{}) erro
...
+		}
+		rebootInstance, diskIDLabelMap, err = updateInstanceDisks(client, d, *instance, tfDisksOld, tfDisksNew)
+	} else {
+		rebootInstance, diskIDLabelMap, err = updateInstanceDisks(client, d, *instance, tfDisksOld, tfDisksNew)
```
jnschaeffer: The error returned from `updateInstanceDisks` isn't checked here. If the disk update operation fails but the instance type change operation succeeds, this operation will look like it succeeded when in reality it failed.

linode/resource_linode_instance.go
```diff
@@ -1160,16 +1160,44 @@ func resourceLinodeInstanceUpdate(d *schema.ResourceData, meta interface{}) erro
...
+	// Get total amount of disk usage before & after
+	for _, disk := range tfDisksOldInterface {
+		oldDiskSize += disk.(map[string]interface{})["size"].(int)
+	}
+
+	for _, disk := range tfDisksNewInterface {
+		newDiskSize += disk.(map[string]interface{})["size"].(int)
+	}
+
+	if newDiskSize > oldDiskSize {
```
jnschaeffer: Why is the order of steps dependent on the size of the two disks? Is there any reason we can't just choose one of these orderings?

linode/linode_instance_helpers.go
```diff
@@ -702,25 +712,53 @@ func changeInstanceType(client *linodego.Client, instance *linodego.Instance, ta
...
+			return fmt.Errorf("Error waiting for instance %d to enter online state: %s", instance.ID, err)
+		}
+	}
 	return nil
 }
 
 func changeInstanceDiskSize(client *linodego.Client, instance linodego.Instance, disk linodego.InstanceDisk, targetSize int, d *schema.ResourceData) error {
 	if instance.Specs.Disk > targetSize {
-		client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize)
+		if instance.Status == linodego.InstanceOffline || instance.Status == linodego.InstanceShuttingDown {
```
jnschaeffer: This checks to see if `instance.Status == linodego.InstanceOffline`, but the next line waits for the status to become `linodego.InstanceOffline`. That doesn't seem right; maybe make this a `switch` like so:
```go
switch instance.Status {
case linodego.InstanceShuttingDown:
	if _, err := client.WaitForInstanceStatus(context.Background(), instance.ID, linodego.InstanceOffline, int(d.Timeout(schema.TimeoutUpdate).Seconds())); err != nil {
		return fmt.Errorf("Error waiting for instance %d to go offline: %s", instance.ID, err)
	}
case linodego.InstanceOffline:
	continue
default:
	if errShutdown := client.ShutdownInstance(context.Background(), instance.ID); errReturn != nil {
		return errShutdown
	}
}
```

linode/linode_instance_helpers.go
```diff
@@ -702,25 +712,53 @@ func changeInstanceType(client *linodego.Client, instance *linodego.Instance, ta
...
+		} else {
+			client.ShutdownInstance(context.Background(), instance.ID)
+		}
+
+		// Wait for instance to go offline. Resize the disk once Linode is shut down.
+		if _, err := client.WaitForInstanceStatus(context.Background(), instance.ID, linodego.InstanceOffline, int(d.Timeout(schema.TimeoutUpdate).Seconds())); err != nil {
+			return fmt.Errorf("Error waiting for instance %d to go offline: %s", instance.ID, err)
+		} else {
+			if err := client.ResizeInstanceDisk(context.Background(), instance.ID, disk.ID, targetSize); err != nil {
+				return fmt.Errorf("Error resizing Disk %d for Instance %d: %s", disk.ID, instance.ID, err)
```
jnschaeffer: Are `Disk` and `Instance` proper nouns?

CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
...
+
+* Compatible with Terraform v0.12.0+
+* Uses linodego v0.10.0
+* Examples updated with new TF config syntax
+
+BUG FIXES:
+
+* The Linode API resizes disks by default when an instance is resized. This behavior is now accounted for -- Terraform will not resize disks unless a new size is specified in the config.
+* The provider now waits for instance resizing to complete before attempting to issue disk resize jobs against an instance. This was required because new jobs issued to actively-resizing instances fail.
+* Disk resizing and instance resizing is now executed in the correct order.
```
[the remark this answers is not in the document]
ellisbenjamin: √ fixed.

linode/linode_instance_helpers.go
ellisbenjamin: Looks like we typically tend to treat Instance as a proper noun. Less so for disk. I'm going to continue treating Instance as a proper noun.

linode/resource_linode_instance.go
ellisbenjamin: good call. adding this.

linode/resource_linode_instance.go
ellisbenjamin: imagining the following change:
1 GB Linode with 25 GB SSD -> 2 GB Linode with 40 GB SSD
We need to change the instance size here first, and then the disk size, or else the disk will be too large for the instance.

Going the other way:
2 GB instance with 50 GB disk -> 1 GB instance with 20 GB disk
We need to shrink the disk first, or else we will not be able to get all 50 GB on the smaller linode instance type.

another case:
2 GB w/ 10 GB ssd -> 1 GB with 25 GB SSD:
order does not matter

2 GB w/ 25 -> 1 GB with 50 GB:
produce error

.. is that sensible?

linode/linode_instance_helpers.go
ellisbenjamin: Doing this.

linode/resource_linode_instance.go
ellisbenjamin: √ doing this.

linode/resource_linode_instance.go
ellisbenjamin: Yes, 💯 . I'm adding getDiskSizeChange to linode_instance_helpers to deal with this.

linode/linode_instance_helpers.go
```diff
@@ -702,25 +712,53 @@ func changeInstanceType(client *linodego.Client, instance *linodego.Instance, ta
...
-	if err := client.ResizeInstance(context.Background(), instance.ID, targetType); err != nil {
+	// Issue the resize job
+	if err := client.ResizeInstance(context.Background(), instance.ID, resizeOpts); err != nil {
 		return fmt.Errorf("Error resizing instance %d: %s", instance.ID, err)
 	}
 
-	_, err := client.WaitForEventFinished(context.Background(), instance.ID, linodego.EntityLinode, linodego.ActionLinodeResize, *instance.Created, int(d.Timeout(schema.TimeoutUpdate).Seconds()))
-	if err != nil {
-		return fmt.Errorf("Error waiting for instance %d to finish resizing: %s", instance.ID, err)
+	// wait for instance to begin resizing after issuing resize job
```
displague: Waiting for a transitional status is risky because the event may have already been transitioned out by the time the code gets here, or it may have transitioned from pre-resizing to post-resizing in between polling loops.  This is why EventForEventFinished was used.

linode/linode_instance_helpers.go
```diff
@@ -692,8 +692,18 @@ func createRandomRootPassword() (string, error) {
...
+
+	diskResize := false
+	waitForOnline := true
+
+	resizeOpts := linodego.InstanceResizeOptions{
+		AllowAutoDiskResize: &diskResize,
+		Type:                targetType,
+	}
+
 	// Instance must be either offline or running (with no extra activity) to resize.
```
displague: Is it still possible (in the Linode API) to resize an instance while it is online or does it always have to be shutdown first?

If resizing requires an offline instance, it may be saner to always send a shutdown and await that status before sending a ResizeInstance (simplifying this code).

ellisbenjamin: We can issue a shutdown job to a Running instance... the resize job will shutdown the instance if necessary, and after the resize is complete, the Linode is returned to its original state (either Offline or Running).

If we always issued a shutdown job, we would still need to check the original status of the Linode (which we already do) and then return it to that state manually (which i think would add back the additional complexity we would save by always shutting down the instance)

linode/linode_instance_helpers.go
ellisbenjamin: I see.. 

I swapped out WaitForEventFinished in favor of WaitForInstanceStatus because there are situations -- namely when resizing a disk after resizing an instance -- where the instance resize event will have completed, but the issuing additional jobs before the Linode has been returned to its original status results in the API returning an error because the Linode is still "busy"..

I understand your concern about waiting for the transitional "InstanceResizing" status here. If we were to nix waiting for the transitional "resize" status, though, and only wait for "Running" or "Offline," then there are situations where this will simply move ahead while the instance is in state "Running" or "Offline", while waiting for its turn in the migration queue. I'm not sure that I see a good alternative but I welcome suggestions.

ellisbenjamin: I nixed the code where it waits for the InstanceResizing status. Like you suggested, I'm using WaitForEventFinished, and then waiting for the instance to return to its original state. This seems to work and I agree that it will avoid the possibility of missing the transitional status.

PR #63, opened 2019-11-03
Add persist_across_boots argument for the linode_volume resource

aqche:
I was taking a look at the `VolumeCreateOptions` struct in the Linode Go Client and noticed there was the `PersistAcrossBoots` option so I took a crack at adding the `persist_across_boots` argument for the `linode_volume` resource. Hopefully it is helpful!


aqche:
Revisiting this, looks like since the `PersistAcrossBoots` value is not available in the `Volume` struct so there is no way to set the value of `persist_across_boots` when importing a `linode_volume`.  Closing this PR for now.

PR #71, opened 2019-11-27
Update docs with info about rate limiting

golightlyb:
Add a section:

## Rate Limiting

The Linode API may apply rate limiting when you update the state for a large inventory:

```
Error: Error getting Linode DomainRecord ID 123456: [002] unexpected end of JSON input



Error: Error finding the specified Linode DomainRecord: [002] unexpected end of JSON input
```

If this affects you, run Terraform with [--parallelism=1](/docs/commands/apply.html#parallelism-n)


phillc:
Thank you!

PR #74, opened 2019-12-06
TPT-57 Volume data source

Jskobos:
**WIP**

### DIFF
--- a/website/docs/d/volume.html.md
+++ b/website/docs/d/volume.html.md
@@ -42,7 +42,7 @@ The Linode Volume resource exports the following attributes:
 
 - `size` - The Volume's size, in GiB.
 
-- `region` - The region in which this Volume is located.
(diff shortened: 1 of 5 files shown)


website/docs/d/volume.html.md
```diff
@@ -0,0 +1,51 @@
+---
```
displague: You'll also want to update https://github.com/terraform-providers/terraform-provider-linode/blob/master/website/linode.erb

Jskobos: Thanks Marques, I did this but can't find instructions on how to test the generated html file.

phillc: It's this I believe https://github.com/terraform-providers/terraform-provider-linode/blob/master/GNUmakefile#L52

PR #79, opened 2020-01-29
WIP: LKE support

sgmac:
Still, in progress, I have a few tasks ongoing

*  ~~Add tests for resource `linode_lke_cluster`~~
* Define a way to determine when the cluster is created, @jnschaeffer gave me a few ideas. For this, I lean to poll the API server endpoint until I get 403. However, this doesn't guarantee the nodes will be in status `ready`. If I go this way maybe is worth to do add a helper to `linodego` ?

Suggestions are welcome.

**UPDATE** 

**1.) linodego Volume**

I commented out just to be able to build the provider, we should fix this first and I could rebase later.
```
linode/data_source_linode_volume.go:95:26: volume.CreatedStr undefined (type *linodego.Volume has no field or method CreatedStr)
linode/data_source_linode_volume.go:96:26: volume.UpdatedStr undefined (type *linodego.Volume has no field or method UpdatedStr)
```
Apparently the `linodego` version in `master` removed  this from v0.12.2

```go
type Volume struct {
    CreatedStr string `json:"created"`
    CreatedStr string `json:"updated"`
...
}
```

**2.) Updated vendor**

**3.) Wait for LKE cluster ready**
I'm currently using ` client.WaitForLKEClusterStatus `, this returns almost immediately, however, the underlying linode instances may not be ready.

**4.) Pending Docs** 
I have to work on the website documentation.

Also worth to mention this is not implement, and it could be useful. https://github.com/linode/linodego/blob/master/lke_clusters.go#L83


linode/provider.go
```diff
@@ -57,6 +57,7 @@ func Provider() terraform.ResourceProvider {
 			"linode_sshkey":                 dataSourceLinodeSSHKey(),
 			"linode_user":                   dataSourceLinodeUser(),
 			"linode_volume":                 dataSourceLinodeVolume(),
+			"linode_lke":                    dataSourceLinodeLKE(),
```
jnschaeffer: This looks great so far! Just one thought at first glance: LKE includes clusters, kubeconfigs, API server endpoints, and node pools. It might make more sense to call this resource `linode_lke_cluster`, the same way `linode_nodebalancer_config` is used for NodeBalancers.

sgmac: Thanks! It's a good point, I was debating myself the naming, I'll make that change and update the PR.

phillc:
Hiya,

Just wanted to let you know that I really appreciate this PR. I am a little hesitant to merge right now, as I am still waiting to see how status polling matures for WaitForLKEClusterStatus. Once I have some clarity, I will help get this over the line.

sgmac:
Thanks, let's see how the LKE Cluster status ends and from there we can see how to add to the terraform provider. I'm glad to help :)

sgmac:
@phillc Any update on this? I would like to polish this PR and work on the feature if that's okay.

Charliekenney23:
Hey @sgmac! Thanks for your early work on this. We really appreciate it!! 😄 

 Once we decided internally on acceptance criteria for the resulting state of creating a `linode_lke_cluster` resource (from the conversation in linode/linodego), we were able to implement a [wait strategy](https://github.com/linode/linodego/pull/139) for the cluster (credited you with co-author).

So the resource was implemented [here](https://github.com/terraform-providers/terraform-provider-linode/pull/143) a bit after LKE reached GA and has been available as part of [v1.10](https://github.com/terraform-providers/terraform-provider-linode/blob/master/CHANGELOG.md#1100-may-05-2020).

PR #82, opened 2020-02-10
make obj secret key accessible

### LINKED ISSUE #78: cannot access the value of the object storage key
### Terraform Version
Terraform v0.12.20

### Affected Resource(s)
linode_object_storage_key

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
provider "linode"{
  token = var.linode_token
}

variable "linode_token" {
  default = ""
}

resource "linode_object_storage_key" "storage_keysxx" {
    label = "staging access - created by terraform"
}


output "storage_keys" {
    value = ["${linode_object_storage_key.storage_keysxx.access_key}","${linode_object_storage_key.storage_keysxx.secret_key}"]
}

```

### Debug Output
https://gist.github.com/pilvikala/3bcb3836256677166484cded7b0039eb

### Panic Output
N/A

### Expected Behavior
The output prints the secret key and the secret key is in the state file. This way the secret key can be pushed to another service that can use it.

### Actual Behavior
Terraform prints [REDACTED] instead of the secret key and the same value is in the state file.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
N/A

### References
N/A

Charliekenney23:
Closes #78

This will allow a given `linode_object_sotrage_key.secret_key` to be stored in tfstate on initial creation. The `secret_key` will never be refetched in subsequent read ops. So as long as the initial tfstate is preserved, the key is accessible.

Because at an API level we restrict refetching the `secret_key`, the importer for this resource has been removed. It's worthless to import as it cannot be used to provision access to buckets for customers.

### DIFF
--- a/linode/resource_linode_object_storage_key.go
+++ b/linode/resource_linode_object_storage_key.go
@@ -15,9 +15,6 @@ func resourceLinodeObjectStorageKey() *schema.Resource {
 		Read:   resourceLinodeObjectStorageKeyRead,
 		Update: resourceLinodeObjectStorageKeyUpdate,
 		Delete: resourceLinodeObjectStorageKeyDelete,
-		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
-		},
 
 		Schema: map[string]*schema.Schema{
 			"label": {
@@ -57,6 +54,9 @@ func resourceLinodeObjectStorageKeyCreate(d *schema.ResourceData, meta interface
 	d.Set("label", objectStorageKey.Label)
 	d.Set("access_key", objectStorageKey.AccessKey)
 
+	// secret_key only available on creation
+	d.Set("secret_key", objectStorageKey.SecretKey)
+
 	return resourceLinodeObjectStorageKeyRead(d, meta)
 }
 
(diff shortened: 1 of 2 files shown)


linode/resource_linode_object_storage_key.go
```diff
@@ -15,9 +15,6 @@ func resourceLinodeObjectStorageKey() *schema.Resource {
 		Read:   resourceLinodeObjectStorageKeyRead,
 		Update: resourceLinodeObjectStorageKeyUpdate,
 		Delete: resourceLinodeObjectStorageKeyDelete,
-		Importer: &schema.ResourceImporter{
```
Charliekenney23: No longer useful to import these keys.

linode/resource_linode_object_storage_key.go
```diff
@@ -75,8 +75,6 @@ func resourceLinodeObjectStorageKeyRead(d *schema.ResourceData, meta interface{}
 
 	d.Set("label", objectStorageKey.Label)
 	d.Set("access_key", objectStorageKey.AccessKey)
-	d.Set("secret_key", objectStorageKey.SecretKey)
```
Charliekenney23: We don't want to overwrite the actual key with `[REDACTED]` as returned by the API for GET requests.

PR #85, opened 2020-02-17
add linode_domain_record datasource

Charliekenney23:


displague:
@Charliekenney23 You'll want to include docs for the new datasource. See https://github.com/terraform-providers/terraform-provider-linode/pull/74/files for an example.

Charliekenney23:
@displague Just did both. Thanks!

linode/data_source_linode_domain_record.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/linode/linodego"
+)
+
+func dataSourceLinodeDomainRecord() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceLinodeDomainRecordRead,
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
```
displague: TypeInt - also affects https://github.com/terraform-providers/terraform-provider-linode/pull/85/files#diff-7afd8b7c70ffcf69664d792604948e34R84-R86

linode/data_source_linode_domain_record.go
```diff
@@ -0,0 +1,140 @@
...
+				Optional: true,
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The name of the Reocrd.",
```
displague: "record" typo

linode/data_source_linode_domain_record.go
```diff
@@ -0,0 +1,140 @@
...
+				Optional:    true,
+			},
+			"domain_id": {
+				Type:        schema.TypeString,
```
displague: typeint - also affects https://github.com/terraform-providers/terraform-provider-linode/pull/85/files#diff-7afd8b7c70ffcf69664d792604948e34R84-R86

linode/data_source_linode_domain_record.go
displague: For example, https://github.com/terraform-providers/terraform-provider-linode/blob/master/linode/data_source_linode_volume.go#L16-L20 (there are other `id` fields that are String, so it wasn't clear if this was intentional. In other resources, like `image`, the `id` is actually a string.

Charliekenney23: Ah okay. I definitely felt like this made most sense as an `int` but was trying to be consistent with the [other file](https://github.com/terraform-providers/terraform-provider-linode/blob/master/linode/data_source_linode_domain.go#L18) I was referencing.

We can start to make some progress on correcting these ids.

Charliekenney23: I went ahead and fixed this and [this](https://github.com/terraform-providers/terraform-provider-linode/pull/85#discussion_r380328916).

PR #86, opened 2020-02-17
Add stackscripts datasource

Charliekenney23:


linode/data_source_linode_stackscript_test.go
```diff
@@ -0,0 +1,45 @@
...
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testDataSourceLinodeStackScriptBasic(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "label", "my_stackscript"),
+					resource.TestCheckResourceAttr(resourceName, "description", "test"),
+					resource.TestCheckResourceAttr(resourceName, "script", "#!/bin/bash\necho hello world\n"),
+				),
```
displague: Images is also set so you can check for that too. There are other fields whose setters would be cheap to test as well but it will require more stackscript parameters in the fixture.

linode/data_source_linode_stackscript.go
```diff
@@ -0,0 +1,149 @@
...
+		d.Set("rev_note", ss.RevNote)
+		d.Set("is_public", ss.IsPublic)
+		d.Set("images", ss.Images)
+		d.Set("user_gravatar_id", ss.UserGravatarID)
+		d.Set("deployments_active", ss.DeploymentsActive)
+		d.Set("deployments_total", ss.DeploymentsTotal)
+		d.Set("username", ss.Username)
+		d.Set("created", ss.Created)
+		d.Set("updated", ss.Updated)
+		d.Set("user_defined_fields", ss.UserDefinedFields)
```
displague: I think this field is more complex than a simple `Set`.

Compare with https://github.com/terraform-providers/terraform-provider-linode/blob/02c1b76d2c3c1b93eee1ef822d896d0a6dffbc8c/linode/resource_linode_stackscript.go#L202-L221

PR #120, opened 2020-03-23
Version go tools with go modules

Charliekenney23:
Use go modules for versioning our golang build-time tools. This negates pulling large docker images.

Follows best practice: [Tools as dependencies](https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md#tools-as-dependencies).


phillc:
I think you need a go mod tidy

Charliekenney23:
@phillc I have run `go mod tidy`. A huge caveat to this approach of tool dependency management is that go mod v2 has no concept of build dependencies and both of these tools add lots of indirect dependencies. 

Fortunately, we don't have to worry about downstream consumers incurring these dependencies since we only ship the binary.

PR #131, opened 2020-04-09
Allow A record creation without a name

synical:
I recently noticed that as of https://github.com/terraform-providers/terraform-provider-linode/pull/88 it is no longer possible to create an A record via this terraform provider without the name field being set.

This broke a workflow for me and I see that the constraint is not within the Linode API itself, but rather the terraform validation logic.

Given that the Linode API does not have this constraint (works fine via the portal and via this branch), could we update this logic so that it matches the API?

I've only made this change to illustrate the issue, aware that there is more work to be done and I'm happy to do that.


Charliekenney23:
LGTM, thanks!

jelligett:
Encountering the same issue with [AAAA records](https://github.com/terraform-providers/terraform-provider-linode/blob/master/examples/main.tf#L113):

```
Error: name is required for non-SRV records
```

PR #184, opened 2020-09-22
disable release.draft in goreleaser workflow

Charliekenney23:
This configuration prevents releases from being published after the
workflow completes. See https://goreleaser.com/customization/release.


phillc (approved):
Is that really the only change needed? Part of me thinks there is something else up.

PR #188, opened 2020-09-30
remove vendor

Charliekenney23:
red diffs beat green diffs 😄


phillc (changes requested):
this no longer requires go111module in travis yml, correct?

phillc (approved):

PR #201, opened 2020-12-18
Updating stackscript.html.md

jordanfelle:
Interpolation-only expressions are deprecated, suggest updating the docs to cover that.


Charliekenney23 (approved):
Good catch

PR #221, opened 2021-03-03
allow upgrading lke_cluster.k8s_version

Charliekenney23:


linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+		}
+	}
+
+	poolRecyclesCh := make(chan int)
+	errCh := make(chan error)
+
+	eventTicker := time.NewTicker(200 * time.Millisecond)
+	lastEventID := 0
+
+	go func() {
```
phillc, line 465: this is going to end up in linodego someday I bet. Good job keeping it clear of any terraform specific logic.

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+
+		for len(clusterInstances) != 0 {
+			select {
+			case <-ctx.Done():
+				log.Printf("[ERROR] timed out waiting for all original nodes of LKE Cluster (%d) to be deleted (%d remaining)\n", clusterID, len(clusterInstances))
+				return
+
+			case <-eventTicker.C:
+				filterBytes, _ := json.Marshal(map[string]interface{}{
+					"+order_by": "created",
```
phillc, line 478: Is ID a valid value here? Makes me nervous that one logic is on dates, and the other is on IDs.

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+				})
+
+				events, err := client.ListEvents(ctx, linodego.NewListOptions(1, string(filterBytes)))
+				if err != nil {
+					errCh <- err
+					return
+				}
+
+				if len(events) != 0 {
+					lastEventID = events[0].ID
```
phillc, line 492: lastEventID is declared outside the goroutine. I think it can come inside?

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+				}
+
+				for _, event := range events {
+					if event.Action != linodego.ActionLinodeDelete {
+						continue
+					}
+					id := int(event.Entity.ID.(float64))
```
phillc, line 499: float?

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
 	return
 }
 
+func waitForClusterPoolReady(ctx context.Context, client *linodego.Client, errCh chan<- error, wg *sync.WaitGroup, clusterID, poolID int) {
+	eventTicker := time.NewTicker(500 * time.Millisecond)
```
phillc, line 418: should this be configurable like Min/MaxRetryDelayMilliseconds

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+			pool, err := client.GetLKEClusterPool(ctx, clusterID, poolID)
+			if err != nil {
+				errCh <- fmt.Errorf("failed to get LKE Cluster (%d) Pool (%d): %w", clusterID, poolID, err)
+			}
+
+			ready := true
+			for _, instance := range pool.Linodes {
+				if instance.Status == linodego.LKELinodeNotReady {
+					ready = false
+				}
```
phillc, line 436: as soon as you hit one, you can stop looping. Maybe instead of ready bool, just bring that logic from below right here and return?

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+		return fmt.Errorf("failed to recycle LKE Cluster (%d): %s", id, err)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	poolRecyclesCh, errCh := waitForClusterPoolsToStartRecycle(ctx, client, id, pools)
+	poolsRecycled := new(sync.WaitGroup)
+	poolsRecycled.Add(len(pools))
+	poolsRecycledCh := waitGroupCh(poolsRecycled)
```
phillc, line 533: Is this really necessary? We can discuss later.

phillc (changes requested):

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+
+	readyErrCh := make(chan error)
```
phillc, line 535: close me

Charliekenney23: oops

linode/resource_linode_lke_cluster.go
Charliekenney23: I discussed the acceptance criteria for upgrading a cluster with Tim and he said we need to ensure every cluster instance is ready, in case it fails. So I have to wait for all instance IDs to be ready.

linode/resource_linode_lke_cluster.go
Charliekenney23: good point, yeah this can feed off of a global configuration for polling.

linode/resource_linode_lke_cluster.go
```diff
@@ -406,6 +414,150 @@ func reconcileLKEClusterPoolSpecs(poolSpecs []linodeLKEClusterPoolSpec, pools []
...
+						poolRecyclesCh <- poolID
+					}
+				}
+			}
+		}
+	}()
+	return poolRecyclesCh, errCh
+}
+
+func recycleLKECluster(ctx context.Context, client *linodego.Client, id int, pools []linodego.LKEClusterPool) error {
```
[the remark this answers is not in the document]
Charliekenney23: I log every time a cluster pool is ready and every time a full pool recycle has finished getting started (all pre-existing nodes have been deleted). There is also a log that says "still updating" every 10s. Any other points that should be covered?

linode/resource_linode_lke_cluster.go
Charliekenney23: It does make control flow easier in my opinion, handling the wg in the same place as the error channels. I'm open to suggestions though, the whole thing is a bit complex.

linode/resource_linode_lke_cluster.go
Charliekenney23: this would result in
```
panic: interface conversion: interface {} is float64, not int
```

Go will always map numeric values to int64 when trying to unwrap an `interface{}`. See https://play.golang.org/p/acM6ltDKIYY

phillc (approved):

PR #234, opened 2021-03-11
Add firewall datasource

LBGarber:
This pull request adds a datasource for `linode_firewall`.

### DIFF
--- a/linode/data_source_linode_firewall.go
+++ b/linode/data_source_linode_firewall.go
@@ -1,0 +1,156 @@
+package linode
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/linode/linodego"
+)
+
+func dataSourceLinodeFirewallRule() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"label": {
+				Type:        schema.TypeString,
+				Description: `Used to identify this rule. For display purposes only.`,
+				Computed:    true,
+			},
+			"action": {
+				Type:        schema.TypeString,
+				Description: `Controls whether traffic is accepted or dropped by this rule. Overrides the Firewall’s inbound_policy if this is an inbound rule, or the outbound_policy if this is an outbound rule.`,
+				Computed:    true,
+			},
+			"ports": {
+				Type:        schema.TypeString,
+				Description: `A string representation of ports and/or port ranges (i.e. "443" or "80-90, 91").`,
+				Computed:    true,
+			},
+			"protocol": {
+				Type:        schema.TypeString,
+				Description: "The network protocol this rule controls.",
+				Computed:    true,
+			},
+			"ipv4": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "A list of IP addresses, CIDR blocks, or 0.0.0.0/0 (to allow all) this rule applies to.",
+				Computed:    true,
+			},
+			"ipv6": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Description: "A list of IPv6 addresses or networks this rule applies to.",
+				Computed:    true,
+			},
+		},
+	}
+}
+
+func dataSourceLinodeFirewall() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: datasourceLinodeFirewallRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeInt,
+				Required: true,
+			},
+			"label": {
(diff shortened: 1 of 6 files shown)


linode/data_source_linode_firewall.go
```diff
@@ -0,0 +1,189 @@
...
+					Type: schema.TypeString,
+				},
+				Description: "A list of IPv6 addresses or networks this rule applies to.",
+				Computed:    true,
+			},
+		},
+	}
+}
+
+func dataSourceLinodeFirewallDevice() *schema.Resource {
```
Charliekenney23, line 55: maybe it would make sense to make this function in `resource_linode_firewall.go` so it can be used in both the datasource and resource, as both are fully computed schemas.

linode/data_source_linode_firewall.go
```diff
@@ -0,0 +1,189 @@
...
+				Description: "An array of tags applied to this object. Tags are for organizational purposes only.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Computed:    true,
+				Set:         schema.HashString,
+			},
+			"disabled": {
+				Type:        schema.TypeBool,
+				Description: "If true, the Firewall is inactive.",
+				Computed:    true,
+				Default:     nil,
```
Charliekenney23, line 111: We don't need a default for this, it will always be populated by the API.

linode/data_source_linode_firewall_test.go
```diff
@@ -0,0 +1,64 @@
...
+	firewallName := acctest.RandomWithPrefix("tf_test")
+	devicePrefix := acctest.RandomWithPrefix("tf_test")
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLinodeLKEClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testDataSourceLinodeFirewallBasic(firewallName, devicePrefix),
+				Check: resource.ComposeTestCheckFunc(
```
Charliekenney23, line 23: Love how exhaustive this test is !

linode/data_source_linode_firewall_test.go
```diff
@@ -0,0 +1,64 @@
...
+					resource.TestCheckResourceAttrSet(testFirewallDataName, "devices.0.id"),
+					resource.TestCheckResourceAttrSet(testFirewallDataName, "devices.0.entity_id"),
+					resource.TestCheckResourceAttrSet(testFirewallDataName, "devices.0.label"),
+				),
+			},
+		},
+	})
+}
+
+func testDataSourceLinodeFirewallBasic(firewallName string, devicePrefix string) string {
```
Charliekenney23, line 58: The function signature would be better written as `(firewallName, devicePrefix string) string`

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 FEATURES:
 
 * **New Data Source** `linode_lke_cluster`
+* **New Data Source** `linode_firewall`
```
Charliekenney23, line 6: This release has already shipped so we'll use the release drafter instead.

Charliekenney23 (changes requested):
Looks good! Just a few things

linode/data_source_linode_firewall.go
LBGarber: Great idea!

Charliekenney23 (approved):

linode/data_source_linode_firewall.go
```diff
@@ -0,0 +1,156 @@
...
+	d.SetId(strconv.Itoa(id))
+	d.Set("label", firewall.Label)
+	d.Set("tags", firewall.Tags)
+	d.Set("disabled", firewall.Status == linodego.FirewallDisabled)
+	d.Set("inbound", flattenLinodeFirewallRules(rules.Inbound))
+	d.Set("inbound_policy", rules.InboundPolicy)
+	d.Set("outbound", flattenLinodeFirewallRules(rules.Outbound))
+	d.Set("outbound_policy", rules.OutboundPolicy)
+	d.Set("status", firewall.Status)
+	d.Set("linodes", flattenLinodeFirewallLinodes(devices))
```
phillc, line 152: good job finding the code reuse

linode/resource_linode_firewall.go
```diff
@@ -56,6 +56,38 @@ func resourceLinodeFirewallRule() *schema.Resource {
...
+			},
+			"type": {
+				Type:        schema.TypeString,
+				Description: "The type of firewall device.",
+				Computed:    true,
+			},
+			"label": {
+				Type:        schema.TypeString,
+				Description: "The label of the underlying entity for the firewall device.",
+				Computed:    true,
```
phillc, line 80: is that really computed? If it is like that in the resource maybe it needs to be changed?

LBGarber: Oh hmm, let me look into that.

LBGarber: Just looked through the resource in master and everything under devices seems to be computed, so I'll assume this can stay.

PR #243, opened 2021-03-22
Make NodeBalancer config protocol case-insensitive

LBGarber:


LBGarber:
It may be a good idea for to also make the `check` field case-insensitive, but I'll wait for feedback on that idea.

Charliekenney23 (approved):

PR #245, opened 2021-03-26
Add lifecycle and versioning for linode_object_storage_bucket resource

LBGarber:


Charliekenney23 (changes requested):

linode/resource_linode_object_storage_bucket.go
```diff
@@ -85,6 +188,31 @@ func resourceLinodeObjectStorageBucketRead(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("failed to find the access config for the specified Linode ObjectStorageBucket: %s", err)
 	}
 
+	// Funtionality requiring direct S3 API access
+	if accessKey != "" && secretKey != "" {
```
Charliekenney23, line 192: Instead of going off the keys alone here, we should see if lifecycleRule is present or versioning is enabled, then try to set these on the bucket.

linode/resource_linode_object_storage_bucket.go
```diff
@@ -225,3 +397,130 @@ func decodeLinodeObjectStorageBucketID(id string) (cluster, label string, err er
...
+			ruleMap["noncurrent_version_expiration"] = []interface{}{e}
+		}
+
+		result[i] = ruleMap
+	}
+
+	return result
+}
+
+func parseLifecycleRules(ruleSpecs []interface{}) ([]*s3.LifecycleRule, error) {
```
Charliekenney23, line 457: expandLifecycleRules is probably a more accurate name here

Charliekenney23 (changes requested):
Looks good for the most part, just a few things.

linode/resource_linode_object_storage_bucket.go
Charliekenney23: like ideally, we would thrown an error about not having the keys but having one of these fields present.

Charliekenney23 (approved):

PR #247, opened 2021-03-31
Fix issue with NodeBalancers and Linode instance recreation

LBGarber:
This pull request fixes an error that occurs when a Linode instance currently associated with a NodeBalancer is force-recreated (for example: when there are changes to authorized_keys).

### DIFF
--- a/linode/resource_linode_nodebalancer_node.go
+++ b/linode/resource_linode_nodebalancer_node.go
@@ -181,6 +181,18 @@ func resourceLinodeNodeBalancerNodeUpdate(d *schema.ResourceData, meta interface
 		return fmt.Errorf("Error parsing Linode NodeBalancer ID %v as int", d.Get("config_id"))
 	}
 
+	// If node doesn't exist, create a new one.
+	// This is necessary because the deletion of an instance
+	// will automatically delete its associated node.
+	_, err = client.GetNodeBalancerNode(context.Background(), nodebalancerID, configID, int(id))
+	if err != nil {
+		if err.(*linodego.Error).Code == 404 {
+			return resourceLinodeNodeBalancerNodeCreate(d, meta)
+		}
+
+		return fmt.Errorf("failed to get nodebalancer node %v: %s", id, err)
+	}
+
 	updateOpts := linodego.NodeBalancerNodeUpdateOptions{
 		Address: d.Get("address").(string),
 		Label:   d.Get("label").(string),
--- a/linode/resource_linode_nodebalancer_node.go
+++ b/linode/resource_linode_nodebalancer_node.go
@@ -57,6 +57,7 @@ func resourceLinodeNodeBalancerNode() *schema.Resource {
 				Type:        schema.TypeString,
 				Description: "The private IP Address and port (IP:PORT) where this backend can be reached. This must be a private IP address.",
 				Required:    true,
+				ForceNew:    true,
 			},
 			"status": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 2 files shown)


linode/resource_linode_nodebalancer_node.go
```diff
@@ -181,6 +181,18 @@ func resourceLinodeNodeBalancerNodeUpdate(d *schema.ResourceData, meta interface
 		return fmt.Errorf("Error parsing Linode NodeBalancer ID %v as int", d.Get("config_id"))
 	}
 
+	// If node doesn't exist, create a new one.
+	// This is necessary because the deletion of an instance
+	// will automatically delete its associated node.
+	_, err = client.GetNodeBalancerNode(context.Background(), nodebalancerID, configID, int(id))
```
Charliekenney23, line 187: The canonical way recreating valid resources that are prone to be deleted through some side-effect is through the `read` function. 

Essentially the same structure but you wipe the id from the state, which will trigger recreation:
```
if linodeErr, ok := err.(*linodego.Error); ok && linodeErr.Code == 404 {
   d.SetId("")
    return nil
}
```

linode/resource_linode_nodebalancer_node_test.go
```diff
@@ -227,6 +238,18 @@ resource "linode_nodebalancer_node" "foonode" {
 `, label)
 }
 
+func testAccCheckLinodeNodeBalancerNodeInstanceUpdates(label string) string {
+	return testAccCheckLinodeInstanceConfigPrivateNetworking(label, publicKeyMaterial+"_new") + testAccCheckLinodeNodeBalancerConfigBasic(label) + fmt.Sprintf(`
```
Charliekenney23, line 242: I'd rather the label change or some other property instead of making the public key invalid

linode/resource_linode_nodebalancer_node.go
```diff
@@ -181,6 +181,18 @@ func resourceLinodeNodeBalancerNodeUpdate(d *schema.ResourceData, meta interface
...
+	if err != nil {
+		if err.(*linodego.Error).Code == 404 {
```
Charliekenney23, line 189: This line will result in a `panic` and crash the provider if `err` isn't `*linodego.Error`

linode/resource_linode_nodebalancer_node_test.go
```diff
@@ -73,6 +73,17 @@ func TestAccLinodeNodeBalancerNode_update(t *testing.T) {
 					resource.TestCheckResourceAttr(resName, "weight", "200"),
 				),
 			},
+			{
```
Charliekenney23, line 76: Can we make this a different test? Maybe `TestAccLinodeNodeBalancerNode_instanceRecreated`

linode/resource_linode_nodebalancer_node.go
Charliekenney23: If this doesn't work, I'd rather just mention that this is a known side effect of deleting a linode that has a nodebalancer dependency because creating in update just doesn't follow terraform design

Charliekenney23 (changes requested):

linode/resource_linode_nodebalancer_node_test.go
LBGarber: Ah yeah, this was mainly to reproduce the original issue (it's an easy way to force recreation). Alternatively we could change the image, region, root pass, etc.

LBGarber:
This issue is a won’t-fix because it conflicts with the design principles of Terraform.

PR #250, opened 2021-04-06
Add linode_images datasource

LBGarber:
- Add `linode_images` datasource

Usage:

```hcl
data "linode_images" "specific-images" {
  filter {
    name = "label"
    values = ["Debian 8"]
  }

  filter {
    name = "is_public"
    values = ["true"]
  }
}
```

```hcl
data "linode_images" "all-images" {}
```

### DIFF
--- a/linode/data_source_linode_images.go
+++ b/linode/data_source_linode_images.go
@@ -1,0 +1,63 @@
+package linode
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/linode/linodego"
+
+	"context"
+	"fmt"
+	"strconv"
+)
+
+func dataSourceLinodeImages() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceLinodeImagesRead,
+		Schema: map[string]*schema.Schema{
+			"filter": filterSchema(),
+			"images": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     dataSourceLinodeImage(),
+			},
+		},
+	}
+}
+
+func dataSourceLinodeImagesRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderMeta).Client
+
+	filter, err := constructFilterString(d, imageValueToFilterType)
+	if err != nil {
+		return fmt.Errorf("failed to construct filter: %s", err)
+	}
+
+	images, err := client.ListImages(context.Background(), &linodego.ListOptions{
+		Filter: filter,
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to list linode images: %s", err)
+	}
+
+	imagesFlattened := make([]interface{}, len(images))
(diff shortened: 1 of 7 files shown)


linode/data_source_linode_images.go
```diff
@@ -0,0 +1,88 @@
...
+		imagesFlattened[i] = flattenLinodeImage(&image)
+	}
+
+	d.SetId(filter)
+	d.Set("image", imagesFlattened)
+
+	return nil
+}
+
+func flattenLinodeImage(image *linodego.Image) map[string]interface{} {
```
Charliekenney23, line 54: could this be reused?

linode/data_source_linode_images.go
```diff
@@ -0,0 +1,88 @@
...
+	if image.Expiry != nil {
+		result["expiry"] = image.Expiry.Format(time.RFC3339)
+	}
+
+	return result
+}
+
+func imageValueToFilterType(filterName, value string) (interface{}, error) {
+	switch filterName {
+	case "deprecated", "is_public":
```
Charliekenney23, line 80: perhaps a default case here would be good to error on unsupported filters?

linode/data_source_linode_images_test.go
```diff
@@ -0,0 +1,100 @@
...
+				Config: testDataSourceLinodeImagesNoFilters(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccDataSourceCheckLinodeAllImagesNotEmpty(resourceName),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataSourceCheckLinodeAllImagesNotEmpty(resourceName string) resource.TestCheckFunc {
```
Charliekenney23, line 57: I recognize that this is a common thing, could this be a more abstract helper? For instance in the test helpers file you could make the function `testAccCheckResourceNonEmptyList`

linode/data_source_linode_images_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"testing"
+)
+
+func TestAccDataSourceLinodeImages_basic(t *testing.T) {
+	t.Parallel()
+
+	imageID := "linode/debian8"
+	resourceName := "data.linode_images.foobar"
+
+	resource.Test(t, resource.TestCase{
```
Charliekenney23, line 18: This test could fail with later changes. It's probably best to build an image of our own and test filtering for it. See `resource_linode_image_test.go`

Charliekenney23 (changes requested):

linode/data_source_linode_images.go
LBGarber: Unfortunately in the image data source it writes directly to the `ResourceData`, so it doesn't seem like we can reuse it there.

Charliekenney23: I meant can we put this function in the resource and reuse it in the data source like with the other types?

Charliekenney23 (approved):

PR #266, opened 2021-04-13
Add grants to linode_user resource

LBGarber:
- Add `grants` field to linode_user resource

### DIFF
--- a/linode/resource_linode_user.go
+++ b/linode/resource_linode_user.go
@@ -1,12 +1,109 @@
 package linode
 
 import (
-	"context"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/linode/linodego"
+
+	"context"
+	"fmt"
 )
+
+var resourceLinodeUserGrantFields = []string{"global_grants", "domain_grant", "image_grant", "linode_grant",
+	"longview_grant", "nodebalancer_grant", "stackscript_grant", "volume_grant"}
+
+func resourceLinodeUserGrantsGlobal() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"account_access": {
+				Type: schema.TypeString,
+				Description: "The level of access this User has to Account-level actions, like billing information. " +
+					"A restricted User will never be able to manage users.",
+				Optional: true,
+			},
+			"add_domains": {
+				Type:        schema.TypeBool,
+				Description: "If true, this User may add Domains.",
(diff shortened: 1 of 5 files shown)


linode/resource_linode_user.go
```diff
@@ -59,6 +168,11 @@ func resourceLinodeUserCreate(ctx context.Context, d *schema.ResourceData, meta
 	if err != nil {
 		return diag.Errorf("failed to create user: %s", err)
 	}
+
+	if err := updateUserGrants(ctx, d, meta); err != nil {
```
Charliekenney23, line 172: Could we make a function `hasGrantsConfigured` which takes in the `schema.ResourceData` and returns whether or not any of these told level attributes have been specificed? We don't want to make a call to update grants (here or in update) unless it's explicitly configured as users might want to configure Grants using cloud manager.

linode/resource_linode_user.go
```diff
@@ -107,3 +243,140 @@ func resourceLinodeUserDelete(ctx context.Context, d *schema.ResourceData, meta
...
+		return err
+	}
+
+	return nil
+}
+
+// isUserGrantSchemaUpdated returns whether or not any grant-related field is updated.
+func isUserGrantSchemaUpdated(d *schema.ResourceData) bool {
+	result := false
+	for key := range resourceLinodeUser().Schema {
```
Charliekenney23, line 286: We should probably keep all user grant-related field names in a static string slice for use in multiple places. That would read much cleaner than dynamically filtering fields in the schema everytime

linode/resource_linode_user.go
```diff
@@ -107,3 +243,140 @@ func resourceLinodeUserDelete(ctx context.Context, d *schema.ResourceData, meta
...
+}
+
+func flattenGrantsEntities(entities []linodego.GrantedEntity) []interface{} {
+	var result []interface{}
+
+	for _, entity := range entities {
+		// Filter out entities without any permissions set.
+		// This is necessary because Linode will automatically
+		// create empty entities that will trigger false diffs.
+		if len(entity.Permissions) < 1 {
```
Charliekenney23, line 348: What is this magic number `1`?

linode/resource_linode_user.go
```diff
@@ -43,12 +137,27 @@ func resourceLinodeUser() *schema.Resource {
 				Computed:    true,
 				Description: "If the User has Two Factor Authentication (TFA) enabled.",
 			},
+			"global_grants": {
```
Charliekenney23, line 140: Add `MaxLen: 1`

Charliekenney23 (changes requested):

Charliekenney23 (approved):

PR #386, opened 2021-07-20
Fix LKE shuffled node pool diffing

LBGarber:
This pull requests addresses an intermittent non-empty diff on the `TestAccLinodeLKECluster_poolUpdates` acceptance test. As the Linode API does not guarantee the order of returned node pools, we must adapt the returned node pool list to best match the schema configuration.

Currently this cannot be achieved using a DiffSuppressFunc, so the API response must be sorted in the read routine.
See: https://github.com/hashicorp/terraform-plugin-sdk/issues/477

### DIFF
--- a/linode/resource_linode_lke_cluster.go
+++ b/linode/resource_linode_lke_cluster.go
@@ -140,6 +140,11 @@ func resourceLinodeLKEClusterRead(ctx context.Context, d *schema.ResourceData, m
 		return diag.Errorf("Error parsing Linode LKE Cluster ID: %s", err)
 	}
 
+	declaredPools, ok := d.Get("pool").([]interface{})
+	if !ok {
+		return diag.Errorf("failed to parse linode lke cluster pools: %d", id)
+	}
+
 	cluster, err := client.GetLKECluster(context.Background(), id)
 	if err != nil {
 		if lerr, ok := err.(*linodego.Error); ok && lerr.Code == 404 {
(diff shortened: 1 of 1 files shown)


linode/resource_linode_lke_cluster.go
```diff
@@ -140,6 +140,11 @@ func resourceLinodeLKEClusterRead(ctx context.Context, d *schema.ResourceData, m
 		return diag.Errorf("Error parsing Linode LKE Cluster ID: %s", err)
 	}
 
+	schemaPools, ok := d.Get("pool").([]interface{})
```
Charliekenney23, line 143: I think this would be more aptly named as `declaredPools`.

linode/resource_linode_lke_cluster.go
```diff
@@ -604,3 +610,34 @@ func flattenLinodeLKEClusterAPIEndpoints(apiEndpoints []linodego.LKEClusterAPIEn
 	}
 	return flattened
 }
+
+// This cannot currently be handled efficiently by a DiffSuppressFunc
+// See: https://github.com/hashicorp/terraform-plugin-sdk/issues/477
+func matchPoolsWithSchema(pools []linodego.LKEClusterPool, schemaPools []interface{}) []linodego.LKEClusterPool {
```
Charliekenney23, line 616: same as above `schemaPools` -> `declaredPools`

linode/resource_linode_lke_cluster.go
```diff
@@ -604,3 +610,34 @@ func flattenLinodeLKEClusterAPIEndpoints(apiEndpoints []linodego.LKEClusterAPIEn
...
+	result := make([]linodego.LKEClusterPool, len(schemaPools))
+
+	poolMap := make(map[int]linodego.LKEClusterPool)
```
Charliekenney23, line 619: `poolMap := make(map[int]linodego.LKEClusterPool, len(schemaPools))` to avoid unnecessary reallocation.

Charliekenney23 (changes requested):
Looks good! just some small changes

Charliekenney23 (approved):

PR #394, opened 2021-07-27
Make all resources and datasources context-aware

jriddle-linode:
removing context.background() instances to make sure we use the same contexts throughout actions


Charliekenney23 (approved):
LGTM. good to merge after linter errors for long lines are fixed

LBGarber (approved):
Perfect, looks like you've covered everything!

PR #398, opened 2021-07-29
handle instance reboot if creating IP

jriddle-linode:
moved reboot into its own function and use it between instance.go and instance_ip.go

### DIFF
--- a/linode/resource_linode_instance_ip.go
+++ b/linode/resource_linode_instance_ip.go
@@ -110,6 +110,13 @@ func resourceLinodeInstanceIPCreate(ctx context.Context, d *schema.ResourceData,
 	}
 
 	d.SetId(ip.Address)
+
+	// TODO(jxriddle): check instance status and react to state
+
+	// setting bootConfig to 0 to use current bootConfig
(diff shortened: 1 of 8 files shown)


linode/resource_linode_instance_ip.go
```diff
@@ -110,6 +110,13 @@ func resourceLinodeInstanceIPCreate(ctx context.Context, d *schema.ResourceData,
 	}
 
 	d.SetId(ip.Address)
+
+	// TODO(jxriddle): check instance status and react to state
```
LBGarber, line 114: We might want to add some logic here to ensure the instance is `running` to avoid unwanted reboots if an instance is shut down. This will not handle _all_ edge cases, but it should improve the user experience.

jriddle-linode: moved this into the function itself instead of the call

LBGarber (approved):
Looks great! We should look into handling more unique instances statuses in the future, but this is perfect for now!

Charliekenney23 (approved):
LGTM

Charliekenney23 (approved):

PR #435, opened 2021-08-20
moving object objectcluster objectbucket and object key to their own packages

jriddle-linode:
- renamed to remove unnecessary ObjectStorageObject redunancy
- moved object helpers to helper
- cleaned up function names
- added new destroy check to ObjectStorageObject to check if object is destroyed, not key


LBGarber (approved):
LGTM!

PR #450, opened 2021-08-30
Use test templating for NodeBalancer test configurations

LBGarber:
This change replaces hardcoded test configurations with templates for all nodebalancer related resources and data sources.


jriddle-linode (approved):
so much cleaner than the sprintfs in the bottom, and easier to reuse

PR #451, opened 2021-08-30
moving last features to packages

jriddle-linode:
moving the instance monolith to packages and cleaning up code

will do one more PR for packaging after this to align all the naming conventions and check for sanity
may need to modify makefile to work with testing the full suite


LBGarber (approved):
Looks perfect! Just fix up those linting errors and merge :sunglasses:

PR #485, opened 2021-09-16
Add additional valid domain TTLs

### LINKED ISSUE #484: linode_domain_record.ttl_sec omits options under 300 seconds, despite them being available in the UI
### Terraform Version

```console
Terraform v1.0.1
on linux_amd64
```

### Effected Resource(s)

linode_domain_record

### Terraform Configuration Files

```terraform
resource "linode_domain" "main" {
  type = "master"
  domain = var.domain
  soa_email = ".................."
  tags = [var.name]
}

resource "linode_domain_record" "home_ipv4" {
  count = var.servers
  domain_id = linode_domain.main.id
  name = ""
  record_type = "A"
  target = element(linode_instance.servers.*.ip_address, count.index)
  ttl_sec = 30 # the issue
}
```

### Expected Behavior

My record is configured to have 30 second ttl, as in:
image

This is consistent with the options presented in the UI:
image

My hope is that this lower ttl will increase the chances of my Round-Robin DNS loadbalancing strategy working.

### Actual Behavior

My record is configured to have 300 second ttl, as in:
image

This is consistent with the options presented in the docs:
image

### Steps to Reproduce

1. Create a `linode_domain`
2. Create a `linode_domain_record` with `ttl = 30`
3. Observe ttl is rounded up to 300

### Note

I believe the bug is that the array here is missing several valid values, namely 30 and 120: https://github.com/linode/terraform-provider-linode/blob/56b85508d9284709a69c41e7f439fe071d61bcfb/linode/helper/domain.go#L11

But I also realize this may not be a bug, and you have simply changed your mind about offering low TTLs. Still, I would question the logic of offering them in the UI but not via terraform.

LBGarber:
This pull request adds additional valid values for seconds-related fields in the `linode_domain` and `linode_domain_record` resources. These values are accepted by the API and are available in Cloud Manager. 

This pull request also adds additional test cases for TTLs of `0` in both resources.

Resolves #484

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/hashicorp/terraform-plugin-go v0.3.0 // indirect
 	github.com/hashicorp/yamux v0.0.0-20181012175058-2f1d1f20f75d // indirect
 	github.com/imdario/mergo v0.3.12 // indirect
-	github.com/jmespath/go-jmespath v0.3.0 // indirect
+	github.com/jmespath/go-jmespath v0.4.0 // indirect
 	github.com/json-iterator/go v1.1.8 // indirect
 	github.com/jstemmer/go-junit-report v0.9.1 // indirect
 	github.com/klauspost/compress v1.11.2 // indirect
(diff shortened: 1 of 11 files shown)


jriddle-linode (approved):
LGTM

PR #503, opened 2021-10-01
Add linode_instance_types data source

### LINKED ISSUE #501: [Feature]: datasource linode_instance_types
### Description

Something that I normally do, is explore data sources when I am looking at a terraform provider. When doing that, I normally explore the necessary datasources that I need for constructing a resource. I was going to try and deploy a `linode_instances` resource, and when looking at the docs it showed that I would need an `instance_type`. Unfortunately it doesn't appear to exist `instance_types` data source ( from my understanding this would be analogous to DigitalOcean's (DO) [droplets](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/droplets) terraform data source.

So, ideally there would be an `instance_types` module, and then you can specify optional tags ( similar to the `linode_images` ). That way you could return all instance_types and can filter down as necessary.

### New or Affected Terraform Resources

instance_types

### Potential Terraform Configuration

```terraform
data "linode_instance_types" "all" {}
```

or optionally apply a filter:

```terraform
data "linode_instance_types" "all" {
  filter {
    key    = "vcpus"
    values = ["1"]
  }
}

### PR DESCRIPTION
**Pull request by LBGarber**
This feature allows users to query all available Instance types using filters.

For example:

**Retrieving all Instance types**
```hcl
data "linode_instance_types" "specific-types" {}
```

**Retrieving all Instance types with two VCPUs**
```hcl
data "linode_instance_types" "specific-types" {
  filter {
    name = "vcpus"
    values = [2]
  }
}
```

Closes #501

### DIFF
--- a/linode/instancetypes/datasource.go
+++ b/linode/instancetypes/datasource.go
@@ -1,0 +1,91 @@
+package instancetypes
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/linode/linodego"
+	"github.com/linode/terraform-provider-linode/linode/helper"
+
+	"context"
+	"strconv"
+)
+
+func DataSource() *schema.Resource {
+	return &schema.Resource{
+		Schema:      dataSourceSchema,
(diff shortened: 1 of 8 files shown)


jriddle-linode (approved):
lgtm

PR #535, opened 2021-11-02
Add `linode_stackscripts` data source

LBGarber:
This pull request allows users to obtain a filtered list of StackScripts using the provider's filter data source framework.

For example:
```hcl
data "linode_stackscripts" "specific-stackscripts" {
  filter {
    name = "label"
    values = ["my-cool-stackscript"]
  }

  filter {
    name = "is_public"
    values = [false]
  }
}
```


jriddle-linode (approved):
lgtm

PR #543, opened 2021-11-12
Avoid unnecessary requests in `linode_instances` data source

LBGarber:
This change prevents additional API requests from being made when collecting instances for client-side filtering. Instances are now flattened into a simple and filterable map for the initial filtering pass and additional requests are made for the remaining instances. This change additionally prevents race conditions between concurrent actions on separate Linodes.


jriddle-linode (approved):
LGTM

PR #552, opened 2021-11-18
Do not reattach already-attached volumes on `linode_instance` config updates

LBGarber:
This pull request reduces the number of hidden Block Storage volume reattachments by only detaching/attaching a volume if it is not already attached to the Instance.

### DIFF
--- a/linode/helper/events.go
+++ b/linode/helper/events.go
@@ -2,7 +2,6 @@ package helper
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/linode/linodego"
 )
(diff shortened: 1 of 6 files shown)


linode/helper/events.go
```diff
@@ -35,7 +33,7 @@ func GetLatestEvent(ctx context.Context, client *linodego.Client,
 	}
 
 	if len(events) < 1 {
-		return nil, fmt.Errorf("failed to get event %s of %s (%d)", action, entityType, entityID)
```
LBGarber, line 38: No event being returned should not be an error case

jriddle-linode (approved):
LGTM

PR #569, opened 2021-12-13
LKE - Add support for pool tags and custom disk partitions.

kekoav:
For LKE, adds pool tags and custom disk partitions. This support is already present in the golinode library and API, just exposing it for Terraform.

### DIFF
--- a/linode/lke/cluster.go
+++ b/linode/lke/cluster.go
@@ -338,6 +338,14 @@ func waitGroupCh(wg *sync.WaitGroup) <-chan struct{} {
 	return done
 }
 
+func poolMatchesDeclaredPool(pool *linodego.LKEClusterPool, declaredPool map[string]interface{}) bool {
+	if pool.Type != declaredPool["type"] || pool.Count != declaredPool["count"] {
+		return false
+	}
+
+	return true
+}
+
 // This cannot currently be handled efficiently by a DiffSuppressFunc
 // See: https://github.com/hashicorp/terraform-plugin-sdk/issues/477
 func matchPoolsWithSchema(pools []linodego.LKEClusterPool, declaredPools []interface{}) []linodego.LKEClusterPool {
@@ -352,13 +360,12 @@ func matchPoolsWithSchema(pools []linodego.LKEClusterPool, declaredPools []inter
 		declaredPool := declaredPool.(map[string]interface{})
 
 		for key, pool := range poolMap {
-			if pool.Count != declaredPool["count"] || pool.Type != declaredPool["type"] {
-				continue
+			if pool.ID == declaredPool["id"] || poolMatchesDeclaredPool(&pool, declaredPool) {
+				result[i] = pool
+				delete(poolMap, key)
+				break
 			}
 
-			result[i] = pool
-			delete(poolMap, key)
-			break
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


linode/lke/cluster.go
```diff
@@ -338,6 +338,14 @@ func waitGroupCh(wg *sync.WaitGroup) <-chan struct{} {
 	return done
 }
 
+func poolMatchesDeclaredPool(pool *linodego.LKEClusterPool, declaredPool map[string]interface{}) bool {
+	if pool.Type != declaredPool["type"] || pool.Count != declaredPool["count"] {
+		return false
+	}
+
+	return true
+}
```
LBGarber, line 347: Just a small stylistic change :+1: 
```suggestion
return pool.Type == declaredPool["type"] && pool.Count == declaredPool["count"]
```

LBGarber (changes requested):
Hey there! Thanks for the great contribution :smile: 

Just a few small changes and we should be good to go. Do you mind if I push a commit to this PR shortly to add integration tests?

website/docs/r/lke_cluster.html.md
```diff
@@ -75,6 +75,11 @@ The following arguments are supported in the `pool` specification block:
 
 * [`autoscaler`](#autoscaler) - (Optional) If defined, an autoscaler will be enabled with the given configuration.
 
+* `tags` - (Optional) An array of tags applied to the Node Pool. Tags are for organizational purposes only.
+
+* [`disk`](#disk) - (Optional) An array of custom partitions to create on each node, in addition to the boot partition.
+
+
```
LBGarber, line 82: This should fix the docs linting workflow
```suggestion
```

website/docs/d/lke_cluster.html.md
```diff
@@ -62,4 +62,12 @@ In addition to all arguments above, the following attributes are exported:
 
     * `max` - The maximum number of nodes to autoscale to.
 
+  * `tags` - The tags applied to the node pool.
+
+  * `disk` - The custom disk partitions for the node pool.
+  
+    * `size` - The size in MB of the disk partition.
+
+    * `type` - The filesystem type of the partition. (`raw`, `ext4`)
```
LBGarber, line 71: Remove trailing space for docs linting
```suggestion
    * `type` - The filesystem type of the partition. (`raw`, `ext4`)
```

LBGarber (changes requested):
Just a few changes to get the docs workflow passing

kekoav:
@LBGarber I don't mind at all! It's all yours, your changes look good to me.

kekoav:
@LBGarber I was trying to add some integration tests, but it turned out it was going to be more than I bargained for... Adding Tags and Disk to `ClusterPoolSpec` no longer made it suitable for use as a map key which it was in several places, and I just didn't have the time to dig into how all that worked.

LBGarber:
While writing the integration tests I ran into a few issues with how update logic is handled. I'll see what I can do about this and get back to you! :+1:

LBGarber:
@kekoav Thanks for the feedback! I definitely agree with you there, a full cluster recreation should only be a last resort and shouldn't be triggered by a node pool update. I think it would vastly simplify the update logic here to have node pools be their own resource, but this currently doesn't mesh well with the existing LKE API architecture. I'll see what I can do about implementing disks into the existing node pool update logic.
 
> Let me know how I can help, I don't want to get in the way but would be happy to provide use cases from the perspective of a typical Kubernetes user.

Feedback is always appreciated! If you have any ideas or suggestions, feel free to let me know :slightly_smiling_face:

LBGarber:
Sorry about that, looks like the stale workflow didn't respect the `keep` tag.

akoidan:
Is there anything preventing reopening and merging this PR? I'm facing the same issue, unable to get nodes of a newly created cluster described [here](https://github.com/linode/terraform-provider-linode/issues/595#issuecomment-1566092555)

PR #570, opened 2021-12-17
Add `wait_for_available` field to `linode_rdns` resource

LBGarber:
This pull request adds a `wait_for_available` field to the `linode_rdns` resource. When set to true, the RDNS assignment will be retried within the resource's operation timeout (default 10 minutes).

For example:
```hcl
resource "linode_rdns" "foo" {
  depends_on = [linode_domain_record.foobar]

  address = linode_instance.foo.ip_address
  rdns = "myrecord.mydomain.com"
  wait_for_available = true
}

resource "linode_domain_record" "foobar" {
  domain_id = 12345
  name = "myrecord"
  record_type = "A"
  target = linode_instance.foo.ip_address
}

  image = "linode/alpine3.14"
  region = "us-southeast"
  type = "g6-standard-2"
}
```

### DIFF
--- a/linode/rdns/resource.go
+++ b/linode/rdns/resource.go
@@ -2,13 +2,17 @@ package rdns
 
 import (
 	"context"
-	"log"
-
+	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/linode/linodego"
 	"github.com/linode/terraform-provider-linode/linode/helper"
+	"log"
+	"strings"
+	"time"
 )
+
+const updateRDNSTimeout = time.Minute * 10
 
 func Resource() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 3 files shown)


LBGarber:
It might be better to instead have some sort of opt-in polling logic in the `linode_domain_record` resource, but this should cover more use cases.

stvnjacobs:
I like it. Only two pieces of feedback would be ...
1. remove the 10 minute timeout and keep the 20 minute default on Create and Update actions. This is just my opinion, but the amount of time folks want to wait is up to them. 10 minutes is already eternity, so why not 20 if it can save the user from a failed apply and they can have all their rdns records set?
2. retrying every second could hit rate limits very quickly if a lot of rdns records are created at once and need to retry.

stvnjacobs:
For context, here is where I got the 20 minute number:
https://www.terraform.io/plugin/sdkv2/resources/retries-and-customizable-timeouts#default-timeouts-and-deadline-exceeded-errors

stvnjacobs (approved):

jriddle-linode (approved):
LGTM

PR #584, opened 2022-01-26
Use supported k8s version for lke_cluster doc

codybswaney:
1.17 is not a supported version of LKE, at least in us-east which supports only 1.21 and 1.22

Picked 1.21 because 1.21 is referenced earlier in this doc (lines 20 and 36).


LBGarber (approved):

LBGarber:
Looks good! Thanks for the contribution 🙂

PR #593, opened 2022-02-17
new: Add `resize_disk` field to resize implicit disks with Linode type

LBGarber:
This pull request adds an optional `resize_disk` field that triggers an automatic disk resize for Linode type changes.

This feature has the following constraints:
- Cannot be used on explicit disks
- Cannot be used on Linode downsizes


jriddle-linode (commented):
Looks good to me, as long as we force it to be one or the other we should be fine.

jriddle-linode (approved):

Issue #591, opened 2022-02-11
[Feature]: Change power state of linode


SebDanielsson:
### Description

After installing all the programs and restoring from backup it would be nice to be able to restart the linode with the Linode Terraform Provider.

I found a nice hack to achieve this but I would love to have this capability built in.
https://www.linode.com/community/questions/21394/how-can-i-use-terraform-to-change-the-power-state-of-a-linode

LBGarber:
Hello! Thanks for the request 🙂 

I can definitely see a bunch of use cases for being able to control the power state of a `linode_instance`. I'll add this to our internal backlog and let you know when we start looking into it further.

PR #594, opened 2022-02-17
new: Add booted field to linode_instance resource

LBGarber:
This pull request adds a `booted` field to the `linode_instance` resource.

This field has the following outcomes:
- If `booted` == `true`, the instance will be powered on if necessary
  - An explicit image, config, or disk must be defined else a validation error will be thrown 
- If `booted` == `false`, the instance will be shut down if necessary
- If `booted` is undefined, the power state of the Linode will not be managed by Terraform.

If the Linode is in a transition state (e.g. `shutting_down`, `rebooting`, etc.), the resource waits for the event to finish before performing the appropriate operation.
 
All tests are passing 🙂 

#591


jriddle-linode (approved):
LGTM

PR #597, opened 2022-02-24
Fix bad link in documentation for `linode_instance`

ilyasotkov:


LBGarber (approved):
Good catch! Thanks for the contribution 🙂

LBGarber (approved):

PR #643, opened 2022-05-10
new: Add test to measure API request volume

LBGarber:
This pull request adds a test to measure the volume of requests made to the API for parallel Linode creation. If the number of requests exceeds the `maxRequests` value, the test will fail.

This test also reports the frequency of the requests and the number of requests made:
```
[INFO] results from 12 linode parallel creation:
total requests: 186
frequency: ~1.823047 requests/second
```

Depends on https://github.com/linode/linodego/pull/250


jriddle-linode:
other pr approved, pre-approving this one since all you need to do is merge it

jriddle-linode (approved):

PR #650, opened 2022-05-22
docs: Improving documentation for booted flag

venkatamutyala:
docs: Improving documentation for booted flag to help save some testing for future users


LBGarber (approved):
Looks good! This is a lot more concise than the previous wording 👍

PR #687, opened 2022-08-08
ref: Use EventPoller object for event polling operations

LBGarber:
This pull request replaces all uses of `client.WaitForEventFinished(...)` with the equivalent `EventPoller` logic. This should significantly reduce the number of context deadline errors caused by time desyncs.


LBGarber:
All non-flapping tests are passing

jriddle-linode (approved):

PR #723, opened 2022-11-17
Rename website docs from .md -> .markdown

srust:
Other official terraform providers use .markdown as the extension for web docs files.

For example:

https://github.com/hashicorp/terraform-provider-aws/tree/main/website/docs/r
https://github.com/hashicorp/terraform-provider-azurerm/tree/main/website/docs/r
https://github.com/hashicorp/terraform-provider-google/tree/main/website/docs/r

There are external tools that parse this documentation and look for the .markdown extension specifically that are having issues with the linode provider because the extension is .md.

One such tool:

https://github.com/upbound/upjet
https://github.com/upbound/upjet/tree/main/cmd/scraper

This change renames the linode documentation from .md to .markdown file extensions.

Please let me know how/if I can help test this change to make sure the rename doesn't cause any problems.

Thanks!


lgarber-akamai:
@srust Thanks for the contribution! I'll look into testing this change to make sure it doesn't break our docs on the Terraform registry.

Is there any chance you could also rename the files in the `d` (data source) directory?

srust:
Thanks @lgarber-akamai ! Yes, just pushed a commit with the `d` rename as well.

I can squash the commits if you'd like.

lgarber-akamai (approved):
Nice, thanks @srust!

I can handle squashing the PR while merging. I've verified that the documentation still works as expected with the new extensions, so we should be good to go!

jriddle-linode (approved):

PR #725, opened 2022-11-18
fix: Bump Alpine 3.12 -> 3.16 in test suite

lgarber-akamai:
This change bumps the Alpine image version in tests that use recently deprecated versions. In the future, we may want to dynamically resolve these images.


lgarber-akamai:
/acctest sha=36cc3d869950bc25fc40dd78b0ce51e2e51b584d pkg=linode/networkingip

PR #733, opened 2022-12-27
new: add `available` field to snapshots and backups

jriddle-linode:
## 📝 Description

Add `available` field to backups and snapshots
Updated linodego version to 1.10

## ✔️ How to Test

`make PKG_NAME='linode/backup' testacc`


linode/backup/datasource_test.go
```diff
@@ -76,6 +77,7 @@ func TestAccDataSourceInstanceBackups_basic(t *testing.T) {
 					resource.TestCheckResourceAttrSet(resourceName, "current.0.id"),
 					resource.TestCheckResourceAttr(resourceName, "current.0.label", snapshotName),
 					resource.TestCheckResourceAttr(resourceName, "current.0.status", "successful"),
+					resource.TestCheckResourceAttr(resourceName, "current.0.available", "true"),
```
lgarber-akamai, line 80: Could we add a check here to make sure the instance `backups.0.available` field is true?

e.g.

```suggestion
					resource.TestCheckResourceAttr(resourceName, "current.0.available", "true"),
					resource.TestCheckResourceAttr("linode_instance.foobar", "backups.0.available", "true"),
```

lgarber-akamai (changes requested):
Looks great! Just one small thing

lgarber-akamai (approved):

PR #742, opened 2023-01-10
new: adding datasource for Instance IPs

jriddle-linode:
## 📝 Description

adding Datasource for: https://www.linode.com/docs/api/linode-instances/#networking-information-list

## ✔️ How to Test

`make PKG_NAME='linode/instanceip' TESTARGS='-run TestAccDataSourceInstanceIP_basic' testacc`

### DIFF
--- a/linode/instanceip/flatten.go
+++ b/linode/instanceip/flatten.go
@@ -1,0 +1,69 @@
+package instanceip
+
+import (
+	"github.com/linode/linodego"
+)
+
+func flattenIPv4(network *linodego.InstanceIPv4Response) map[string]interface{} {
+	result := make(map[string]interface{})
+
+	result["private"] = flattenIPs(network.Private)
+	result["public"] = flattenIPs(network.Public)
+	result["reserved"] = flattenIPs(network.Reserved)
+	result["shared"] = flattenIPs(network.Shared)
+
+	return result
+}
+
+func flattenIPv6(network *linodego.InstanceIPv6Response) map[string]interface{} {
+	result := make(map[string]interface{})
+
+	result["global"] = flattenGlobal(network.Global)
+	result["link_local"] = flattenIP(network.LinkLocal)
+	result["slaac"] = flattenIP(network.SLAAC)
+
+	return result
+}
+
+func flattenIP(network *linodego.InstanceIP) map[string]interface{} {
+	result := make(map[string]interface{})
+
+	result["address"] = network.Address
+	result["gateway"] = network.Gateway
+	result["prefix"] = network.Prefix
+	result["rdns"] = network.RDNS
+	result["region"] = network.Region
+	result["subnet_mask"] = network.SubnetMask
+	result["type"] = network.Type
+
+	return result
+}
+
+func flattenIPs(network []*linodego.InstanceIP) map[string]interface{} {
+	result := make(map[string]interface{})
+
+	for _, net := range network {
+		result["address"] = net.Address
+		result["gateway"] = net.Gateway
+		result["prefix"] = net.Prefix
+		result["rdns"] = net.RDNS
+		result["region"] = net.Region
+		result["subnet_mask"] = net.SubnetMask
+		result["type"] = net.Type
+	}
+
+	return result
+}
+
+func flattenGlobal(network []linodego.IPv6Range) map[string]interface{} {
+	result := make(map[string]interface{})
+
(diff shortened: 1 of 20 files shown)


linode/provider.go
```diff
@@ -170,6 +170,7 @@ func Provider() *schema.Provider {
 			"linode_instance_backups":       backup.DataSource(),
 			"linode_instance_type":          instancetype.DataSource(),
 			"linode_instance_types":         instancetypes.DataSource(),
+			"linode_instance_ip":            instanceip.DataSource(),
```
lgarber-akamai, line 173: We might want to make this into `linode_instance_networking` since `linode_instance_ip` refers to a single instance of a Linode IP. Definitely open to discussion though 🙂

linode/instanceip/flatten.go
```diff
@@ -0,0 +1,69 @@
...
+
+func flattenIPv4(network *linodego.InstanceIPv4Response) map[string]interface{} {
+	result := make(map[string]interface{})
+
+	result["private"] = flattenIPs(network.Private)
+	result["public"] = flattenIPs(network.Public)
+	result["reserved"] = flattenIPs(network.Reserved)
+	result["shared"] = flattenIPs(network.Shared)
+
+	return result
```
lgarber-akamai, line 15: Due to SDKv2 limitations this will need to be returned as a list.
e.g.
```suggestion
	return []map[string]any{result}
```

linode/instanceip/flatten.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func flattenIPv6(network *linodego.InstanceIPv6Response) map[string]interface{} {
+	result := make(map[string]interface{})
+
+	result["global"] = flattenGlobal(network.Global)
+	result["link_local"] = flattenIP(network.LinkLocal)
+	result["slaac"] = flattenIP(network.SLAAC)
+
+	return result
```
lgarber-akamai, line 25: Same as above

linode/instanceip/flatten.go
```diff
@@ -0,0 +1,69 @@
...
+		result["address"] = net.Address
+		result["gateway"] = net.Gateway
+		result["prefix"] = net.Prefix
+		result["rdns"] = net.RDNS
+		result["region"] = net.Region
+		result["subnet_mask"] = net.SubnetMask
+		result["type"] = net.Type
+	}
+
+	return result
```
lgarber-akamai, line 55: Assuming this is supposed to return a list, you might be able to use your existing flattenIP function:

```suggestion
func flattenIPs(network []*linodego.InstanceIP) []map[string]interface{} {
	result := make([]map[string]interface{}, len(network))

	for i, net := range network {
		result[i] = flattenIP(net)
	}

	return result
```

linode/instanceip/flatten.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, net := range network {
+		result["prefix"] = net.Prefix
+		result["range"] = net.Range
+		result["region"] = net.Region
+		result["route_target"] = net.RouteTarget
+	}
+
+	return result
+}
```
lgarber-akamai, line 69: Same as above, looks like it should be returning a list

lgarber-akamai (changes requested):
Nice work so far! General logic looks good; just a bit of SDKv2 jank to work around :+1:

linode/instanceip/tmpl/data_basic.gotf
```diff
@@ -0,0 +1,15 @@
+{{ define "instance_ip_data_basic" }}
```
lgarber-akamai, line 1:
```suggestion
{{ define "instance_networking_data_basic" }}
```

linode/instanceip/tmpl/template.go
```diff
@@ -21,6 +21,14 @@ func Basic(t *testing.T, instanceLabel, region string, applyImmediately bool) st
 		})
 }
 
+func DataBasic(t *testing.T, instanceLabel, region string) string {
+	return acceptance.ExecuteTemplate(t,
+		"instance_ip_data_basic", TemplateData{
```
lgarber-akamai, line 26: instance_networking_data_basic

lgarber-akamai (changes requested):
Implementation looks great! Could we move this logic over to an `instancenetworking` package to keep it consistent with the data source name?

PR #757, opened 2023-02-03
Added account_settings datasource

ezilber-akamai:
## 📝 Description

Added `linode_account_settings` datasource.


ezilber-akamai:
/acctest sha=d459a6a49f172c5d86b14db68f267ddafffe612c pkg=linode/accountsettings

website/docs/d/account_settings.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+
+The following example shows how one might use this data source to access information about Linode account settings.
+
+```hcl
+data "linode_account_settings" "example" {}
+```
+
+## Attributes Reference
+
+Each Linode LKE Version will be stored in the `versions` attribute and will export the following attributes:
```
lgarber-akamai, line 23: I think you can drop/edit this line 👍

lgarber-akamai (changes requested):
Great work! Just one small change

website/docs/d/account_settings.html.markdown
ezilber-akamai: Good catch! I missed that when I copy-pasted :)

lgarber-akamai (approved):

zliang-akamai (approved):

PR #760, opened 2023-02-09
Added `linode_account_settings` resource

ezilber-akamai:
## 📝 Description

Added a `linode_account_settings` resource to allow users to modify their account settings through the terraform provider.


jriddle-linode (approved):

linode/accountsettings/resource_test.go
```diff
@@ -0,0 +1,69 @@
...
+		PreCheck:  func() { acceptance.PreCheck(t) },
+		Providers: acceptance.TestAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: tmpl.Updates(t, updatedLongviewPlan, updatedBackupsEnabled, updatedNetworkHelper),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "longview_subscription", updatedLongviewPlan),
+					resource.TestCheckResourceAttr(resourceName, "backups_enabled", strconv.FormatBool(updatedBackupsEnabled)),
+					resource.TestCheckResourceAttr(resourceName, "network_helper", strconv.FormatBool(updatedNetworkHelper)),
+				),
```
lgarber-akamai, line 65: We might want an additional test step here to revert settings back to their original settings.

e.g.
```
tmpl.Updates(t, currLongviewPlan, currBackupsEnabled, currNetworkHelper)
```

lgarber-akamai (changes requested):
Great work! Just one testing suggestion

linode/accountsettings/resource_test.go
ezilber-akamai: Good idea! Should I test that the second update went through successfully afterwards or is it fine to just run the update?

lgarber-akamai: Probably good to validate the results just in case :+1:

lgarber-akamai (approved):
Nice work!

PR #766, opened 2023-02-15
Updated `linode_lke_cluster` resource to wait for nodes to be ready before returning from scale-up operation

ezilber-akamai:
## 📝 Description

Updated `linode_lke_cluster` resource to wait for nodes to be ready before returning from scale-up operation

### DIFF
--- a/linode/lke/resource.go
+++ b/linode/lke/resource.go
@@ -203,7 +203,6 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
 
 	for _, createOpts := range updates.ToCreate {
 		pool, err := client.CreateLKENodePool(ctx, id, createOpts)
-
 		if err != nil {
 			return diag.Errorf("failed to create LKE Cluster %d Pool: %s", id, err)
 		}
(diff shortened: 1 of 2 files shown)


ezilber-akamai:
/acctest sha=ffa6033aca563a636bc1d8df88a166f84ead13e4 pkg=linode/lke

jriddle-linode (approved):

linode/lke/resource.go
```diff
@@ -209,6 +216,12 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
 		}
 	}
 
+	for _, id := range updatedIds {
+		client.WaitForLKEClusterConditions(ctx, id, linodego.LKEClusterPollOptions{
```
lgarber-akamai, line 220: I was thinking about this earlier and I think it might actually be better to have a separate polling function to check if each node pool has at least one `ready` node (from the Linode API). It looks like the current implementation would pass automatically if any nodes that were previous provisioned remained since this function checks for only one `ready` node.

Fortunately it looks like we already have a function for doing this here 🎉 : https://github.com/linode/terraform-provider-linode/blob/dev/linode/lke/cluster.go#L181

ezilber-akamai:
/acctest sha=4f7c9e0e38ef838193b8617d5ae240d2affe95ab pkg=linode/lke

lgarber-akamai (approved):
Great work!

PR #775, opened 2023-03-08
fix: Make login-related tests opt-in

lgarber-akamai:
## 📝 Description

This change makes the login-related tests opt-in. This is necessary as this endpoint only returns logins within the past 90 days and we cannot guarantee that our automated testing account always meets this condition.

Additionally, this change adds error handling when listing logins in test cases.

## ✔️ How to Test

```
make PKG_NAME=linode/accountlogins testacc
make PKG_NAME=linode/accountlogins testacc
```

### DIFF
--- a/linode/accountlogin/datasource_test.go
+++ b/linode/accountlogin/datasource_test.go
@@ -12,6 +12,7 @@ import (
 )
 
 func TestAccDataSourceLinodeAccountLogin_basic(t *testing.T) {
+	acceptance.OptInTest(t)
 	t.Parallel()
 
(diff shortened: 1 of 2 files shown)


lgarber-akamai:
/acctest sha=32ce3165a4917b52d5d3bfccc9126e4592419314 pkg=linode/accountlogins

lgarber-akamai:
/acctest sha=2684317654e1df6740d607f87e530eed53c5a55d pkg=linode/accountlogins

jriddle-linode (approved):

PR #776, opened 2023-03-15
fix: Increase complexity for root passwords used in tests (again)

lgarber-akamai:
## 📝 Description

This change alters some acctest suite passwords to be compliant with the raised root password requirements enforced by the API.

## ✔️ How to Test

```
make testacc
```

### DIFF
--- a/linode/backup/datasource_test.go
+++ b/linode/backup/datasource_test.go
@@ -96,7 +96,7 @@ resource "linode_instance" "foobar" {
 	type = "g6-nanode-1"
 	image = "linode/alpine3.15"
 	region = "%s"
-	root_pass = "terraform-test"
+	root_pass = "myr00tp@ssw0rd!!!"
 	swap_size = 256
 	backups_enabled = true
 }`, label, region)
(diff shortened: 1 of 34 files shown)


lgarber-akamai:
/acctest sha=ca6c4f97390007e3b8084294b5f35f395b534a03 pkg=linode/...

lgarber-akamai:
/acctest sha=439f067fd3899758e0d9f86f79b82e9d5a492d89 pkg=linode/...

yec-akamai (approved):
LGTM! Just curious, how do we know that the complexity for the root passwords need to be increased? Should we keep us updated with any docs/announcement, or did we start to see the api calls failed?

zliang-akamai (approved):

lgarber-akamai:
> LGTM! Just curious, how do we know that the complexity for the root passwords need to be increased? Should we keep us updated with any docs/announcement, or did we start to see the api calls failed?

Great question! The raised password requirements by the API were planned, but in this case I just noticed some related failures in our integration test suite

PR #786, opened 2023-04-11
Handle defaults in config function

zliang-akamai:
## 📝 Description

This is necessary because the Terraform Framework dropped the support of default fields in the schema. And in order to use muxing server with both SDKv2 provider and framework provider, it is required that usage of default fields in SDKv2 provider to be moved to config function.

Issue reported and solved in HashiCorp's repo:
https://github.com/hashicorp/terraform-plugin-framework/issues/640

I separated this change from the main migration branch to make the review of this PR easier. Other migration stuff will be coming in another PR after merging of this one.

## ✔️ How to Test
`make testacc`


linode/provider.go
```diff
@@ -251,31 +233,90 @@ func Provider() *schema.Provider {
...
-		ConfigPath:    d.Get("config_path").(string),
-		ConfigProfile: d.Get("config_profile").(string),
-
 		SkipInstanceReadyPoll:  d.Get("skip_instance_ready_poll").(bool),
 		SkipInstanceDeletePoll: d.Get("skip_instance_delete_poll").(bool),
 
 		DisableInternalCache: d.Get("disable_internal_cache").(bool),
 
 		MinRetryDelayMilliseconds: d.Get("min_retry_delay_ms").(int),
 		MaxRetryDelayMilliseconds: d.Get("max_retry_delay_ms").(int),
```
yec-akamai, line 315: do we need to move these to `handleDefault` as well?

yec-akamai (approved):

linode/provider.go
zliang-akamai: `d.Get()` would return the default values of that golang type if the key exist in schema but not set by the user.
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema#ResourceData.Get

The default value of`SkipInstanceReadyPoll`, `SkipInstanceDeletePoll`, `DisableInternalCache` we set in the schema were `false`, which is also the default value of `bool` in go lang. Which means when you define a `bool` variable without assigning any value to it in go lang, it would be a `false`.

So, if we don't set a default value of `false` to them on Terraform application layer, the `d.Get()` will still return the desired default value when the user don't configure this attribute. So, having a default value for them is not necessary.

zliang-akamai:
/acctest sha=b5361480e428b4f7550876b2fa5f676f5e0eb333 pkg=linode/...

zliang-akamai:
/acctest sha=ee2212c04fb39a99d9cc1f1ec21bbcd8baf903a5 pkg=linode/...

PR #791, opened 2023-04-12
Framework Migration

zliang-akamai:
## 📝 Description

1. Migrating Terraform SDK v2 provider to be Terraform Framework Provider.

2. Creating mux server for supporting both SDK v2 provider and the framework provider concurrently.

3. Migrate resource `token` to Framework provider's resource implementation as an example for the future migration work.

## ✔️ How to Test

Test the migrated resource:
`make PKG_NAME="linode/token" testacc`

Test everything else:
`make testacc`


linode/framework_provider.go
```diff
@@ -0,0 +1,111 @@
...
+		ProviderVersion: version,
+	}
+}
+
+func (p *FrameworkProvider) Metadata(
+	ctx context.Context,
+	req provider.MetadataRequest,
+	resp *provider.MetadataResponse,
+) {
+	resp.TypeName = "Linode cloud"
```
lgarber-akamai, line 30: Judging by the framework documentation, it looks like we might just want this string to become something like `linode`: https://developer.hashicorp.com/terraform/plugin/framework/resources#metadata-method

linode/framework_provider_config.go
```diff
@@ -0,0 +1,302 @@
...
+	_, err := url.Parse(data.APIURL.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to parse the base API URL in the configuration",
+			err.Error(),
+		)
+	}
+}
+
+func GetIntFromEnv(
```
lgarber-akamai, line 77: I think we can flatten this function out a bit:

```go
func GetIntFromEnv(
	key string,
	defaultValue basetypes.Int64Value,
	diags *diag.Diagnostics,
) basetypes.Int64Value {
	envVarVal := os.Getenv(key)
	
	if envVarVal == "" {
		return defaultValue
	}
	
	intVal, err := strconv.ParseInt(envVarVal, 10, 64)
	if err != nil {
		diags.AddWarning(
			fmt.Sprintf(
				"Failed to parse the environment variable %v "+
					"to an integer. Will use default value: %v instead",
				key,
				defaultValue.ValueInt64(),
			),
			err.Error(),
		)
		
		return defaultValue
	}
	
	return types.Int64Value(intVal)
}
```

linode/framework_provider_config.go
```diff
@@ -0,0 +1,302 @@
...
+			)
+			result = defaultValue
+		} else {
+			result = types.Int64Value(intVal)
+		}
+	}
+	return result
+}
+
+func GetStringFromEnv(key string, defaultValue basetypes.StringValue) basetypes.StringValue {
```
lgarber-akamai, line 108: This can simplify down a bit:
```go
func GetStringFromEnv(key string, defaultValue basetypes.StringValue) basetypes.StringValue {
	envVarVal := os.Getenv(key)

	if envVarVal == "" {
		return defaultValue
	}
	
	return types.StringValue(envVarVal)
}
```

linode/token/framework_resource.go
```diff
@@ -0,0 +1,270 @@
...
+		updateOpts.Label = plannedTokenLabel
+		token, err = client.UpdateToken(ctx, token.ID, updateOpts)
+
+		if err != nil {
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Failed to update the token with id %v", tokenID),
+				err.Error(),
+			)
+			return
+		}
```
lgarber-akamai, line 233: To generalize this logic a bit and prevent nesting, could we use a variable to track whether the resource has been updated and call `client.UpdateToken(...)` from within another if statement. This is a stylistic change so feel free to ignore 🙂

linode/token/framework_validators.go
```diff
@@ -0,0 +1,47 @@
...
+			request.Path,
+			dtv.Description(ctx),
+			v,
+		))
+	}
+}
+
+func NewDateTimeStringValidator(format string) DateTimeStringValidator {
+	return DateTimeStringValidator{Format: format}
+}
```
lgarber-akamai, line 47: Could we move this into a helper package so we can re-use it in other resources?

linode/token/resource_test.go
```diff
@@ -54,8 +56,25 @@ func TestAccResourceToken_basic(t *testing.T) {
...
+				}
+
+				muxServer, err := tf5muxserver.NewMuxServer(ctx, providers...)
+				if err != nil {
+					return nil, err
+				}
+
+				return muxServer.ProviderServer(), nil
+			},
+		},
```
lgarber-akamai, line 77: Could this be moved to the `acceptance` package for re-use in other tests?

lgarber-akamai (changes requested):
Great work so far! Just a few suggestions

linode/token/framework_resource.go
```diff
@@ -0,0 +1,270 @@
...
+			fmt.Sprintf(
+				"Expected *http.Client, got: %T. Please report this issue to the provider developers.",
+				req.ProviderData,
+			),
+		)
+
+		return
+	}
+
+	r.client = meta.Client
```
lgarber-akamai, line 57: This might be useful as a helper function to be reused across all resources.

linode/framework_provider.go
zliang-akamai: Good catch, thank you!

linode/framework_provider_config.go
zliang-akamai: Thank you for the refactoring suggestion. The function is much more clean now.

linode/token/framework_validators.go
zliang-akamai: Moved. Thank you!

linode/token/framework_resource.go
zliang-akamai: Just wrapped the get meta logic into a helper function. Thank you for the suggestion!

linode/token/framework_resource.go
zliang-akamai: Pushed.

main.go
```diff
@@ -1,12 +1,51 @@
...
+	ctx := context.Background()
+
+	var debug bool
+
+	flag.BoolVar(&debug, "debug", false, "set to true to run the provider with support for debuggers like delve")
+	flag.Parse()
+
+	providers := []func() tfprotov5.ProviderServer{
+		providerserver.NewProtocol5(
+			linode.CreateFrameworkProvider(version.ProviderVersion),
```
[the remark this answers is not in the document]
lgarber-akamai: @zliang-akamai Yep! The `ProviderVersion` is injected into the `version` package at build-time. 

See: https://github.com/linode/terraform-provider-linode/blob/dev/.goreleaser.yml#L13

lgarber-akamai:
/acctest sha=3addf8af2acb3f50a5850ca31a7d9d235fddc45d pkg=linode/...

lgarber-akamai:
No idea if that will work but it's worth a try ^

zliang-akamai:
@lgarber-akamai 
We can also try token resource since it's the only resource placed in the framework provider.

zliang-akamai:
/acctest sha=3addf8af2acb3f50a5850ca31a7d9d235fddc45d pkg=linode/token

yec-akamai:
Can you add some description to this PR? It's a large change, will be difficult to understand and maintain in the future without any context.

lgarber-akamai (approved):
Looks like all relevant tests are passing 🎉

zliang-akamai:
> Can you add some description to this PR? It's a large change, will be difficult to understand and maintain in the future without any context.

Thank you! I just added the description.

ezilber-akamai (approved):
Good work! Test are passing locally 🥳

PR #801, opened 2023-05-01
Migrated `account` datasource to Framework

ezilber-akamai:
## 📝 Description

Migrated `account` datasource from SDKv2 to Framework

## ✔️ How to Test

`make PKG_NAME="linode/account" testacc`


ezilber-akamai:
/acctest sha=1a0fc2f8a7f03b09ce81d7b28cb7616ccf3b7d61 pkg=linode/account

linode/account/framework_datasource.go
```diff
@@ -0,0 +1,110 @@
...
+	data.Phone = types.StringValue(account.Phone)
+	data.City = types.StringValue(account.City)
+	data.State = types.StringValue(account.State)
+	data.Country = types.StringValue(account.Country)
+	data.Zip = types.StringValue(account.Zip)
+	data.Balance = types.Float64Value(float64(account.Balance))
+	data.ID = types.StringValue(account.Email)
+}
+
+func (d *Datasource) Configure(
```
yec-akamai, line 37: Is `Configure` necessary here? I know it's in the token migration example, but I'm only able to find it in the [provider migration](https://developer.hashicorp.com/terraform/plugin/framework/migrating/providers#framework-1) overview, instead of any data source or resource migration instructions.

linode/account/framework_datasource.go
```diff
@@ -0,0 +1,110 @@
...
+
+	account, err := client.GetAccount(ctx)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to get Account",
+			fmt.Sprintf(
+				"Error finding Linode Account: %s",
+				err.Error(),
+			),
+		)
```
zliang-akamai, line 104: I think `err.Error()` itself is enough for the second parameter of `resp.Diagnostics.AddError` function and it's unnecessary to add `"Error finding Linode Account:"` because we already said something similar in line 99. Do you think so?

linode/helper/resource_datasource_config.go
```diff
@@ -25,3 +26,23 @@ func GetMetaFromProviderData(
...
+			fmt.Sprintf(
+				"Expected *http.Client, got: %T. Please report this issue to the provider developers.",
+				req.ProviderData,
+			),
+		)
+		return nil
+	}
+
+	return meta
+}
```
zliang-akamai, line 48: Can we reuse `GetMetaFromProviderData` function? I can't see what's the difference

zliang-akamai:
Overall it looks good!

zliang-akamai:
We will need to merge this in order to run the acceptance test for this PR on GitHub.
https://github.com/linode/terraform-provider-linode/pull/800

zliang-akamai:
/acctest sha=7ced026f0c9bc1fd5f2f65b4f3cabc2102f7a541 pkg=linode/account

zliang-akamai (approved):

linode/account/framework_datasource.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (d *DataSource) Read(
+	ctx context.Context,
+	req datasource.ReadRequest,
+	resp *datasource.ReadResponse,
+) {
+	client := d.client
+
+	var data DataSourceModel
+
```
zliang-akamai, line 94:
```suggestion
	var data DataSourceModel
	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
```

I feel it might still be worth to read the config because we might have some config values for this data source in the future, and it probably won't hurt when there is not a config for now. But I am not very sure about this, what do you think?

ezilber-akamai: I think it doesn't hurt, I'll make the change.

zliang-akamai:
/acctest sha=5f10c0103a3162cc5e8b699fb7e95946a74dbcf7 pkg=linode/account

zliang-akamai (approved):

yec-akamai (approved):
Nice work!

ezilber-akamai:
/acctest sha=19f8c2a1e0b81b5375570c5bc354bd5d3e6c04c8 pkg=linode/account

PR #802, opened 2023-05-02
ref: Migrate "linode_stackscript" resource and data source to framework

lgarber-akamai:
## 📝 Description

This change migrates the SDKv2 implementations of `linode_stackscript` to terraform-plugin-framework and removes the old SDKv2 implementations.

All existing tests are currently passing.

## ✔️ How to Test

```
make PKG_NAME=linode/stackscript testacc
```

PR #805, opened 2023-05-03
Improved update method of token resource

zliang-akamai:
## 📝 Description

This improvement was inspired by @lgarber-akamai's PR https://github.com/linode/terraform-provider-linode/pull/802

Basically only to check the state of the resource before the update instead of using the API to check the resource remotely.

## ✔️ How to Test

`make PKG_NAME="linode/token" testacc`

### DIFF
--- a/linode/token/framework_resource.go
+++ b/linode/token/framework_resource.go
@@ -180,8 +180,7 @@ func (r *Resource) Update(
 	req resource.UpdateRequest,
 	resp *resource.UpdateResponse,
 ) {
-	var state ResourceModel
-	var plan ResourceModel
+	var plan, state ResourceModel
 
 	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
--- a/linode/token/framework_resource.go
+++ b/linode/token/framework_resource.go
@@ -185,9 +185,7 @@ func (r *Resource) Update(
 	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
 	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
 
-	resourceUpdated := !state.Label.Equal(plan.Label)
-
-	if resourceUpdated {
+	if !state.Label.Equal(plan.Label) {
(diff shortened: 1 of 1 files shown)


zliang-akamai:
/acctest sha=a97f602689d83a44d09a02b24000f13f8ae0b252 pkg=linode/token

linode/token/framework_resource.go
```diff
@@ -180,48 +180,32 @@ func (r *Resource) Update(
...
-	var data ResourceModel
-	var tokenIDString string
-	resp.Diagnostics.Append(
-		req.State.GetAttribute(ctx, path.Root("id"), &tokenIDString)...,
-	)
-	if resp.Diagnostics.HasError() {
-		return
-	}
+	var state ResourceModel
+	var plan ResourceModel
```
yec-akamai, line 184: nit:
```suggestion
	var plan, state ResourceModel
```

linode/token/framework_resource.go
```diff
@@ -180,48 +180,32 @@ func (r *Resource) Update(
...
-		resp.Diagnostics.AddError(
-			fmt.Sprintf("Failed to get the token with id %v", tokenID),
-			err.Error(),
-		)
-		return
-	}
-
-	updateOpts := token.GetUpdateOptions()
-	plannedTokenLabel := data.Label.ValueString()
+	if resourceUpdated {
```
yec-akamai, line 191: I think you can directly compare these two values instead of storing it as a flag:
```suggestion
	if !state.Label.Equal(plan.Label) {
	    ...
	}
```

yec-akamai (approved):
Looks good! Approved pending on small suggestions.

linode/token/framework_resource.go
zliang-akamai: Thank you for the suggestion!

zliang-akamai:
/acctest sha=67103ae2d3953d53369fc6563b5d5bce5e9dcb52 pkg=linode/token

zliang-akamai:
/acctest sha=0f7dc4df2a64824274a0def0565a96d05d8c7806 pkg=linode/token

jriddle-linode (approved):

PR #807, opened 2023-05-04
feat: ipv6_address

dennisme:
## 📝 Description

Adds ipv6_address with the prefix stripped. 

## ✔️ How to Test

## 📷 Preview


lgarber-akamai (changes requested):
Thanks for the contribution! Could you also add the corresponding documentation entry in `website/docs/r/instance.html.markdown`?

dennisme:
> Thanks for the contribution! Could you also add the corresponding documentation entry in `website/docs/r/instance.html.markdown`?

Absolutely! I'll also update my PR comment with more details here once we know more! For now this is just a test to help us rule out the linode provider. We are getting terraform crashes when trying to create AAAA records via the Akamai provider.

zliang-akamai:
@lgarber-akamai Does the doc look good now?
@dennisme I think Akamai developers are working on the fix now. Would you considered this PR as ready so we can proceed with it?

PR #829, opened 2023-05-15
Migrated `instancenetworking` datasource to Framework

ezilber-akamai:
## 📝 Description

Migrated `instancenetworking` datasource to Framework

## ✔️ How to Test

`make PKG_NAME="linode/instancenetworking" testacc`

Ticket: TPT-1983


ezilber-akamai:
/acctest sha=7656432407ead1db16591a1ca5394ed00e1f1e94 pkg=linode/instancenetworking

linode/instancenetworking/flatten.go
yec-akamai, line 1: Probably better to rename this file to be `framework_models.go`, and move `parseInstanceIPAddressResponse` and `DataSourceModel` into it as well. So we can have one place focusing on parsing the data model and simplify the datasource file.

ezilber-akamai:
/acctest sha=6aa1a1d5571d090717d04d43a6a4092767462c3f pkg=linode/instancenetworking

ezilber-akamai:
/acctest sha=6ec92ecebf6b5088679660d32e0e41b646067a77 pkg=linode/instancenetworking

lgarber-akamai (approved):
Nice work!

linode/instancenetworking/framework_datasource.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to get Instance Networking Information: ", err.Error(),
+		)
+		return
+	}
+
+	data.parseInstanceIPAddressResponse(ctx, netInfo)
+
+	resp.Diagnostics.Append(data.parseInstanceIPAddressResponse(ctx, netInfo)...)
```
zliang-akamai, line 77:
```suggestion
	resp.Diagnostics.Append(data.parseInstanceIPAddressResponse(ctx, netInfo)...)
```

I think you would only need to call `data.parseInstanceIPAddressResponse` once. Or is there any reason to call it twice?

yec-akamai (approved):
Looks good!

linode/instancenetworking/framework_datasource_schema.go
```diff
@@ -0,0 +1,68 @@
...
+	AttrTypes: map[string]attr.Type{
+		"address":     types.StringType,
+		"gateway":     types.StringType,
+		"prefix":      types.Int64Type,
+		"rdns":        types.StringType,
+		"region":      types.StringType,
+		"subnet_mask": types.StringType,
+		"type":        types.StringType,
+	},
+}
```
zliang-akamai, line 19: In the latest API, the response include an extra attribute `public`. Although we have the key values to show whether it's public (e.g., `ipv4.public` only contains public IP and `ipv4.private` only contains private IP), can we also include it here for user's convenient?

linode/instancenetworking/framework_datasource.go
ezilber-akamai: You're right. Just a copy-paste mistake. I will fix this!

zliang-akamai:
Others look good! Nice work!

linode/instancenetworking/framework_datasource_schema.go
ezilber-akamai: I don't see an issue with keeping the datasource schema aligned with the current state of the API. I also noticed that I was missing `LinodeID` so I added that as well. Does anyone else have any thoughts?

ezilber-akamai:
/acctest sha=1e48fcd45f6310a7ab1d03ebd4c77ac7094a90b2 pkg=linode/instancenetworking

zliang-akamai (approved):

PR #834, opened 2023-05-18
TPT 1976: Migrate `firewall` data source to terraform framework

yec-akamai:
## 📝 Description

Migrate the `firewall` datasource from SDKv2 to terraform framework, and remove the SDKv2 implementations.

## ✔️ How to Test

make PKG_NAME="linode/firewall" testacc


yec-akamai:
/acctest sha=2dc8fc41ec4d68b94ef2602aa8118ceeee8ca8db pkg=linode/firewall

ezilber-akamai (approved):
LGTM!

linode/firewall/framework_models.go
```diff
@@ -0,0 +1,168 @@
...
+	Status         types.String `tfsdk:"status"`
+}
+
+func (data *FirewallModel) parseComputedAttributes(
+	ctx context.Context,
+	firewall *linodego.Firewall,
+	rules *linodego.FirewallRuleSet,
+	devices []linodego.FirewallDevice,
+) diag.Diagnostics {
+	data.ID = types.Int64Value(int64(firewall.ID))
```
zliang-akamai, line 35: I think ID isn't computed, right?

linode/firewall/framework_models.go
```diff
@@ -0,0 +1,168 @@
...
+	data.Disabled = types.BoolValue(firewall.Status == linodego.FirewallDisabled)
+	data.Status = types.StringValue(string(firewall.Status))
+
+	linodes, diag := types.SetValueFrom(ctx, types.Int64Type, parseFirewallLinodes(devices))
```
zliang-akamai, line 39: Can you consider rename all `diag` to be `diags` if they are `type Diagnostics []Diagnostic`? I think naming them as `diag` can make it confused with a single `Diagnostic` object, right?

linode/firewall/framework_models.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (data *FirewallModel) parseComputedAttributes(
+	ctx context.Context,
+	firewall *linodego.Firewall,
+	rules *linodego.FirewallRuleSet,
+	devices []linodego.FirewallDevice,
+) diag.Diagnostics {
+	data.ID = types.Int64Value(int64(firewall.ID))
+	data.Disabled = types.BoolValue(firewall.Status == linodego.FirewallDisabled)
+	data.Status = types.StringValue(string(firewall.Status))
```
zliang-akamai, line 37: If this function will be shared with resource in the future, I think `status` is also non-computed, as defined in the original resource schema.

linode/firewall/framework_models.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (data *FirewallModel) parseComputedAttributes(
+	ctx context.Context,
+	firewall *linodego.Firewall,
+	rules *linodego.FirewallRuleSet,
+	devices []linodego.FirewallDevice,
+) diag.Diagnostics {
+	data.ID = types.Int64Value(int64(firewall.ID))
+	data.Disabled = types.BoolValue(firewall.Status == linodego.FirewallDisabled)
```
zliang-akamai, line 36: If this function will be shared with resource in the future, I think `disabled` is also non-computed, as defined in the original resource schema.

linode/firewall/framework_models.go
```diff
@@ -0,0 +1,168 @@
...
+
+func parseFirewallLinodes(devices []linodego.FirewallDevice) []int {
+	linodes := make([]int, 0, len(devices))
+	for _, device := range devices {
+		if device.Entity.Type == linodego.FirewallDeviceLinode {
+			linodes = append(linodes, device.Entity.ID)
+		}
+	}
+	return linodes
+}
```
zliang-akamai, line 139: Since this function wasn't changed, can we still use `flattenFirewallLinodes` instead of defining it again?

linode/firewall/framework_models.go
yec-akamai: ID will need to be computed for the resource. https://github.com/linode/terraform-provider-linode/blob/dev/linode/firewall/schema_resource.go#L97 Id is not included in the sdkv2 resource schema, however, when we migrate it to the framework, we have to explicitly add it in the resource schema and mark it computed.

linode/firewall/framework_models.go
yec-akamai: I put it here because it's computed in the datasource schema. https://github.com/linode/terraform-provider-linode/blob/dev/linode/firewall/schema_datasource.go#L67 So when we decide whether we should put it in computed parsing or not, do we only consider if it's computed in the resource schema?

linode/firewall/framework_models.go
yec-akamai: I'm thinking we may want to remove the `helpers.go` entirely after migrating both, so I copied it here. Leave the `flattenFirewallLinodes` function in the helpers for now to make sure resource still can work.

zliang-akamai: Since firewall module contains both resource and data source, I think we may keep the helpers that could be used to share some common functions?

yec-akamai: My understanding is the functions in the helpers.go currently can all be modified and moved to `framework_models` later when doing resource migration.

linode/firewall/framework_models.go
yec-akamai: Fixed. Thanks for catching it.

linode/firewall/framework_models.go
zliang-akamai: Ah, yes. You are right.

linode/firewall/framework_models.go
zliang-akamai: I think for we split the parse function into two parts, one for computed and another for non-computed attributes because in the resource, sometimes we don't want to edit the non-computed attributes defined by the resource schema. So I think wether a attribute should be placed in parsing computed or non-computed function is decided by the resource schema. You can take a look at the [stackscript module](https://github.com/linode/terraform-provider-linode/tree/dev/linode/stackscript) where we migrated both resource and data source. Although some of the attributes are computed in the data source schema and non-computed in resource schema, they are still in parse non-computed function because they are non-computed in the resource schema. 

For example, `label`

https://github.com/linode/terraform-provider-linode/blob/dev/linode/stackscript/framework_models.go#L40

https://github.com/linode/terraform-provider-linode/blob/dev/linode/stackscript/framework_datasource_schema.go#L15-L18

https://github.com/linode/terraform-provider-linode/blob/dev/linode/stackscript/framework_datasource_schema.go#L15-L18

linode/firewall/framework_models.go
zliang-akamai: > we only consider if it's computed in the resource schema

I think this would be the way.

linode/firewall/framework_models.go
yec-akamai: I see. We probably should consider rename these computed vs non-computed functions once we have a plan for the resource migration in the future. The current naming is a bit confusing.

zliang-akamai:
/acctest sha=06abdad1e15ec5dac2e34b53e043f934faa4374a pkg=linode/firewall

zliang-akamai (approved):

PR #838, opened 2023-05-19
new: Implement Linode API fitler framework for terraform-plugin-framework; migrate `linode_regions` data source

lgarber-akamai:
## 📝 Description

This change implements an equivalent Linode API filter framework for filter data sources that is compatible with terraform-plugin-framework.

An example implementation using the filter framework can be seen in the `linode_regions` data source.


lgarber-akamai:
/acctest sha=a533e4514060e1cd00627233abd35e0913d81154 pkg=linode/helper/frameworkfilter

lgarber-akamai:
/acctest sha=a533e4514060e1cd00627233abd35e0913d81154 pkg=linode/regions

linode/helper/frameworkfilter/latest.go
```diff
@@ -0,0 +1,126 @@
...
+		val = reflect.Indirect(val)
+	}
+
+	// Check the type
+	if val.Kind() != reflect.String {
+		return nil, diag.NewErrorDiagnostic(
+			"Field has incorrect type",
+			fmt.Sprintf("Field %s is not type string", attr),
+		)
+	}
```
yec-akamai, line 114: nit: we can probably wrap these checking/validating steps into a helper function. They are reused in both `getCreatedTime` and `getVersion.`

yec-akamai (commented):
Looks good! The code comments are really helpful for understanding each step.

go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/go-plugin v1.4.9 // indirect
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
-	github.com/hashicorp/go-version v1.6.0 // indirect
+	github.com/hashicorp/go-version v1.6.0
```
zliang-akamai, line 36: Can we move this to the [above section](https://github.com/linode/terraform-provider-linode/pull/838/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R3-R12)?

linode/regions/framework_datasource_schema.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+var frameworkDataSourceSchema = schema.Schema{
+	Attributes: map[string]schema.Attribute{
+		"id": schema.StringAttribute{
+			Description: "The data source's unique ID.",
+			Computed:    true,
+		},
+	},
+	Blocks: map[string]schema.Block{
```
zliang-akamai, line 22: We may want to switch to non-block normal list style as discussed in the last Friday.

lgarber-akamai: Good point! I think the main advantage to this approach is that descriptions, etc. can be embedded in nested structures, which can come in handy for [automatic code completion](https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform).

That said, I think consistency is key so I'm definitely open to any feedback on this approach. Does anyone else have any thoughts?

yec-akamai: Do we need to change the implementation of the filter schema, if we switch to a non-block style?

lgarber-akamai: @yec-akamai Assuming I'm understanding the question correctly,  I believe the implementation would change a good bit if we switched to non-block style schema for nested objects. Fortunately it shouldn't affect any existing data sources 👍

yec-akamai: So if we want to stick to non-block, we will switch to use a similar style as we do for the nested list currently, right? personally I think having the nested object here to take the advantage of descriptions and more details is better. Probably limit the usage for filter needed schemas only, will be helpful for consistency?

yec-akamai (approved):

linode/regions/framework_datasource_schema.go
lgarber-akamai: > So if we want to stick to non-block, we will switch to use a similar style as we do for the nested list currently, right?
Yep, the alternative to the current block layout would be using the nested list structure that we use in our already migrated data sources.

> personally I think having the nested object here to take the advantage of descriptions and more details is better.
After giving it some thought I agree with this. Being able to include descriptions, etc. in nested fields can definitely come in handy with these data sources.

> Probably limit the usage for filter needed schemas only, will be helpful for consistency?
I think there's room for discussion on this but generally I think scoping this to filter schemas should be fine 👍

lgarber-akamai: > So if we want to stick to non-block, we will switch to use a similar style as we do for the nested list currently, right?

Yep, the alternative to the current block layout would be using the nested list structure that we use in our already migrated data sources.

> personally I think having the nested object here to take the advantage of descriptions and more details is better.

After giving it some thought I agree with this. Being able to include descriptions, etc. in nested fields can definitely come in handy with these data sources.

> Probably limit the usage for filter needed schemas only, will be helpful for consistency?

I think there's room for discussion on this but generally I think scoping this to filter schemas should be fine 👍

PR #841, opened 2023-05-22
Checking the permission by API

zliang-akamai:


zliang-akamai:
Testing result in personal repo:
https://github.com/zliang-akamai/terraform-provider-linode/pull/2

PR #850, opened 2023-05-25
TPT 2004: Migrate volume datasource

zliang-akamai:
## 📝 Description

Migrated Linode volume module data source to be using tf framework.

## ✔️ How to Test

`make PKG_NAME="linode/volume" testacc`

Note that the test case was updated to align with the change. Previously in SDKv2, there was not null, so when the volume is not attached to any linode, the `linode_id` attribute is `"0"` instead of not exist. And the migrated version in this PR would not set the linode_id in the same situation.


linode/volume/framework_model.go
```diff
@@ -0,0 +1,64 @@
...
+	data.ID = types.Int64Value(int64(volume.ID))
+	data.Status = types.StringValue(string(volume.Status))
+	data.Region = types.StringValue(volume.Region)
+	data.Size = types.Int64Value(int64(volume.Size))
+
+	if volume.LinodeID != nil {
+		data.LinodeID = types.Int64Value(int64(*volume.LinodeID))
+	} else {
+		data.LinodeID = types.Int64Null()
+	}
```
lgarber-akamai, line 40: Just a heads up that this is _technically_ a breaking change:
```terraform
data "linode_volume" "test" {
  id = 1281734
}

output "is_detached" {
  value = data.linode_volume.test.linode_id == 0 // evaluates to false for detached linodes
}
```

zliang-akamai: Breaking changes removed, thanks!

ezilber-akamai (approved):

lgarber-akamai (changes requested):
One small thing and we'll be good to go!

linode/volume/framework_model.go
```diff
@@ -0,0 +1,66 @@
...
+	data.Region = types.StringValue(volume.Region)
+	data.Size = types.Int64Value(int64(volume.Size))
+
+	// Future breaking change:
+	// if volume.LinodeID != nil {
+	// 	data.LinodeID = types.Int64Value(int64(*volume.LinodeID))
+	// } else {
+	// 	data.LinodeID = types.Int64Null()
+	// }
+	data.LinodeID = types.Int64Value(int64(*volume.LinodeID))
```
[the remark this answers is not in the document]
zliang-akamai: Ah thanks. I thought Golang would automatically deref nil pointer to be the default value of the type...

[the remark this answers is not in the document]
yec-akamai: I build a helper function to return a default value for float type https://github.com/linode/terraform-provider-linode/blob/framework-migration-phase-1/linode/helper/default_value.go#L9. Maybe you can build a similar one for int64 and we can reuse it in other data source.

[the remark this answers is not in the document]
zliang-akamai: @yec-akamai good idea!

yec-akamai (approved):
Looks good!

linode/helper/default_value.go
```diff
@@ -13,3 +13,11 @@ func Float64PointerValueWithDefault(value *float64) basetypes.Float64Value {
 		return types.Float64Value(0)
 	}
 }
+
```
yec-akamai, line 16: nit: can we have a quick comment to describe the helper function?

zliang-akamai: Fixed, thanks!

lgarber-akamai (approved):

PR #861, opened 2023-06-01
Migrated `user` datasource to framework

ezilber-akamai-zz:
## 📝 Description

Migrated `user` datasource to framework

## ✔️ How to Test

`make PKG_NAME="linode/user" testacc`

Ticket: TPT-2002

### DIFF
--- a/website/docs/d/user.html.markdown
+++ b/website/docs/d/user.html.markdown
@@ -20,6 +20,18 @@ data "linode_user" "foo" {
 }
 ```
 
+The following example shows a sample grant.
+
+```hcl
+"domain": [
+  {
+    "id": 123,
+    "label": "example-entity",
+    "permissions": "read_only"
+  }
+]
+```
+
 ## Argument Reference
 
 The following arguments are supported:


ezilber-akamai:
/acctest sha=6508e505bbd6a3d63f5742f9e5f1916790f82d44 pkg=linode/user

linode/user/framework_datasource.go
```diff
@@ -0,0 +1,110 @@
...
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+	} else {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Linode User with username %s was not found", data.Username.ValueString()), "",
+		)
+		return
+	}
```
lgarber-akamai, line 109: We can tidy this logic up a tiny bit

```suggestion
	if user.Username == "" {
		resp.Diagnostics.AddError(
			fmt.Sprintf("Linode User with username %s was not found", data.Username.ValueString()), "",
		)
		return
	}

	resp.Diagnostics.Append(data.parseUser(ctx, &user)...)

	if user.Restricted {
		grants, err := client.GetUserGrants(ctx, data.Username.ValueString())
		if err != nil {
			resp.Diagnostics.AddError(
				fmt.Sprintf("Failed to get User Grants (%s): ", data.Username.ValueString()), err.Error(),
			)
			return
		}

		resp.Diagnostics.Append(data.parseUserGrants(ctx, grants)...)
	}

	if resp.Diagnostics.HasError() {
		return
	}
	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```

lgarber-akamai (changes requested):
Great work! One small refactor and this should be good to go

ezilber-akamai:
/acctest sha=da0c1330671ef3c975559e77364377f51beead6a pkg=linode/user

ezilber-akamai:
/acctest sha=293ccdefee19b1394e62eca8bca33f5259de2743 pkg=linode/user

linode/user/framework_datasource_schema.go
```diff
@@ -0,0 +1,82 @@
...
+		"cancel_account":        types.BoolType,
+		"longview_subscription": types.BoolType,
+	},
+}
+
+var linodeUserGrantsEntityObjectType = types.ObjectType{
+	AttrTypes: map[string]attr.Type{
+		"id":          types.Int64Type,
+		"permissions": types.StringType,
+	},
```
zliang-akamai, line 30: Can we add the `label` attribute here?

zliang-akamai:
By the way, [the docs](https://github.com/linode/terraform-provider-linode/blob/dev/website/docs/d/user.html.markdown) of this module seems very outdated. Maybe we can update it in this or another PR?

linode/user/framework_datasource_schema.go
ezilber-akamai: Done!

linode/user/framework_models.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	data.SSHKeys = sshKeys
+
+	id, err := json.Marshal(user)
+	if err != nil {
+		diags.AddError("Error marshalling json: %s", err.Error())
+		return diags
+	}
+
+	data.ID = types.StringValue(string(id))
```
zliang-akamai, line 51: Can we use `username` for the `id` instead? Maybe simpler.

ezilber-akamai: Initially I considered doing that but then I decided to use the JSON-Marshalled id object so that there aren't two fields with the same data. What do you think?

linode/user/framework_datasource.go
```diff
@@ -0,0 +1,106 @@
...
+			break
+		}
+	}
+
+	if user.Username == "" {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Linode User with username %s was not found", data.Username.ValueString()), "",
+		)
+		return
+	}
```
zliang-akamai, line 90: Can we move this check to some where before the API call? I think we can improve user experience this way by getting a faster hint to the users when they configure the `username` incorrectly.

linode/user/framework_datasource.go
```diff
@@ -0,0 +1,106 @@
...
+	resp.Diagnostics.Append(data.parseUser(ctx, &user)...)
```
zliang-akamai, line 91:
```suggestion
	resp.Diagnostics.Append(data.parseUser(ctx, &user)...)
	if resp.Diagnostics.HasError() {
		return
	}
```

linode/user/framework_datasource.go
ezilber-akamai: I think there needs to be an API call to check if the username the user is using actually exists. `user` is not set until after the API call because it depends on the returned list of usernames.

linode/user/framework_datasource.go
```diff
@@ -0,0 +1,111 @@
...
+	client := d.client
+
+	var data DataSourceModel
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	users, err := client.ListUsers(ctx, nil)
```
zliang-akamai, line 68: Let's switch to `GetUser` according the team discussion.

ezilber-akamai: Fixed!

zliang-akamai (approved):

ezilber-akamai:
/acctest sha=d3dcf88e22c15db60cbe4ef33adc4759311f7c84 pkg=linode/user

lgarber-akamai (approved):
Nice work!

PR #864, opened 2023-06-05
Migrated `instancetype` and `instancetypes` datasources to Framework

ezilber-akamai:
## 📝 Description

Migrated `instancetype` and `instancetypes` datasources to Framework

## ✔️ How to Test

`make PKG_NAME="linode/instancetype" testacc`
`make PKG_NAME="linode/instancetypes" testacc`

Ticket: TPT-1985


ezilber-akamai:
/acctest sha=ed7f6c2f5c88e1d24f92a3e3e0a328eb0c593850 pkg=linode/instancetype

ezilber-akamai:
/acctest sha=ed7f6c2f5c88e1d24f92a3e3e0a328eb0c593850 pkg=linode/instancetypes

.golangci.yml
```diff
@@ -14,7 +14,7 @@ linters-settings:
   errcheck:
     check-type-assertions: true
     check-blank: true
-
+
```
zliang-akamai, line 17: Let's remove the spaces here

ezilber-akamai: Removing the space doesn't cause the file to not track any changes. Odd.

linode/helper/frameworkfilter/local_filter.go
```diff
@@ -44,7 +46,7 @@ func (f Config) matchesFilter(
 		filterName := filter.Name.ValueString()
 
 		// Skip if this field should be filtered at an API level
-		if f[filterName].APIFilterable {
+		if f[filterName].APIFilterable && checkFilterSkip(filter) {
```
zliang-akamai, line 49: May I know why calling `checkFilterSkip` here is necessary here?

zliang-akamai (changes requested):
Some cases failed as described above

linode/helper/frameworkfilter/local_filter.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
+const EXACT = "exact"
```
zliang-akamai, line 14: I think it's a little bit weird to only make `"exact"` a constant while leaving others (e.g., `"substring"`) unchanged. Maybe let's do this another PR.

linode/helper/frameworkfilter/local_filter.go
lgarber-akamai: This check is necessary as some fields may be API-filterable but are not using an `exact` matching type, so we should only skip local filtering on fields that are both API filterable and have `exact`/nil values for `match_by`.

lgarber-akamai:
> ```terraform
> data "linode_instance_types" "specific-types" {
>   filter {
>     name = "class"
>     values = ["gp"]
>     match_by = "substring"
>   }
> }
> ```
> 
> This case seems to cause a failure.
> 
> ```
> │ Error: Invalid field type
> │ 
> │   with data.linode_instance_types.specific-types,
> │   on main.tf line 15, in data "linode_instance_types" "specific-types":
> │   15: data "linode_instance_types" "specific-types" {
> │ 
> │ Invalid type for field: linodego.LinodeTypeClass
> ```

I believe this is due to the switch statement filtering on the interface type [here](https://github.com/linode/terraform-provider-linode/blob/dev/linode/helper/frameworkfilter/local_filter.go#L131). 

I think we can address this by instead switching on the `reflect.Kind` of the value like we do [here](https://github.com/linode/terraform-provider-linode/blob/dev/linode/helper/filter.go#L559).

zliang-akamai:
> > ```terraform
> > data "linode_instance_types" "specific-types" {
> >   filter {
> >     name = "class"
> >     values = ["gp"]
> >     match_by = "substring"
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This case seems to cause a failure.
> > ```
> > │ Error: Invalid field type
> > │ 
> > │   with data.linode_instance_types.specific-types,
> > │   on main.tf line 15, in data "linode_instance_types" "specific-types":
> > │   15: data "linode_instance_types" "specific-types" {
> > │ 
> > │ Invalid type for field: linodego.LinodeTypeClass
> > ```
> 
> I believe this is due to the switch statement using the interface type [here](https://github.com/linode/terraform-provider-linode/blob/dev/linode/helper/frameworkfilter/local_filter.go#L131), which is preventing it from treating enums as their underlying types.
> 
> I think we can address this by instead switching on the `reflect.Kind` of the value like we do [here](https://github.com/linode/terraform-provider-linode/blob/dev/linode/helper/filter.go#L559).

Yep, thanks for reviewing. I putted up [an PR](https://github.com/linode/terraform-provider-linode/pull/870) for that.

linode/helper/frameworkfilter/local_filter.go
```diff
@@ -182,3 +184,7 @@ func checkFilterRegex(values []types.String, actualValue string) (bool, diag.Dia
 
 	return false, nil
 }
+
+func checkFilterSkip(filter FilterModel) bool {
```
zliang-akamai, line 188:
```suggestion
func isExactMatchFilter(filter FilterModel) bool {
```

.golangci.yml
ezilber-akamai: Fixed it!

ezilber-akamai:
/acctest sha=8792ef9539fa9ad9a6146b34b02e564d30192086 pkg=linode/instancetypes

lgarber-akamai (approved):
Looks good! There might be some conflicts with [Ye's PR](https://github.com/linode/terraform-provider-linode/pull/869) related to the filter framework changes, but they should be straightforward to resolve 👍

linode/instancetypes/framework_models.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		m.Addons = *addons
+
+		m.NetworkOut = types.Int64Value(int64(instanceType.NetworkOut))
+		m.Memory = types.Int64Value(int64(instanceType.Memory))
+		m.Transfer = types.Int64Value(int64(instanceType.Transfer))
+		m.VCPUs = types.Int64Value(int64(instanceType.VCPUs))
+
+		return m, nil
+	}
```
yec-akamai, line 64: Can we use the `instancetype.DataSourceModel` instead of building the `InstanceTypeModel`, and call the `parseLinodeType` function from `instancetype`? They should do the same work here.

yec-akamai: we can try:
```
for instanceType := range instanceTypes {
    var m instancetype.DataSourceModel
    diags := m.parseLinodeType(ctx, instanceTypes[i])
    if diags.HasError() {
	return diags
    }
    result[i] = m
}
```
probably need to export `parseLinodeType` to make it work here.

ezilber-akamai: Oh I see. I misunderstood what you meant. Thanks!

ezilber-akamai: Fixed!

linode/instancetypes/framework_models.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (model *InstanceTypeFilterModel) parseInstanceTypes(ctx context.Context,
+	instanceTypes []linodego.LinodeType,
+) diag.Diagnostics {
+	result := make([]instancetype.DataSourceModel, len(instanceTypes))
+
+	for i, instanceType := range instanceTypes {
+		var m instancetype.DataSourceModel
+
+		diags := m.ParseLinodeType(ctx, &instanceType)
```
yec-akamai, line 29: Here we need to use `&instanceTypes[i]` to address the pointer issue `Implicit memory aliasing in for loop.`, because `instasnceType` is reused for iteration and it points to the same memory location. We can use `instanceTypes[i]` to get the actual element's address.

ezilber-akamai: Ah that makes sense. Thanks!

ezilber-akamai:
/acctest sha=9c7b7a53289484dc1d5d64ca9fb2bd08871a4625 pkg=linode/instancetype

ezilber-akamai:
/acctest sha=9c7b7a53289484dc1d5d64ca9fb2bd08871a4625 pkg=linode/instancetypes

yec-akamai (approved):
Great work!

zliang-akamai:
I'm taking a final look, very quick

PR #869, opened 2023-06-06
TPT 1978: Migrate `image` & `images` data source; Update filtering framework

yec-akamai:
## 📝 Description

1. Migrate image and images data source from SDKv2 to terraform framework.
2. Add `TypeFunc` in the `FilterAttribute` to make sure we convert filter fields to correct types.
3. Add `TypeFunc` to migrated schema with filterConfig.
4. Change the return type of `GetLatestCreated` from any to []any. It can simplify the logic we need to build in the data source Read, when we only need to return the latest element. It's also a similar behavior in SDKv2's helper function [FilterLatest](https://github.com/linode/terraform-provider-linode/blob/4e80671ea4df41eb8f722299f5c6efb5801621db/linode/helper/filter.go#L297).
6. Apply other filtering framework change from #864

## ✔️ How to Test

`make PKG_NAME="linode/image" testacc`
`make PKG_NAME="linode/images" testacc`


yec-akamai:
/acctest sha=2fcb1307eb1e8c9e8fde44c8a31ac290f8186e0d pkg=linode

yec-akamai:
/acctest sha=2fcb1307eb1e8c9e8fde44c8a31ac290f8186e0d pkg=linode/image

yec-akamai:
/acctest sha=2fcb1307eb1e8c9e8fde44c8a31ac290f8186e0d pkg=linode/images

linode/helper/frameworkfilter/filter.go
```diff
@@ -157,6 +167,17 @@ func (f Config) GetAndFilter(
 	if d != nil {
 		return nil, d
 	}
-
 	return locallyFilteredElements, nil
 }
+
+func FilterTypeString(value string) (interface{}, error) {
```
lgarber-akamai, line 173: nitpick: We can use `any` here rather than `interface{}` to make the code a tiny bit more readable

linode/image/framework_models.go
```diff
@@ -0,0 +1,47 @@
...
+	image *linodego.Image,
+) {
+	data.ID = types.StringValue(image.ID)
+	data.Label = types.StringValue(image.Label)
+
+	data.Description = types.StringValue(image.Description)
+	if image.Created != nil {
+		data.Created = types.StringValue(image.Created.Format(time.RFC3339))
+	}
+	if image.Expiry != nil {
```
lgarber-akamai, line 37: I think we can set these two values to `types.StringNull()` when the fields are `nil`. Realistically this shouldn't be a case we run into, but it might be worth having that extra redundancy just in case 👍

lgarber-akamai (changes requested):
Great work! Just two tiny changes

lgarber-akamai (approved):

PR #876, opened 2023-06-07
ref: Migrate `linode_databases` data source to Framework

lgarber-akamai:
## 📝 Description

This change migrates the `linode_databases` data source from SDKv2 -> Framework

## ✔️ How to Test

```
make PKG_NAME=linode/databases testacc
```


jriddle-linode (approved):

linode/databases/framework_datasource_schema.go
```diff
@@ -0,0 +1,117 @@
...
+	"cluster_size":   {TypeFunc: frameworkfilter.FilterTypeInt},
+	"created":        {TypeFunc: frameworkfilter.FilterTypeString},
+	"encrypted":      {TypeFunc: frameworkfilter.FilterTypeBool},
+	"host_primary":   {TypeFunc: frameworkfilter.FilterTypeString},
+	"host_secondary": {TypeFunc: frameworkfilter.FilterTypeString},
+	"id":             {TypeFunc: frameworkfilter.FilterTypeInt},
+	"instance_uri":   {TypeFunc: frameworkfilter.FilterTypeString},
+	"updated":        {TypeFunc: frameworkfilter.FilterTypeString},
+	"label":          {TypeFunc: frameworkfilter.FilterTypeString},
+}
```
zliang-akamai, line 26: `label`, `region`, `status`, `type`, `version` are API filterable, right?

ezilber-akamai (approved):

linode/databases/framework_datasource_schema.go
lgarber-akamai: Looks like these fields are documented as filterable, but the API returns `[X-Filter] Could not apply filter` if any of these fields are included in the constructed filter. I'm not sure if this is a docs issue or an API issue, but I'm leaning towards it being an API issue.

lgarber-akamai: Just confirmed that this is an API issue and was able to pinpoint the cause of the bug. I've reported it to the API team internally 👍

zliang-akamai: Thanks for the insightful investigation!

linode/databases/framework_datasource_schema.go
```diff
@@ -0,0 +1,117 @@
...
+	"allow_list":     {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+	"cluster_size":   {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeInt},
+	"created":        {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+	"encrypted":      {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeBool},
+	"host_primary":   {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+	"host_secondary": {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+	"id":             {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeInt},
+	"instance_uri":   {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+	"updated":        {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+}
```
zliang-akamai, line 26: They are all marked as `APIFilterable: false`, lol? I think we can still mark these in theory API filterable fields, and wait for the fix from the API team?

lgarber-akamai: Hmm, in the interest of reducing the number of stale PRs, what if we kept these fields as non-API-filterable (which mirrors the SDKv2 implementation) and created a ticket to update them once the API changes have been released

zliang-akamai: Yep, that will work

PR #879, opened 2023-06-08
ref: Migrate `linode_region` data source from SDKv2 -> Framework

lgarber-akamai:
## 📝 Description

This change migrates the `linode_region` data source from SDKv2 -> Framework.

## ✔️ How to Test

```
make PKG_NAME=linode/region testacc
```


linode/region/framework_datasource_schema.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var frameworkDataSourceSchema = schema.Schema{
+	Attributes: map[string]schema.Attribute{
+		"country": schema.StringAttribute{
+			Description: "The country where this Region resides.",
+			Computed:    true,
+		},
```
zliang-akamai, line 13: Do we want to make this as optional?

zliang-akamai: Or the removal was intentional?

lgarber-akamai: Looks like this field being optional was a mistake in the SDKv2 implementation. I've removed it since there is no logic to resolve a region by country in this data source 👍

yec-akamai (approved):
Looks good!

zliang-akamai (approved):

ezilber-akamai (approved):

PR #882, opened 2023-06-09
Migrated `databasebackups` datasource to Framework

ezilber-akamai:
## 📝 Description

Migrated `databasebackups` datasource to Framework

## ✔️ How to Test

`make PKG_NAME="linode/databasebackups" testacc`

TPT-1967


yec-akamai (approved):

linode/databasebackups/datasource_test.go
```diff
@@ -23,65 +20,28 @@ var (
...
-		Created: &currentTime,
-	}
-
-	result := databasebackups.FlattenBackup(backup)
-	if result["id"] != backup.ID {
-		t.Fatal(cmp.Diff(result["id"], backup.ID))
-	}
-
-	if result["label"] != backup.Label {
-		t.Fatal(cmp.Diff(result["label"], backup.Label))
```
zliang-akamai, line 50: Was this test case deleted for some reasons?

zliang-akamai: I think we can make `parseMySQLBackup` and `parsePostgresSQLBackup` as independent functions and call them from `parseXXXXBackups` in the model file, and then bring these tests back to test those functions.

linode/databasebackups/framework_datasource.go
```diff
@@ -0,0 +1,112 @@
...
+			ctx, r.client, data.Filters, listMySQLBackups, data.Order, data.OrderBy)
+		if d != nil {
+			resp.Diagnostics.Append(d)
+			return
+		}
+
+		data.parseMySQLBackups(helper.AnySliceToTyped[linodego.MySQLDatabaseBackup](result))
+	}
+
+	if data.DatabaseType.ValueString() == "postgres" {
```
zliang-akamai, line 100:
```suggestion
	} else if data.DatabaseType.ValueString() == "postgres" {
```

linode/databasebackups/framework_datasource.go
```diff
@@ -0,0 +1,112 @@
...
+	listPostgresSQLBackups := func(ctx context.Context, client *linodego.Client, filter string) ([]any, error) {
+		backups, err := client.ListPostgresDatabaseBackups(ctx, int(data.DatabaseID.ValueInt64()), &linodego.ListOptions{
+			Filter: filter,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		return helper.TypedSliceToAny(backups), nil
+	}
```
zliang-akamai, line 87: This is an optional suggestion.
I feel moving the function definition into the if else block and only define listMySQLBackups when the db type is mysql might improve the memory usage slightly, what do you think?

zliang-akamai:
Overall looks good, nice work!

linode/databasebackups/framework_datasource.go
ezilber-akamai: Good idea. I implemented this change!

linode/databasebackups/datasource_test.go
ezilber-akamai: As far as adding additional test cases, none of the parsing functions return anything so I'm not sure about the best way to test these.

zliang-akamai: I think you can check the attributes in the model instance against the attributes in the linodego object

zliang-akamai: I think this is optional because we didn't test model parsing in other data sources...

https://github.com/linode/terraform-provider-linode/actions/runs/5258218942
Let's wait for the action result, and you can decide whether to bring these tests back or not.

ezilber-akamai: The action result will fail due to an unresolved issue with polling. Still trying to figure it out.

ezilber-akamai: Solved!

PR #888, opened 2023-06-14
Migrated `nbnode` datasource to Framework

ezilber-akamai:
## 📝 Description

Migrated `nbnode` datasource to Framework

## ✔️ How to Test

`make PKG_NAME="linode/nbnode" testacc`


zliang-akamai (approved):
I understand that the model is small but since we separate model in another file in other packages, I think we can also do that here. This is a optional suggestion though.

Nice work!

zliang-akamai (approved):

yec-akamai (approved):

PR #893, opened 2023-06-15
ref: Use composition to reduce boilerplate redundancy in resource implementations

lgarber-akamai:
## 📝 Description

This change introduces two new base classes that eliminate the need for per-resource implementations of `Configure(...)`, `Schema(...)`, `Metadata(...)`, etc.

## ✔️ How to Test

```
make testacc
```


ezilber-akamai (approved):
Huge improvement!

PR #895, opened 2023-06-16
TPT 2101: add `users` data source

yec-akamai:
## 📝 Description

Add `users` data source and apply the terraform framework. Build test cases to test basic, client filter, and substring filter. 
Add missing fields to `user` attributes. 

## ✔️ How to Test

`make PKG_NAME="linode/user" testacc`


go.mod
```diff
@@ -13,6 +13,8 @@ require (
 	golang.org/x/crypto v0.9.0
 )
 
+replace github.com/linode/linodego v1.17.0 => /Users/yechen/linode/linodego
```
jcallahan-akamai, line 16:
```suggestion
```

linode/users/tmpl/template.go
zliang-akamai, line 1: Can we also have a test case for the API filter? Because sub-string is technically still a local filter.

yec-akamai: The data_basic actually tests the API filter `username`

linode/users/framework_models.go
```diff
@@ -0,0 +1,58 @@
...
+func (data *UserFilterModel) parseUsers(
+	ctx context.Context,
+	client *linodego.Client,
+	users []linodego.User,
+) diag.Diagnostics {
+	result := make([]user.DataSourceModel, len(users))
+	for i := range users {
+		var userModel user.DataSourceModel
+		diags := userModel.ParseUser(ctx, &users[i])
+		if diags != nil {
```
zliang-akamai, line 33:
```suggestion
		if diags.HasError() {
```

zliang-akamai:
Another doubt I have is that `verified_phone_number` is a nullable field from API. I think we may want to fix this from linodego by making it as a pointer in the linodego struct.
image

zliang-akamai:
Others look good! Nice work on the sharing of resources between `user` and `users`!

yec-akamai:
> Another doubt I have is that `verified_phone_number` is a nullable field from API. I think we may want to fix this from linodego by making it as a pointer in the linodego struct. image

Thanks for pointing it out! I put up the fix in linodego now https://github.com/linode/linodego/pull/339

zliang-akamai (approved):

zliang-akamai:
GHA runs:
https://github.com/linode/terraform-provider-linode/actions/runs/5358436209
https://github.com/linode/terraform-provider-linode/actions/runs/5358434976

ykim-1 (approved):

zliang-akamai:
@yec-akamai don't forget the docs, lol

yec-akamai:
> @yec-akamai don't forget the [docs](https://github.com/linode/terraform-provider-linode/tree/dev/website/docs/d), lol
> 
> Feel free to do it in another PR.

Thanks for the heads up! I'll merge this one and create another for the docs.

PR #899, opened 2023-06-20
ref: migrate `domain` resource to terraform framework

jriddle-linode:
## 📝 Description

Migrate Domain resources to framework, not able to Modify plans in framework, rounded seconds will not be displayed in plan.

## ✔️ How to Test

`make PKG_NAME="linode/domain" testacc`


linode/domain/framework_resource.go
```diff
@@ -0,0 +1,193 @@
...
+
+	_, err := client.UpdateDomain(ctx, int(id), ops)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Failed to update domain: %v", id),
+			err.Error(),
+		)
+		return
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
yec-akamai, line 163:
```suggestion
	resp.Diagnostics.Append(state.parseDomain(ctx, domain)...)
	if resp.Diagnostics.HasError() {
		return
	}
	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
```

linode/domain/framework_models.go
```diff
@@ -1,42 +1,102 @@
...
+		}
+	}
+	m.Tags = tags
+	return nil
+}
+
+func (m *DomainModel) parseNonComputed(
+	ctx context.Context,
+	domain *linodego.Domain,
+) diag.Diagnostics {
```
yec-akamai, line 94: We don't have to return diagnostics from this function since there isn't any error raised from the parsing.

linode/domain/framework_resource.go
```diff
@@ -0,0 +1,230 @@
...
+	err := client.DeleteDomain(ctx, int(id))
+	if err != nil {
+		if lerr, ok := err.(*linodego.Error); (ok && lerr.Code != 404) || !ok {
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Failed to delete domain with id %v", id),
+				err.Error(),
+			)
+		}
+		return
+	}
```
yec-akamai, line 176: May be nicer adding a warning to let the user know that the resource is not found:
```suggestion
	if err != nil {
		if lerr, ok := err.(*linodego.Error); ok && lerr.Code == 404 {
			resp.Diagnostics.AddWarning(
				"Domain does not exist.",
				fmt.Sprintf("Domain %v does not exist, removing from state.", id),
			)
			return
		} 
		resp.Diagnostics.AddError(
			fmt.Sprintf("Failed to delete domain with id %v", id),
			err.Error(),
		)
		return
	}
```

yec-akamai (commented):
Great work! Just a few small suggestions.

linode/domain/framework_schema_resource.go
```diff
@@ -0,0 +1,119 @@
...
+		"type": schema.StringAttribute{
+			Description: "If this Domain represents the authoritative source of information for the domain it " +
+				"describes, or if it is a read-only copy of a master (also called a slave).",
+			Required: true,
+			Validators: []validator.String{
+				stringvalidator.OneOf("master", "slave"),
+			},
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.RequiresReplace(),
+				stringplanmodifier.UseStateForUnknown(),
```
zliang-akamai, line 45: I think `UseStateForUnknown` isn't necessary here?
```suggestion
```

linode/domain/framework_schema_resource.go
```diff
@@ -0,0 +1,119 @@
...
+			Optional: true,
+			Computed: true,
+		},
+		"ttl_sec": schema.Int64Attribute{
+			Description: "'Time to Live' - the amount of time in seconds that this Domain's records may be " +
+				"cached by resolvers or other domain servers. " + domainSecondsDescription,
+			Optional:   true,
+			Computed:   true,
+			CustomType: customtypes.LinodeDomainSecondsType{},
+		},
```
zliang-akamai, line 87: Maybe we can use `UseStateForUnknown` for these?

linode/domain/framework_resource.go
```diff
@@ -0,0 +1,235 @@
...
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	domainID := int(state.ID.ValueInt64())
+
+	if !domainDeepEqual(plan, state) {
+		r.updateDomain(ctx, resp, plan, domainID)
+	}
```
zliang-akamai, line 147:
```suggestion
	if !domainDeepEqual(plan, state) {
		r.updateDomain(ctx, resp, plan, domainID)
	} else {
		resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
	}
```

> An error is returned unless every null or known value in the request plan is saved exactly as-is into the response state.

Although extra effort may be needed to handle Unknown values in the `else` branch, I think it's nicer do this to ensure we align with the requirement from the TF framework's docs.

linode/domain/framework_schema_resource.go
```diff
@@ -0,0 +1,138 @@
...
+var frameworkResourceSchema = schema.Schema{
+	Attributes: map[string]schema.Attribute{
+		"id": schema.Int64Attribute{
+			Description: "The Domain's unique ID.",
+			Computed:    true,
+			Validators: []validator.Int64{
+				int64validator.ConflictsWith(path.Expressions{
+					path.MatchRoot("domain"),
+				}...),
+			},
```
zliang-akamai, line 31:
```suggestion
			},
			PlanModifiers: []planmodifier.Int64{
				int64planmodifier.UseStateForUnknown(),
			},
```

linode/domain/framework_resource.go
```diff
@@ -0,0 +1,237 @@
...
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	domainID := int(state.ID.ValueInt64())
+
+	if !domainDeepEqual(plan, state) {
+		r.updateDomain(ctx, resp, plan, domainID)
+		return
```
yec-akamai, line 147: Can we remove the `return` here to make use of the `State.Set` in the end of the function, instead of calling the `State.Set` inside the updateDomain()? It's a bit more clear to see that we always set the user plan no matter we call the update endpoint or not.

linode/domain/framework_resource.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+func (r *Resource) updateDomain(
+	ctx context.Context,
+	resp *resource.UpdateResponse,
+	plan DomainModel,
+	domainID int,
+) {
```
yec-akamai, line 191: Probably can modify this helper function a bit to be:
```suggestion
func (plan *DomainModel) updateDomain(
	ctx context.Context,
	resp *resource.UpdateResponse,
	client *linodego.Client,
	domainID int,
) {
```
So that it's easier for us to set the state from plan when head back to the main function.

linode/domain/framework_resource.go
```diff
@@ -0,0 +1,237 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	domainID := int(state.ID.ValueInt64())
+
+	if !domainDeepEqual(plan, state) {
+		r.updateDomain(ctx, resp, plan, domainID)
+		return
+	}
```
yec-akamai, line 148: May need an else clause to handle those unknown values when `updateDomain()` is not called. For example, [https://github.com/linode/terraform-provider-linode/pull/947](https://github.com/linode/terraform-provider-linode/pull/947/files)

linode/domain/framework_schema_resource.go
```diff
@@ -0,0 +1,141 @@
...
+		"refresh_sec": schema.Int64Attribute{
+			Description: "The amount of time in seconds before this Domain should be refreshed. " +
+				domainSecondsDescription,
+			Optional:   true,
+			Computed:   true,
+			CustomType: customtypes.LinodeDomainSecondsType{},
+			PlanModifiers: []planmodifier.Int64{
+				int64planmodifier.UseStateForUnknown(),
+			},
+		},
```
yec-akamai, line 130: I don't think we should add `UseStateForUnknown()` to these attributes, because they're updatable in the Update plan. cc @zliang-akamai how do you think?

zliang-akamai: @yec-akamai So `UseStateForUnknown()` would only be using the state values when the user doesn't explicitly configure these values in the tf file. I don't think these values can be updated during an update operation when the user didn't configure them. Am I correct?

linode/domain/resource_test.go
```diff
@@ -68,8 +67,8 @@ func TestAccResourceDomain_basic_smoke(t *testing.T) {
 					resource.TestCheckResourceAttrSet(resName, "retry_sec"),
 					resource.TestCheckResourceAttrSet(resName, "expire_sec"),
 					resource.TestCheckResourceAttrSet(resName, "status"),
-					resource.TestCheckNoResourceAttr(resName, "master_ips"),
-					resource.TestCheckNoResourceAttr(resName, "axfr_ips"),
+					resource.TestCheckResourceAttr(resName, "master_ips.#", "0"),
+					resource.TestCheckResourceAttr(resName, "axfr_ips.#", "0"),
```
zliang-akamai, line 71: This seems like a breaking change? Will this framework resource work well with the state created by the previous SDKv2 resource? Did we test that?

linode/domain/resource_test.go
```diff
@@ -116,36 +115,6 @@ func TestAccResourceDomain_update(t *testing.T) {
 	})
 }
 
-func TestAccResourceDomain_roundedDomainSecs(t *testing.T) {
```
zliang-akamai, line 119: Was there some reason caused this test case fails?

linode/helper/customtypes/domainsecond.go
```diff
@@ -0,0 +1,137 @@
...
+		for _, value := range accepted {
+			if n <= value {
+				return value
+			}
+		}
+		return accepted[len(accepted)-1]
+	}
+
+	return rounder(newValue) == oldValue
+}
```
zliang-akamai, line 137: > A value of 0 is equivalent to the default value of 1209600.

https://www.linode.com/docs/api/domains/#domain-view__responses

Let's don't forget this case

zliang-akamai: Actually, it's different among 4 different domain seconds... Maybe we can create some more extended struct for each of them? I am not sure.

linode/domain/framework_resource.go
zliang-akamai: `master_ips` and `axfr_ips` are with the use state for unknown plan modifier, and `tags` isn't a computed attribute so won't be an unknown, right? ...

linode/domain/resource_test.go
jriddle-linode: I believe this is because now we have to set State to a specific null value instead of letting it be empty.

linode/domain/framework_schema_resource.go
yec-akamai: https://github.com/linode/terraform-provider-linode/pull/899/files#diff-f38fff6b6d425a04b761a731e00f7c4101b5361046353b365db5ba4be83aef71R198-R201
These fields are included in the updateOptions, so I think they can be updated in the config. 

> when the user didn't explicitly configure them.
I remember last time we discussed, if we set `UseStateForUnknown()` to an attribute, it will not be applied `known after apply`, even if it can be configured explicitly and it may be misleading to users. That's why I think we don't want to have `UseStateForUnknown()` to these fields because they can be configured.

linode/domain/framework_resource.go
yec-akamai: @zliang-akamai I raised it up is because I think some of the attributes shouldn't use `UseStateForUnknown()` and we probably want to update them here.

linode/domain/framework_schema_resource.go
zliang-akamai: So my observations are:
- When the attribute is a **computed** attribute **and is not explicitly configured** by the user, it will show `known after apply` during `terraform plan`.
- For some computed attributes, TF provider will update them without explicitly configured by the user, e.g. `updated`. And we shouldn't apply `UseStateForUnknown()` for them because of misleading annotations (state values) instead of `known after apply` shown during the `terraform plan`.
- For some other (computed or not) attributes, TF provider will only update them when the user explicitly configured a new value. We should be safe to use `UseStateForUnknown()`. And I think these attributes belong to this case.

Did I miss anything? Or these attributes can be updated without explicitly configured new value?

linode/domain/resource_test.go
zliang-akamai: Ah, got it

linode/domain/framework_schema_resource.go
yec-akamai: Gotcha. I thought `UseStateForUnknown()` will impact explicitly configured attributes as well. Now it's all good, thanks for the explanation

jriddle-linode:
Creates a breaking change, closing for now, revisiting Resource Framework at a later date.

PR #908, opened 2023-06-23
Migrated `RDNS` resource to Framework

ezilber-akamai:
## 📝 Description

Migrated `RDNS` resource to Framework

## ✔️ How to Test

`make PKG_NAME="linode/rdns" testacc`


jriddle-linode (approved):

zliang-akamai:
@lgarber-akamai I think `ImportState` is already implemented in the base resource?

lgarber-akamai:
> @lgarber-akamai I think `ImportState` is already implemented in the base resource?

@zliang-akamai As a part of another resource migration, the default `ImportState` function was altered to expect an integer as the ID attribute, however this resource uses a string ID. Because of this, the resource needs to override the `ImportState` function to use the string import helper.

We have a ticket in our internal tracker to improve the base class and remove the need for type-specific import functions, so I'll address this inconsistency soon :+1:

linode/rdns/framework_resource.go
```diff
@@ -0,0 +1,166 @@
...
+) {
+	var data ResourceModel
+
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client := r.Meta.Client
+	ip, err := client.GetIPAddress(ctx, data.ID.ValueString())
```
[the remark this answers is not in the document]
lgarber-akamai: I think you're right; the state should be updated with the latest RDNS address during the read step so this request is redundant. I'll fix this real quick :+1:

linode/rdns/framework_resource_schema.go
```diff
@@ -0,0 +1,47 @@
...
+				helper.NewStringLengthValidator(3, 254),
+			},
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.UseStateForUnknown(),
+			},
+		},
+		"wait_for_available": schema.BoolAttribute{
+			Description: "If true, the RDNS assignment will be retried within the operation timeout period.",
+			Optional:    true,
+			Computed:    true,
```
zliang-akamai, line 38: Can we default it to be true (or false) instead of saying it's a computed field? I am not sure if this is better though, but it seems a little bit more clear this way. So it's an optional suggestion.

linode/rdns/framework_resource_schema.go
```diff
@@ -0,0 +1,47 @@
...
+				stringplanmodifier.RequiresReplace(),
+			},
+			Validators: []validator.String{
+				helper.NewIPStringValidator(),
+			},
+		},
+		"rdns": schema.StringAttribute{
+			Description: "The reverse DNS assigned to this address. For public IPv4 addresses, this will be set " +
+				"to a default value provided by Linode if not explicitly set.",
+			Required: true,
```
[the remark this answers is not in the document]
zliang-akamai: This is very tricky because TF try to keep user configured values as the [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth) but the Linode API will reset the value if this field is empty. So an empty field (null or an empty string) will be semantically equaled to the reset value (something like `45-80-141-70.ip.linodeusercontent.com`), and the custom type's semantic equivalence checking can't check it because it doesn't know the IP address in the [`StringSemanticEquals` function](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/custom-types#semantic-equality). The best way in my mind is to manually checking the semantic equivalence in the parse function of the model, I am not very sure.

linode/rdns/framework_resource.go
```diff
@@ -0,0 +1,149 @@
...
+	updateOpts := linodego.IPAddressUpdateOptions{
+		RDNS: nil,
+	}
+
+	_, err := client.UpdateIPAddress(ctx, data.Address.ValueString(), updateOpts)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to delete Linode RDNS",
+			err.Error(),
+		)
```
zliang-akamai, line 146: Should ignore 404? When the user delete linode first and then the rdns, even if the rdns was gone with the linode, the deletion can still be considered as success, right?

linode/rdns/framework_resource.go
```diff
@@ -0,0 +1,149 @@
...
+	ip, err := client.UpdateIPAddress(ctx, data.Address.ValueString(), updateOpts)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to create Linode RDNS",
+			err.Error(),
+		)
+		return
+	}
+
+	data.parseIP(ip)
```
zliang-akamai, line 54: Technically we can only change computed attributes here...

linode/rdns/framework_resource.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Failed to update the Linode RDNS",
+				err.Error(),
+			)
+			return
+		}
+
+		state.parseIP(ip)
+		resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
```
zliang-akamai, line 119: > An error is returned unless every null or known value in the request plan is saved exactly as-is into the response state. Only unknown plan values can be modified.

I think we should save `plan`, not API response `ip` into the state here.

linode/rdns/framework_resource_schema.go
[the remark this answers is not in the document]
lgarber-akamai: The previous implementation of this resource was only concerned with non-null (default) RDNS values since a Linode without RDNS should already use the default value. It might be worth adding this functionality though :+1:

linode/rdns/framework_resource_schema.go
lgarber-akamai: Unlike in SDKv2, Framework requires that all optional values with defaults also be computed:

> Attribute: 
                Summary: Schema Using Attribute Default For Non-Computed Attribute
                Detail: Attribute "wait_for_available" must be computed when using default. This is an issue with the provider and should be reported to the provider developers.

It's super weird, but I think this can stay as-is for now :slightly_smiling_face:

linode/rdns/framework_resource.go
lgarber-akamai: Great catch!

linode/rdns/framework_resource_schema.go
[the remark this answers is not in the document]
zliang-akamai: But I think we will still have to face the problem that an empty string semantically equals to the default rdns, right? This can happen when a user want to set the rdns to be the default value explicitly without manually concatenating the string.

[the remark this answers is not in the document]
zliang-akamai: Ah, I miss the validator, so an empty string is not allowed?

zliang-akamai (approved):
Lol, we finally got this, thanks for the hard work! @lgarber-akamai @ezilber-akamai

yec-akamai (approved):
Great work!

zliang-akamai (changes requested):
Found some issues, still investigating

zliang-akamai:
@lgarber-akamai @yec-akamai I pushed some commits for fixing some issues I discovered when testing the update of resource. Please let me know if they look fine to you guys.

zliang-akamai (approved):

zliang-akamai:
Due to a test issue fixed in https://github.com/linode/terraform-provider-linode/pull/908/commits/269de1a20de276e7bba846f6acf63da239ba85e3, we were not able to catch that by the acceptance test.

lgarber-akamai:
@zliang-akamai Looks good to me, thanks for making these changes! Are we good to merge?

linode/rdns/framework_resource.go
```diff
@@ -143,14 +143,20 @@ func (r *Resource) Update(
...
 				err.Error(),
 			)
 			return
 		}
+		plan.parseComputedAttributes(ip)
+	} else {
+		// computed attributes without stringplanmodifier.UseStateForUnknown()
+		// will be an unknown value when during an update operation, and it's
+		// required to be a known or null value in the response.
+		plan.ID = plan.Address
```
yec-akamai, line 159: why don't we apply `UseStateForUnknown()` to ID? IMO it should remain no change during the plan update.

zliang-akamai: @yec-akamai You are right.

I didn't apply `UseStateForUnknown()` because I thought the address change can make the ID change. But now I realize that the address change will force a replacement of the resource and the ID will still be shown as `(known after apply)` anyway, so I think we can apply it.

yec-akamai (approved):
Nice work! Thanks for sorting out the update issues.

PR #911, opened 2023-06-26
TPT 2306: Migrate `nb` resource to terraform framework

yec-akamai:
## 📝 Description

Migrate `nb` resource from SDKv2 to terraform framework. This resource also implements `UpgradeState` to realize versioning the resource schemas which was supported in the SDKv2 implementation. Terrafrom Framework doc for State Upgrade: https://developer.hashicorp.com/terraform/plugin/framework/resources/state-upgrade

## ✔️ How to Test

`make PKG_NAME="linode/nb" testacc`


linode/nb/tmpl/basic.gotf
```diff
@@ -5,6 +5,7 @@ resource "linode_nodebalancer" "foobar" {
     region = "{{ .Region }}"
     client_conn_throttle = 20
     tags = ["tf_test"]
+    transfer {}
```
yec-akamai, line 8: I was asked to add the `transfer` block here to compile the code, probably because I implemented the `transfer` as `ListNestedBlock` in the schema. Will it introduce a breaking change?

zliang-akamai:
> Use the [resource.UseStateForUnknown() attribute plan modifier](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#UseStateForUnknown) for Computed attributes that are known to not change during resource updates. This will enhance the Terraform plan to not show => (known after apply) differences.

https://developer.hashicorp.com/terraform/plugin/framework/resources/update#recommendations

Should we use some `resource.UseStateForUnknown()` attribute plan modifiers for the attributes?

linode/nb/framework_resource_schema.go
```diff
@@ -0,0 +1,146 @@
...
+			Computed:    true,
+		},
+		"created": schema.StringAttribute{
+			Description: "When this NodeBalancer was created.",
+			Computed:    true,
+		},
+		"updated": schema.StringAttribute{
+			Description: "When this NodeBalancer was last updated.",
+			Computed:    true,
+		},
```
zliang-akamai, line 60: I think we may use rfc 3339 custom types here, not very necessary though, but I think that's a nice to have thing and an additional layer of redundancy.

linode/nb/tmpl/basic.gotf
zliang-akamai: Let me check if there is a possible workaround...

lgarber-akamai: Judging from the migration docs, it looks like we'll need to use a `ListAttribute` with a `types.ObjectType` for the `ElementType` field (similar to how we used to do data sources). 

See https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed#framework

linode/nb/framework_resource.go
```diff
@@ -0,0 +1,241 @@
...
+		return
+	}
+
+	data.ID = types.Int64Value(int64(nodebalancer.ID))
+	resp.Diagnostics.Append(data.parseNodeBalancer(ctx, nodebalancer)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```
zliang-akamai, line 70: > any attribute that does not have Computed set to true cannot be influenced by the provider; it can only be changed via the config or via state refreshing.

https://developer.hashicorp.com/terraform/plugin/framework/handling-data/schemas#computed

We are required not to change user configured (non-computed) values except in state refreshing (Read function). So we will have to separate parse computed and non-computed attributes functions.

linode/nb/framework_resource.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	err := client.DeleteNodeBalancer(ctx, int(data.ID.ValueInt64()))
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to delete Nodebalancer",
+			err.Error(),
+		)
+		return
+	}
```
zliang-akamai, line 173: > Ignore errors that signify the resource is no longer existent.

Such as a 404 error...
https://developer.hashicorp.com/terraform/plugin/framework/resources/delete#recommendations

linode/nb/framework_resource.go
lgarber-akamai: The language used by HashiCorp is a bit ambiguous in meaning, but it seems to imply that non-computed attributes should only be updated during read like you're saying. That said, it seems a bit odd to arbitrarily split non-computed and computed attributes into separate parse functions. I wonder if there's a better way to approach this.

yec-akamai:
> > Use the [resource.UseStateForUnknown() attribute plan modifier](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#UseStateForUnknown) for Computed attributes that are known to not change during resource updates. This will enhance the Terraform plan to not show => (known after apply) differences.
> 
> https://developer.hashicorp.com/terraform/plugin/framework/resources/update#recommendations
> 
> Should we use some `resource.UseStateForUnknown()` attribute plan modifiers for the attributes?

Resolved it.

lgarber-akamai:
@yec-akamai This PR looks great, nice work!

Somewhat unrelated to this change, the `nbnode` tests are failing on this branch due to never having been migrated to Protov5, so they're unable to use the migrated `linode_nodebalancer` resource: https://github.com/linode/terraform-provider-linode/blob/TPT-2306/migrate-nb-resource/linode/nbnode/resource_test.go#L87

Is there any chance you could move those tests over to Protov5 as a part of this PR? 

Here's an example of how we can handle provider argument overrides in ProtoV5 acceptance tests: https://github.com/linode/terraform-provider-linode/blob/dev/linode/networkingip/tmpl/data_basic.gotf#L3

yec-akamai:
> 

Thanks for pointing this out! I've moved the nbnode tests to use ProtoV5 and it passed locally. Can you take a look at it as well?

yec-akamai:
> @yec-akamai This PR looks great, nice work!
> 
> Somewhat unrelated to this change, the `nbnode` tests are failing on this branch due to never having been migrated to Protov5, so they're unable to use the migrated `linode_nodebalancer` resource: https://github.com/linode/terraform-provider-linode/blob/TPT-2306/migrate-nb-resource/linode/nbnode/resource_test.go#L87
> 
> Is there any chance you could move those tests over to ProtoV5 as a part of this PR?
> 
> Here's an example of how we can handle provider argument overrides in ProtoV5 acceptance tests: https://github.com/linode/terraform-provider-linode/blob/dev/linode/networkingip/tmpl/data_basic.gotf#L3

Thanks for pointing this out! I've moved the nbnode tests to use ProtoV5 and it passed locally. Can you take a look at it as well?

lgarber-akamai (approved):
All tests pass locally, great work!

linode/nb/framework_resource_schema.go
```diff
@@ -0,0 +1,142 @@
...
+
+var resourceNodebalancerV0 = schema.Schema{
+	Attributes: map[string]schema.Attribute{
+		"id": schema.Int64Attribute{
+			Description: "The unique ID of the Linode NodeBalancer.",
+			Computed:    true,
+		},
+		"label": schema.StringAttribute{
+			Description: "The label of the Linode NodeBalancer.",
+			Optional:    true,
```
zliang-akamai, line 94:
```suggestion
			Optional:    true,
			Validators: []validator.String{
				stringvalidator.LengthBetween(3, 32),
			},
```

linode/nb/framework_resource_schema.go
```diff
@@ -0,0 +1,142 @@
...
+	Version: 1,
+	Attributes: map[string]schema.Attribute{
+		"id": schema.Int64Attribute{
+			Description:   "The unique ID of the Linode NodeBalancer.",
+			Computed:      true,
+			PlanModifiers: []planmodifier.Int64{int64planmodifier.UseStateForUnknown()},
+		},
+		"label": schema.StringAttribute{
+			Description: "The label of the Linode NodeBalancer.",
+			Optional:    true,
```
zliang-akamai, line 26:
```suggestion
			Optional:    true,
			Validators: []validator.String{
				stringvalidator.LengthBetween(3, 32),
			},
```

zliang-akamai: The import would also need to be added if this suggestion applied

linode/nb/framework_resource.go
```diff
@@ -0,0 +1,269 @@
...
+	result["total"] = total
+
+	transferObj, diags := types.ObjectValue(TransferObjectType.AttrTypes, result)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	resultList, diags := basetypes.NewListValue(
```
zliang-akamai, line 257: I think it would be nicer if we switch all `basetypes.NewSomeValue(...)` to be `types.SomeValue(...)`, which is a higher level API.

linode/nb/framework_resource.go
```diff
@@ -0,0 +1,269 @@
...
+		if err != nil {
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Failed to update Nodebalancer %v", plan.ID.ValueInt64()),
+				err.Error(),
+			)
+			return
+		}
+
+		resp.Diagnostics.Append(plan.parseComputedAttrs(ctx, nodebalancer)...)
+		resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
zliang-akamai, line 154: > An error is returned unless every null or known value in the request plan is saved exactly as-is into the response state. Only unknown plan values can be modified.

I think we should technically save `plan` outside of this if block in this case

linode/nb/framework_resource.go
```diff
@@ -0,0 +1,269 @@
...
+	}
```
zliang-akamai, line 155: > An error is returned unless every null or known value in the request plan is saved exactly as-is into the response state. Only unknown plan values can be modified.

I think we should technically save `plan` outside of the if block in this case.

```suggestion
	}
	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```

zliang-akamai, line 155: > An error is returned unless every null or known value in the request plan is saved exactly as-is into the response state. Only unknown plan values can be modified.

I think we should technically save `plan` outside of the if block in this case.

```suggestion
		resp.Diagnostics.Append(plan.parseComputedAttrs(ctx, nodebalancer)...)
		if resp.Diagnostics.HasError() {
			return
		}
	}
	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```

zliang-akamai: I think this is not necessary though.

linode/nb/framework_resource.go
yec-akamai: Good point! Updated it.

zliang-akamai (approved):

PR #914, opened 2023-06-27
ref: Improve base resource and data source structs

lgarber-akamai:
## 📝 Description

This change makes various improvements to the base resource and data source structures that are consumed in individual resource and data source implementations. Particularly this change adds a consistent and clean way to handle imports on ID attributes with varying types, adds a better method to configure base objects, and provides an intended way to optionally override the Schema function.

## ✔️ How to Test

```
make testacc
```


zliang-akamai (approved):
Nice!

yec-akamai (approved):
Nice improvement!

PR #916, opened 2023-06-28
TPT 2154: Add documentations to users data source; update user data source docs

yec-akamai:
## 📝 Description

Write documentations to describe the users data source that recently added. Also modify the user data source to reflect the update of new fields.


jriddle-linode (approved):

lgarber-akamai (changes requested):
Looks great!

Could you add an entry for the `linode_users` data source docs [here](https://github.com/linode/terraform-provider-linode/blob/d84f15aa691b9521ce814d912780ca28df0b4c83/website/linode.erb) in order to include it in the sidebar?

lgarber-akamai (approved):
Nice work!

PR #931, opened 2023-07-03
TPT 2035: Migrate `ipv6range` resource to terraform framework

yec-akamai:
## 📝 Description

Migrate `ipv6range` resource from SDKv2 to terraform framework. Move parsing related functions in data source to framework_models.

Modified test cases implementation because one account can only create one IPv6 range in each region. 

## ✔️ How to Test

`make PKG_NAME="linode/ipv6range" testacc`


linode/ipv6range/framework_models.go
```diff
@@ -0,0 +1,69 @@
...
+	LinodeId     types.Int64  `tfsdk:"linode_id"`
+	RouteTarget  types.String `tfsdk:"route_target"`
+	Range        types.String `tfsdk:"range"`
+	IsBGP        types.Bool   `tfsdk:"is_bgp"`
+	Linodes      types.Set    `tfsdk:"linodes"`
+	Region       types.String `tfsdk:"region"`
+	ID           types.String `tfsdk:"id"`
+}
+
+func (data *DataSourceModel) parseIPv6Range(
```
lgarber-akamai, line 32: I think we'll need to apply the same split-logic parse functions as discussed here: https://github.com/linode/terraform-provider-linode/pull/911#discussion_r1251153929

linode/ipv6range/tmpl/template.go
```diff
@@ -48,11 +49,12 @@ func ReassignmentStep2(t *testing.T, label, region string) string {
 		})
 }
 
-func RaceCondition(t *testing.T, label, region string) string {
+func RaceCondition(t *testing.T, label, region1, region2 string) string {
```
lgarber-akamai, line 52: For context, this test was created to test a race condition in the API that would occur between multiple IPv6 ranges created in the same DC due to an improperly scoped lock (only locking on Linode ID rather than Linode ID + DC). 

If possible, could we try to revert to a single region for the test template the ensure there are no regressions?

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,210 @@
...
+		)
+		return
+	}
+
+	ipv6range, err := client.CreateIPv6Range(ctx, createOpts)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to create ipv6 range.",
+			err.Error(),
+		)
```
lgarber-akamai, line 64: Looks like there's a missing return here :+1:

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,210 @@
...
+				"IPv6 range does not exist.",
+				fmt.Sprintf("IPv6 range \"%s\" does not exist, removing from state.", data.ID.ValueString()),
+			)
+			resp.State.RemoveResource(ctx)
+			return
+		}
+		resp.Diagnostics.AddError(
+			"Failed to get ipv6 range after creation.",
+			err.Error(),
+		)
```
lgarber-akamai, line 84: I wonder if we could break this error handling logic out into a helper function since it's also used in the read function.

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,210 @@
...
+		return
+	}
+
+	if err := client.DeleteIPv6Range(ctx, data.ID.ValueString()); err != nil {
+		if lerr, ok := err.(*linodego.Error); ok && lerr.Code == 404 {
+			resp.Diagnostics.AddWarning(
+				"IPv6 range does not exist.",
+				fmt.Sprintf("IPv6 range \"%s\" does not exist, removing from state.", data.ID.ValueString()),
+			)
+			resp.State.RemoveResource(ctx)
```
lgarber-akamai, line 201: I think we can remove this line since the `Delete(...)` function will implicitly remove the resource from the state :slightly_smiling_face:

lgarber-akamai (changes requested):
Great work! Just a few things

linode/ipv6range/tmpl/template.go
yec-akamai: I see. How to make this test case work locally? Due to the API limitation I can't create multiple ipv6 ranges in the same region when running it. Should I submit a request to the API team?

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,222 @@
...
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	ipv6range, diag := getIPv6Range(ctx, client, data.ID.ValueString())
+
+	if diag != nil {
+		if diag.Summary() == "IPv6 range does not exist." {
```
lgarber-akamai, line 110: It might be better for this error string to be a constant so we can be confident that the errors match

lgarber-akamai (approved):
Works well in my local testing, great work!

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,224 @@
...
+			"Either linode_id or route_target must be specified.",
+		)
+		return
+	}
+
+	ipv6range, err := client.CreateIPv6Range(ctx, createOpts)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to create ipv6 range.",
+			err.Error(),
```
zliang-akamai, line 66: Can we put `linode_id` or `route_target` to be a part of the error message so the user will know which IPv6 range creation was failed?

linode/ipv6range/framework_resource.go
zliang-akamai:
```suggestion
		if diag.Summary() == ipv6rangeNotExist {
```

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,224 @@
...
+	// computed fields (region, is_bgp, linodes).
+	ipv6rangeR, diag := getIPv6Range(ctx, client, data.ID.ValueString())
+
+	if diag != nil {
+		if diag.Summary() == ipv6rangeNotExist {
+			resp.State.RemoveResource(ctx)
+		}
+		resp.Diagnostics.Append(diag)
+		return
+	}
```
zliang-akamai, line 87: > An error is returned if the response state has the RemoveResource() method called. This method is not valid during creation.
https://developer.hashicorp.com/terraform/plugin/framework/resources/create#caveats

We can't call RemoveResource() in create function...

zliang-akamai:
As discussed, "there could be an edge case that when an IPv6 range initially assigned to a linode by Terraform, then the user remove it outside of the Terraform then reassign (although the range is randomly selected, there is a chance that same range can be assigned to another linode) it to another linode. In this case, we should probably call the RemoveResource() in Read function for this case."

I think "it's hard to check for this. It will have to compare linodes against linode_id, and if route_target was configured, it's even more complicated, it will need to call get linode instance function and inspect the IPv6 SLAAC address"

lgarber-akamai:
> As discussed in DM, "there could be an edge case that when an IPv6 range initially assigned to a linode by Terraform, then the user remove it outside of the Terraform then reassign (although the range is randomly selected, there is a chance that same range can be assigned to another linode) it to another linode. In this case, we should probably call the RemoveResource() in Read function for this case."
> 
> I think "it's hard to check for this. It will have to compare linodes against linode_id, and if route_target was configured, it's even more complicated, it will need to call get linode instance function and inspect the IPv6 SLAAC address"

I wonder if it's worth the extra API request(s) and added complexity to cover this edge case considering users shouldn't be removing managed infrastructure outside of Terraform, and if they do they should be manually tainting it.

zliang-akamai:
@lgarber-akamai That makes sense. Maybe we can just document this case to warn them about this uncovered edge case? Because theoretically TF should handle the change made outside of TF during the refresh.

lgarber-akamai:
> @lgarber-akamai That makes sense. Maybe we can just document this case to warn them about this uncovered edge case? Because theoretically TF should handle the change made outside of TF during the refresh.

Sure, I think that'd be a good addition to the documentation 👍

website/docs/r/ipv6_range.html.markdown
```diff
@@ -10,6 +10,7 @@ description: |-
 
 Manages a Linode IPv6 range.
 
+When an IPv6 range initially assigned to a linode by Terraform, then user removes it outside of the Terraform and reassigns it to another linode, there is a chance that same range can be assigned to another linode, although the range is randomly selected. In this case, user should be manually tainting it.
```
zliang-akamai, line 13:
```suggestion
We highly recommend that users do not remove an IPv6 range created by this Terraform resource outside of Terraform. This is because if a user manually removes an IPv6 range created by Terraform, and then assigns some IPv6 ranges to other linodes outside of Terraform, there is a chance that the same IPv6 range can be assigned to another linode, even though the new range is randomly selected. This will result in the newly assigned IPv6 range being managed by this Terraform resource. In this case, the user should manually taint this resource.
```

Rewrite it up a little bit with a GPT, lol

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,209 @@
...
+	var data ResourceModel
+	client := r.Meta.Client
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if err := client.DeleteIPv6Range(ctx, data.ID.ValueString()); err != nil {
+		if lerr, ok := err.(*linodego.Error); ok && lerr.Code == 404 {
```
zliang-akamai, line 196:
```suggestion
		if lerr, ok := err.(*linodego.Error); ok && (lerr.Code == 404 || lerr.Code == 405) {
```
IPv6 range GET API would return 405 when the resource is missing or not under this account.

linode/ipv6range/framework_resource.go
```diff
@@ -0,0 +1,217 @@
...
+			)
+			return
+		}
+
+		resp.Diagnostics.Append(plan.parseIPv6RangeResourceDataComputedAttrs(ctx, ipv6range)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
zliang-akamai, line 187:
```suggestion
	} else {
		//set the unknown values here, instead of use state for unknown in the schema for a value that could be change during an update.
	}
	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```

zliang-akamai (approved):
Nice work!

PR #933, opened 2023-07-05
Technically correct the CURD of ssh key and object storage key 

zliang-akamai:
Although this made no difference in the real use case, this is to align with Terraform Framework's recommendations and requirements, and is a redundancy to prevent errors. It can also serve as an example of how to sort things out correctly with the framework.

> An error is returned unless every null or known value in the request plan is saved exactly as-is into the response state. Only unknown plan values can be modified.

> Any attribute that does not have Computed set to true cannot be influenced by the provider; it can only be changed via the config or via state refreshing.

https://developer.hashicorp.com/terraform/plugin/framework/handling-data/schemas#computed

### DIFF
--- a/linode/objkey/framework_model.go
+++ b/linode/objkey/framework_model.go
@@ -22,18 +22,8 @@ type ResourceModel struct {
 	BucketAccess []BucketAccessModelEntry `tfsdk:"bucket_access"`
 }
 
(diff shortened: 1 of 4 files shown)


lgarber-akamai (approved):
Related tests are passing as expected; great work and thanks for addressing this!

yec-akamai (approved):

PR #934, opened 2023-07-05
new: Add `skip_implicit_reboots` provider argument

lgarber-akamai:
## 📝 Description

This change adds a new `skip_implicit_reboots` provider argument that allows users to skip implicit Linode reboots when altering instance configs and interfaces.

## ✔️ How to Test

```
```

### DIFF
--- a/linode/instance/resource.go
+++ b/linode/instance/resource.go
@@ -644,7 +644,9 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
 	}
 
 	// Only reboot the instance if implicit reboots are not skipped
-	rebootInstance = rebootInstance && !meta.(*helper.ProviderMeta).Config.SkipImplicitReboots
+	if !meta.(*helper.ProviderMeta).Config.SkipImplicitReboots {
+		rebootInstance = false
+	}
 
 	if rebootInstance && len(diskIDLabelMap) > 0 && len(updatedConfigMap) > 0 && bootConfig > 0 {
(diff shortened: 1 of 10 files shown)


yec-akamai (approved):

linode/instance/resource.go
```diff
@@ -643,6 +643,9 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
 		rebootInstance = false
 	}
 
+	// Only reboot the instance if implicit reboots are not skipped
+	rebootInstance = rebootInstance && !meta.(*helper.ProviderMeta).Config.SkipImplicitReboots
```
zliang-akamai, line 647:
```suggestion
	if !meta.(*helper.ProviderMeta).Config.SkipImplicitReboots {
		rebootInstance = false
	}
```

Just an optional suggestion to align with [the pattern above](https://github.com/linode/terraform-provider-linode/pull/934/files#diff-592f999d8f45773b8b2578d21ecb5d971a279e3ef3903c6a3188202235e92049L642-L644).

zliang-akamai (approved):
Nice work!

linode/instance/resource.go
yec-akamai: Unfortunately the new pattern reversed the origin logic: when `skip_implicity_reboots` set as `true` in the config, we should set the `rebootInstance` to be `false`, instead of what we have now. @ykim-1 will put up a fix to remove the `!` in the statement.

PR #942, opened 2023-07-13
fix: `TestConstructFilterString` failed due to missing filter config `TypeFunc`

yec-akamai:
## 📝 Description

The test case `TestConstructFilterString` was failing because `TypeFunc` was missing in `testFilterConfig`. Added it to all existing fields in the filter config, and also added test cases to make sure `int` and `bool` type are correctly converted as well.

## ✔️ How to Test

`make PKG_NAME="linode/helper/frameworkfilter" testacc`


zliang-akamai:
Test run: https://github.com/linode/terraform-provider-linode/actions/runs/5547790966

lgarber-akamai (approved):

PR #947, opened 2023-07-17
Always set state in the `Update` function of `nb` resource

zliang-akamai:
[Last time I thought putting the set state in the if block was okay](https://github.com/linode/terraform-provider-linode/pull/911#discussion_r1258464203) but later I realized that this may be better because normally Terraform framework won't call the update function when the user configured values remain unchanged, but if there is a bug or intentional change to the behavior, we will still want to make sure the `Update` function is running correctly. And this can also serve as a pattern for the future resource implementation.

I apologize about going back and forth on this issue.

### DIFF
--- a/linode/nb/framework_resource.go
+++ b/linode/nb/framework_resource.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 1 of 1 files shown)


yec-akamai (approved):
Tests passed successfully! Nice work. Thanks for taking care of this implementation.

jriddle-linode (approved):

PR #965, opened 2023-07-29
`device` block implementation for linode instance config

zliang-akamai:
PR description writing in progress...


linode/instanceconfig/helper.go
```diff
@@ -60,10 +93,48 @@ func flattenInterfaces(interfaces []linodego.InstanceConfigInterface) []map[stri
...
+		return nil
+	}
+
+	// devices := devicesBlock.([]map[string]any)
+
+	for _, rawDevice := range devices {
+		device := rawDevice.(map[string]any)
+		linodeGoDevice := createDevice(device)
+
+		if deviceName, ok := device["device_name"]; ok {
```
zliang-akamai, line 125: Should we detect duplicated `device_name`?

website/docs/r/instance_config.html.markdown
zliang-akamai, line 1: Would like to get feedback on the doc change. Is it okay this way? And any improvement possible?

linode/instanceconfig/schema_resource.go
```diff
@@ -18,16 +18,27 @@ var resourceSchema = map[string]*schema.Schema{
...
-		Computed:    true,
-		MaxItems:    1,
-		Description: "A dictionary of device disks to use as a device map in a Linode’s configuration profile.",
+		Type:          schema.TypeList,
+		Elem:          &schema.Resource{Schema: devicesSchema},
+		Optional:      true,
+		Computed:      true,
+		MaxItems:      1,
+		ConflictsWith: []string{"device"},
+		Deprecated:    "Devices attribute is deprecated in flavor of `devices_map` or `device`.",
```
lgarber-akamai, line 40: I think we can drop `device_map` here :+1:

linode/instanceconfig/schema_resource.go
```diff
@@ -18,16 +18,27 @@ var resourceSchema = map[string]*schema.Schema{
...
+		Description: "The Config's label for display purposes only.",
+	},
+
+	"device": {
+		Type:          schema.TypeList,
+		Elem:          &schema.Resource{Schema: deviceV2Schema},
+		Optional:      true,
+		Computed:      true,
+		ConflictsWith: []string{"devices"},
+		Description:   "A dictionary of device disks to use as a device map in a Linode's configuration profile.",
```
lgarber-akamai, line 30: This description seems to fit `device_map` rather than `device`. 

Maybe something like this:

```
A list of assignments between disks and devices for this configuration profile.
```

website/docs/r/instance_config.html.markdown
```diff
@@ -161,6 +163,22 @@ The following attributes are available on devices:
 
   * `disk_id` - (Optional) The Disk ID to map to this `device` slot
 
+#### device (recommended)
+
+Compared with `devices`, `sda` ... `sdh` is now in the `device_name` attribute in a device block, and the block itself becomes unnamed.
```
lgarber-akamai, line 168: I think the comparison here could be moved to a deprecation notice under `devices`, and this line could be changed to something along the lines of:

```
An assignment between a disk and a configuration profile device. This block supersedes the `devices` block.
```

It might also be helpful to list the available fields for this block, even though it is a bit redundant.

lgarber-akamai (changes requested):
Great work on this, it seems to work well in my local testing. Just a few nitpick changes and we should be good to go :tada:

linode/instanceconfig/schema_resource.go
zliang-akamai: Ah, good catch!

website/docs/r/instance_config.html.markdown
zliang-akamai: Fix incoming, thank you!

linode/instanceconfig/schema_resource.go
zliang-akamai: Fix incoming, thank you!

linode/instanceconfig/helper.go
zliang-akamai: implemented

zliang-akamai:
Test run:
https://github.com/linode/terraform-provider-linode/actions/runs/5872052981/job/15922845223?pr=965

lgarber-akamai (approved):
Tested locally and everything works as expected. Nice work!

PR #967, opened 2023-08-02
new: Add `linode_firewalls` data source

lgarber-akamai:
## 📝 Description

This change introduces a new `linode_firewalls` data source used to list and filter on Linode Firewalls.

NOTE: This change deviates from the `linode_firewall` data source schema to take advantage of `ListNestedBlock` and descriptions for nested fields. We may want to move the `linode_firewall` data source over to this schema and model as a part of this PR, or alternatively use the existing `linode_firewall` schema/models for this data source. 

## ✔️ How to Test

```bash
make PKG_NAME=linode/firewalls testacc
```


amisiorek-akamai (approved):
Looks good to me!

linode/firewalls/framework_schema.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/linode/terraform-provider-linode/linode/helper"
+	"github.com/linode/terraform-provider-linode/linode/helper/frameworkfilter"
+)
+
+var filterConfig = frameworkfilter.Config{
+	"id":    {APIFilterable: true, TypeFunc: helper.FilterTypeInt},
+	"label": {APIFilterable: true, TypeFunc: helper.FilterTypeString},
+	"tags":  {APIFilterable: true, TypeFunc: helper.FilterTypeString},
+
+	"status": {APIFilterable: false, TypeFunc: helper.FilterTypeString},
```
zliang-akamai, line 15: `created` and `updated` are also API filterable
https://www.linode.com/docs/api/networking/#firewalls-list__responses

zliang-akamai (approved):

lgarber-akamai:
Re-requesting reviews due to the changes made to support filtering on `created` and `updated`.

yec-akamai (approved):
Nice work! I believe it unblocks a good amount of other list data sources 👍 Just a small question for my understanding.

linode/helper/frameworkfilter/filter.go
```diff
@@ -45,6 +45,12 @@ type FilterAttribute struct {
 
 	// Converts the filter string to the correct type.
 	TypeFunc FilterTypeFunc
+
+	// (Optional) Allows this attribute during validation
+	// for the order and order_by fields.
+	// This is useful for edge cases where we want order
+	// through the API but filter on the client.
+	AllowOrder bool
```
zliang-akamai, line 53: Can we always allow order so this flag won't be necessary? Any case ordering is not allowed or not appropriated?

linode/sshkey/framework_resource_schema.go
```diff
@@ -34,9 +33,6 @@ var frameworkResourceSchema = schema.Schema{
 		"id": schema.Int64Attribute{
 			Description: "The unique identifier for this SSH key.",
 			Computed:    true,
-			PlanModifiers: []planmodifier.Int64{
-				int64planmodifier.UseStateForUnknown(),
-			},
```
zliang-akamai, line 39: Was this causing some trouble?

linode/helper/frameworkfilter/filter.go
lgarber-akamai: I added this flag so we could keep the client-side `order_by` field validation at plan-time to prevent any unexpected failures during apply. It might be a good idea to defer this logic to the API, though.

linode/sshkey/framework_resource_schema.go
lgarber-akamai: Good catch! Looks like this was just an irrelevant change that I forgot to stash. I'll push up a commit to revert it.

linode/helper/frameworkfilter/filter.go
zliang-akamai: So for the existing data source/resource without `AllowOrder`, we should add them to their filter framework config?

lgarber-akamai: I might need to adjust the name to make the meaning of this field a bit more clear, but right now it only affects non-filterable fields that can be ordered on. I don't think we'll need to make changes to other resources regardless though 👍

zliang-akamai (approved):
Looks good to me!

PR #968, opened 2023-08-02
Adding `linode_kernels` data source

amisiorek-akamai:
## 📝 Description

Adds a kernels data source that allows users to filter kernels.

Note: currently `linode_kernel` exists only as a data source, not a resource, this is not the case for `linode_nodebalancer` and `linode_user`.

## ✔️ How to Test

make PKG_NAME=linode/kernels TESTARGS="-run TestAccDataSourceKernels_basic" testacc

### DIFF
--- a/linode/kernel/framework_datasource_schema.go
+++ b/linode/kernel/framework_datasource_schema.go
@@ -5,39 +5,39 @@ import (
 )
 
 var KernelAttributes = map[string]schema.Attribute{
-		"id": schema.StringAttribute{
-			Description: "The unique ID of this Kernel.",
-			Required:    true,
-		},
-		"architecture": schema.StringAttribute{
-			Description: "The architecture of this Kernel.",
-			Computed:    true,
-		},
-		"deprecated": schema.BoolAttribute{
-			Description: "Whether or not this Kernel is deprecated.",
-			Computed:    true,
-		},
-		"kvm": schema.BoolAttribute{
-			Description: "If this Kernel is suitable for KVM Linodes.",
-			Computed:    true,
-		},
-		"label": schema.StringAttribute{
-			Description: "The friendly name of this Kernel.",
-			Computed:    true,
-		},
-		"pvops": schema.BoolAttribute{
-			Description: "If this Kernel is suitable for paravirtualized operations.",
-			Computed:    true,
-		},
-		"version": schema.StringAttribute{
-			Description: "Linux Kernel version.",
-			Computed:    true,
-		},
-		"xen": schema.BoolAttribute{
-			Description: "If this Kernel is suitable for Xen Linodes.",
-			Computed:    true,
-		},
-	}
+	"id": schema.StringAttribute{
+		Description: "The unique ID of this Kernel.",
+		Required:    true,
+	},
+	"architecture": schema.StringAttribute{
+		Description: "The architecture of this Kernel.",
+		Computed:    true,
+	},
+	"deprecated": schema.BoolAttribute{
+		Description: "Whether or not this Kernel is deprecated.",
+		Computed:    true,
+	},
+	"kvm": schema.BoolAttribute{
+		Description: "If this Kernel is suitable for KVM Linodes.",
+		Computed:    true,
+	},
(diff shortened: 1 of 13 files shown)


linode/kernel/framework_datasource_schema.go
```diff
@@ -4,39 +4,41 @@ import (
...
-		},
-		"version": schema.StringAttribute{
-			Description: "Linux Kernel version.",
-			Computed:    true,
-		},
-		"xen": schema.BoolAttribute{
-			Description: "If this Kernel is suitable for Xen Linodes.",
-			Computed:    true,
-		},
+var KernelAttributes = map[string]schema.Attribute{
```
yec-akamai, line 7: https://www.linode.com/docs/api/linode-instances/#kernels-list According to the api doc, we should also have `built` in the schema to show the date that the kernel was built.

yec-akamai (commented):
Great work! Just a small thing to fix.

yec-akamai:
Could you also run the integration test using github action? You can go to https://github.com/linode/terraform-provider-linode/actions/workflows/integration_tests_pr.yml and start `run workflow`. Just need to input your branch, the module `linode/kernels`, the hash value of your latest commit, and the PR number.

linode/kernel/framework_datasource_schema.go
amisiorek-akamai: Oh good catch, I'll do that now.

linode/kernels/tmpl/data_filter.gotf
```diff
@@ -0,0 +1,12 @@
+{{ define "kernels_data_filter" }}
+
+{{ template "kernels_data_base" . }}
+
+data "linode_kernels" "kernels" {
+    filter {
```
lgarber-akamai, line 6: Could additional filters be added here for each filterable fields?

amisiorek-akamai: Yes, done!

linode/kernels/tmpl/data_basic.gotf
zliang-akamai, line 1: Is this data source necessary for the testing? Can't we just pass the ID to `linode_kernels` data sources directly?

amisiorek-akamai: My understanding is that the ID we would pass into `linode_kernels` would be the ID for the filtered list, not for an individual kernel, which is what I wanted to test here.

zliang-akamai: Ah sorry, I putted the commend into the wrong file. I meant is `kernels_data_base` necessary for these tests?

linode/kernel/framework_models.go
```diff
@@ -0,0 +1,36 @@
...
+	Version      types.String                       `tfsdk:"version"`
+	XEN          types.Bool                         `tfsdk:"xen"`
+}
+
+func (data *DataSourceModel) ParseKernel(ctx context.Context, kernel *linodego.LinodeKernel) {
+	data.ID = types.StringValue(kernel.ID)
+	data.Architecture = types.StringValue(kernel.Architecture)
+	data.Built = customtypes.RFC3339TimeStringValue{
+		StringValue: types.StringValue(kernel.Built.Format(time.RFC3339)),
+	}
```
lgarber-akamai, line 29: For certain kernels the `built` field can be null, which can cause a panic here.

Fortunately we can fix this with a quick null check:
```suggestion
	
	built := types.StringNull()
	if kernel.Built != nil {
		built = types.StringValue(kernel.Built.Format(time.RFC3339))
	}

	data.Built = customtypes.RFC3339TimeStringValue{
		StringValue: built,
	}
	
```

amisiorek-akamai: Oh thanks for the tip!

lgarber-akamai (approved):
Nice work!

zliang-akamai (approved):

zliang-akamai (approved):

zliang-akamai:
I think it should be good to merge?

PR #972, opened 2023-08-04
Add notice mark to the recommendation in IPv6 range doc

zliang-akamai:
## 📝 Description
Make it as a notice similar to:

image


yec-akamai (approved):
Nice change! It's more visible compared to the plain text. Thank you 👍

lgarber-akamai (approved):

PR #976, opened 2023-08-10
Added documentation on volumes to handle the case where you attach vi…

meconlen:
## 📝 Description

This PR updates documentation for `linode_volume`. The existing documentation implies that you should use the `linode_id` in the `linode_volume`. This is correct when you are not using a `linode_instance_config` however if you use a `linode_instance_config` the volume is associated with the linode through the config. If you use the `linode_id` in the `linode_volume` with a `linode_instance_config` then Terraform will attempt to associate the volume with the linode before the config has been created. The API does not allow this and fails. Terraform is unable to complete the request. 

I have included a complete working example based on the documentation for the `linode_instance_config`. 

## ✔️ How to Test

You can use the example in this PR to verify the documentation works. To verify the failure case you can add the `linode_id` to the `linode_volume` with 

```hcl
```

With this addition you will receive the following error from the API. 

```
Error: failed to create volume: failed to create linode volume: [400] Couldn't choose a configuration profile to add this volume to. Please add this volume to the configuration profile you want.
```

### DIFF
--- a/website/docs/r/volume.html.markdown
+++ b/website/docs/r/volume.html.markdown
@@ -62,6 +62,107 @@ resource "linode_volume" "foobar" {
 }
 ```
 
+If you use a `linode_instance_config` you should not include the `linode_id` in the `linode_volume`. The volume will be attached to the linode via the instance config. The folowing is a complete example which demonstrates the usage: 
+
+```hcl
+terraform {
+  required_providers {
+    linode = {
+      source  = "linode/linode"
+      version = "2.5.2"
+    }
+  }
+}
+
+provider "linode" {
+  # use default LINODE_TOKEN env var
+}
+
+resource "linode_instance_config" "my-config" {
+  linode_id = linode_instance.my-instance.id
+  label = "my-config"
+
+  devices {
+    sda {
+      disk_id = linode_instance_disk.boot.id
+    }
+
+    sdb {
+      disk_id = linode_instance_disk.swap.id
+    }
+
+    sdc {
+      volume_id = linode_volume.my-volume.id
+    }
+  }
+
+  helpers {
+    # Disable the updatedb helper
+    updatedb_disabled = false
+  }
+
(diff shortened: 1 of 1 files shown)


website/docs/r/volume.html.markdown
```diff
@@ -62,6 +62,106 @@ resource "linode_volume" "foobar" {
...
+      source  = "linode/linode"
+      version = "2.5.2"
+    }
+  }
+}
+
+provider "linode" {
+  # use default LINODE_TOKEN env var
+}
+
```
lgarber-akamai, line 80: I think we can drop this bit since it's covered in the top-level provider documentation

website/docs/r/volume.html.markdown
```diff
@@ -62,6 +62,106 @@ resource "linode_volume" "foobar" {
...
+  size = 512
+  filesystem = "swap"
+}
+
+# Create a volume
+resource "linode_volume" "my-volume" {
+  region = "us-southeast"
+  label      = "my-volume"
+  size       = 20
+  linode_id = linode_instance.my-instance.id
```
lgarber-akamai, line 155: This line should be good to drop since the volume is referenced in the `linode_instance_config` resource.

lgarber-akamai (changes requested):
Thanks for the contribution! Just a few small changes and we should be good to go :slightly_smiling_face:

website/docs/r/volume.html.markdown
meconlen: I think this would be fine though we would want to note above that this is no longer a complete working example.

website/docs/r/volume.html.markdown
meconlen: Line 155 can be removed but 152-154 are required.

zliang-akamai:
```suggestion
```

PR #995, opened 2023-08-31
fix: Drop DiffSuppressFunc from `linode_instance` disk and config fields

lgarber-akamai:
## 📝 Description

This change drops the `DiffSuppressFunc` from the `linode_instance` config and disk fields, which caused these fields to be unusable from other resources when implicitly created.

Additionally, this change corrects an error in the example for the `linode_image` resource.

This PR is in draft while I investigate the scope of this change.

## ✔️ How to Test

```
```


zliang-akamai (approved):

zliang-akamai:
Test run:
https://github.com/linode/terraform-provider-linode/actions/runs/6052301381

jriddle-linode (approved):

yec-akamai (approved):

PR #997, opened 2023-09-01
Update existing resources to support VPC related linodego changes

zliang-akamai:
## 📝 Description
This is to align with the breaking code changes happened in linodego for the VPC project.

## Test

`make testacc PKG_NAME=linode/instanceconfig`

`make testacc PKG_NAME=linode/instance`

### DIFF
--- a/linode/acceptance/test_util.go
+++ b/linode/acceptance/test_util.go
@@ -590,8 +590,6 @@ func GetRegionsWithCaps(capabilities []string) ([]string, error) {
 
 // GetRandomRegionWithCaps gets a random region given a list of region capabilities.
 func GetRandomRegionWithCaps(capabilities []string) (string, error) {
-	rand.Seed(time.Now().UnixNano())
-
 	regions, err := GetRegionsWithCaps(capabilities)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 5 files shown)


lgarber-akamai (approved):
Looks good to me! Maybe this should point to a VPC project branch since this breaking change will correspond to VPC support being released.

jriddle-linode (approved):
Agree with Lena need to change base.

zliang-akamai:
Thanks for changing the base branch! I forgot to do so, lol.

zliang-akamai:
Also, we might want to switch the linodego dependency to be the vpc feature branch.

yec-akamai:
> ~Also, we might want to switch the linodego dependency to be the vpc feature branch.~
> 
> But we can't do that because our branch name contains `/`, lol https://www.reddit.com/r/golang/comments/ikl3d2/gomod_replace_with_specific_branch_rather_than/jdx60q5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Can we just change the name of the vpc branch to overcome this issue?

lgarber-akamai:
@zliang-akamai For the time being maybe we could pin to the latest commit on the linodego VPC feature branch?

zliang-akamai:
@lgarber-akamai Yes, I think that would be a good idea. But if we can follow @yec-akamai's suggestion to change the branch name to be something like `project-vpc`, we can let go to do it automatically by `go get github.com/linode/linodego@project-vpc`. What do you think?

PR #1004, opened 2023-09-06
new: Support NodeBalancers in Firewall-related resources

lgarber-akamai:
## 📝 Description

This change adds support for attaching NodeBalancers to Cloud Firewalls through the `linode_firewall` and `linode_firewall_device` resources. Additionally, this change updates Firewall-related data sources to reflect the new changes.

Most of the files changed in this PR are small test adjustments.

## ✔️ How to Test

```
make PKG_NAME=linode/firewall testacc
make PKG_NAME=linode/firewalldevice testacc
```


linode/nb/framework_resource.go
```diff
@@ -52,7 +52,7 @@ func (r *Resource) Create(
 	}
 
 	if !data.Tags.IsNull() {
-		resp.Diagnostics.Append(data.Tags.ElementsAs(ctx, &createOpts.Tags, false)...)
+		resp.Diagnostics.Append(data.Tags.ElementsAs(ctx, &createOpts.Tags, true)...)
```
lgarber-akamai, line 55: This fixes an infinite diff issue that was causing tests in this PR to fail

go.mod
```diff
@@ -95,4 +95,7 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
+// TODO: Remove this replacement before merging to dev
+replace github.com/linode/linodego => github.com/linode/linodego v1.21.1-0.20230905223015-1df9dbe40b16
+
```
zliang-akamai, line 100: So this PR should target the feature branch?

lgarber-akamai: So unfortunately `replace` don't support branch names with `/` in them so I had to pin to commit hash. Fortunately this feature has been merged and released in linodego, so I can drop this replace altogether.

zliang-akamai: Oh, sorry about the confusion. I meant it said remove this before merging to dev but this PR is targeting dev, lol

lgarber-akamai: Oh oops, good catch! I've re-targeted it to the feature branch.

yec-akamai (approved):
Nice work!

linode/firewalldevice/resource_test.go
```diff
@@ -19,12 +20,12 @@ import (
...
-	if err != nil {
-		log.Fatal(err)
-	}
+	//region, err := acceptance.GetRandomRegionWithCaps([]string{"Cloud Firewall"})
+	//if err != nil {
+	//	log.Fatal(err)
+	//}
 
-	testRegion = region
+	testRegion = "us-east"
```
zliang-akamai, line 28: Since we are targeting dev, let's restore these?

linode/firewall/resource_test.go
```diff
@@ -8,45 +8,31 @@ import (
...
-	testProviders = providerMap
-
-	region, err := acceptance.GetRandomRegionWithCaps([]string{"Cloud Firewall"})
+	_, err := acceptance.GetRandomRegionWithCaps([]string{"Cloud Firewall"})
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	testRegion = region
+	testRegion = "us-east"
```
zliang-akamai, line 35: Also restore this?

linode/nb/framework_resource_schema.go
```diff
@@ -77,6 +77,7 @@ var frameworkResourceSchema = schema.Schema{
 		"tags": schema.SetAttribute{
 			ElementType: types.StringType,
 			Optional:    true,
+			Computed:    true,
```
zliang-akamai, line 80: I think `tags` isn't computed?

lgarber-akamai: Nice catch! This was related to the infinite diff mentioned above, but I was able to come up with a better solution using defaults after taking another look at this issue.

zliang-akamai (approved):

PR #1005, opened 2023-09-06
new: adding `firewall_id` field to Create

jriddle-linode:
## 📝 Description

Adding new `firewall_id` field from Nodebalancer Firewall project

## ✔️ How to Test

`make PKG_NAME="linode/nb" TESTARGS="-run TestAccResourceNodeBalancer_firewall" testacc`


linode/nb/framework_models.go
yec-akamai, line 1: We probably need to add parsing `FirewallID` in `ParseNonComputedAttrs`, so its value can be refreshed in `Read`.

linode/nb/resource_test.go
```diff
@@ -113,6 +113,44 @@ func TestAccResourceNodeBalancer_update(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config: tmpl.Basic(t, nodebalancerName, testRegion),
```
yec-akamai, line 115: Should we use  `tmpl.Firewall` to config this test case and use `tmpl.Basic` for the `update` testing instead? We can also check if the `firewall_id` has been set in this test case.

yec-akamai (commented):
Looks good! Just some small suggestions.

go.mod
```diff
@@ -94,4 +94,4 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-go 1.19
+go 1.18
```
lgarber-akamai, line 97:
```suggestion
go 1.20
```

linode/nb/framework_resource_schema.go
```diff
@@ -48,6 +48,10 @@ var frameworkResourceSchema = schema.Schema{
 			Computed: true,
 			Default:  int64default.StaticInt64(0),
 		},
+		"firewall_id": schema.Int64Attribute{
+			Description: "ID for the firewall you'd like to use with this NodeBalancer.",
+			Optional:    true,
+		},
```
lgarber-akamai, line 54: In order to prevent user confusion (expecting a change, no change made), we might want to make this field either `ForceNew` or implement read/update logic.

lgarber-akamai (changes requested):

lgarber-akamai (approved):

ykim-1 (approved):

PR #1011, opened 2023-09-08
new: Support region_prices in the linode_instance_type(s) data sources

lgarber-akamai:
## 📝 Description

This change adds support for the `region_prices` field in the `linode_instance_type` and `linode_instance_types` data sources.

## ✔️ How to Test

NOTE: No tests have been added since this feature is not yet available through the API.

```
make PKG_NAME=linode/instancetype testacc
make PKG_NAME=linode/instancetypes testacc
```


linode/instancetype/framework_models.go
```diff
@@ -137,3 +144,28 @@ func FlattenPrice(ctx context.Context, price linodego.LinodePrice) (
...
+			"monthly": types.Float64Value(float64(price.Monthly)),
+		})
+		if d.HasError() {
+			return nil, d
+		}
+
+		result[i] = obj
+	}
+
+	priceList, d := basetypes.NewListValue(
```
zliang-akamai, line 166:
```suggestion
	priceList, d := types.ListValue(
```

I think it might be nicer to use the higher level API wrapper.

zliang-akamai (approved):

amisiorek-akamai (approved):
LGTM

linode/instancetype/framework_models.go
lgarber-akamai: Oops, nice catch!

PR #1036, opened 2023-09-26
doc: Improve documentation for early access resources; display warning on API version mismatch

lgarber-akamai:
## 📝 Description

This change improves the documentation for early access resources endpoints. Additionally, this PR adds warnings that will show when the API version is not `v4beta` for an early access resource.

## ✔️ How to Test

```
make unittest
```

### DIFF
--- a/linode/helper/earlyaccess.go
+++ b/linode/helper/earlyaccess.go
@@ -1,0 +1,41 @@
+package helper
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+// AttemptWarnEarlyAccessSDKv2 will raise a warning if an SDKv2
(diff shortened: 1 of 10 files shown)


jriddle-linode (approved):
Looks good, is there perhaps a way to automate the `IsEarlyAccess` flag from our OpenAPI spec or some other endpoint?

lgarber-akamai:
> Looks good, is there perhaps a way to automate the `IsEarlyAccess` flag from our OpenAPI spec or some other endpoint?

@jriddle-linode Good question! We could definitely automate this using the OpenAPI spec, but unfortunately we don't consume it from any of our Go projects as of now and there isn't an API endpoint that exposes which endpoints are in beta. This would definitely be worth considering in the future once we start OpenAPI into our projects more, though 👍

yec-akamai (approved):
Nice work!

PR #1042, opened 2023-10-02
fix: Correct duration for `EventPollMilliseconds` in SDKv2 client

lgarber-akamai:
## 📝 Description

This change resolves an issue that caused clients created by the SDKv2 provider to use an inaccurate polling interval.

### DIFF
--- a/linode/domainzonefile/framework_datasource.go
+++ b/linode/domainzonefile/framework_datasource.go
@@ -67,7 +67,7 @@ func (d *DataSource) Read(
 		return
 	}
 
-	retry := time.Duration(d.Meta.Config.EventPollMilliseconds.ValueInt64())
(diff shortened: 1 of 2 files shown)


amisiorek-akamai (approved):

zliang-akamai (approved):

zliang-akamai:
Can we also fix it in the framework config?

lgarber-akamai:
> Can we also fix it in the framework config?

Fortunately this only seems to affect the SDKv2 config 👍

zliang-akamai:
And should we also change it here?
https://github.com/linode/terraform-provider-linode/blob/dev/linode/domainzonefile/framework_datasource.go#L70

lgarber-akamai:
> And should we also change it here? https://github.com/linode/terraform-provider-linode/blob/dev/linode/domainzonefile/framework_datasource.go#L70

Great catch, fixed!

zliang-akamai (approved):

PR #1058, opened 2023-10-07
Update go version requirenment in README

zliang-akamai:
We are now requiring go v1.20+ to build the provider


lgarber-akamai (approved):
Nice catch!

amisiorek-akamai (approved):

PR #1067, opened 2023-10-10
Migrate in-house time type to HC's implementation

zliang-akamai:
## 📝 Description

This is to migrate in-house implemented RFC3339 custom time type to HashiCorp's [official implementation](github.com/hashicorp/terraform-plugin-framework-timetypes).

## ✔️ How to Test

`make testacc PKG_NAME="linode/token ./linode/kernel ./linode/firewalls ./linode/nb ./linode/sshkey"`

A test case in `linode/nb` fails due to an unrelated reason, and is being fixed in https://github.com/linode/terraform-provider-linode/pull/1066


linode/token/framework_model.go
```diff
@@ -14,21 +14,17 @@ import (
 type ResourceModel struct {
 	Label   types.String                        `tfsdk:"label"`
 	Scopes  customtypes.LinodeScopesStringValue `tfsdk:"scopes"`
-	Expiry  customtypes.RFC3339TimeStringValue  `tfsdk:"expiry"`
-	Created customtypes.RFC3339TimeStringValue  `tfsdk:"created"`
+	Expiry  timetypes.RFC3339                   `tfsdk:"expiry"`
+	Created timetypes.RFC3339                   `tfsdk:"created"`
```
lgarber-akamai, line 18: Tangentially related to this change, it looks like the creation logic now fails if an expiry isn't explicitly defined: https://github.com/linode/terraform-provider-linode/blob/c5578aca7f01109609d4c60760f3a1ca42d73a0f/linode/token/framework_resource.go#L43C8-L55

This can be resolved with something like this:

```go
	var expiry *time.Time

	if !data.Expiry.IsNull() {
		parsedExpiry, d := data.Expiry.ValueRFC3339Time()
		resp.Diagnostics.Append(d...)
		if d.HasError() {
			return
		}

		expiry = &parsedExpiry
	}

	createOpts := linodego.TokenCreateOptions{
		Label:  data.Label.ValueString(),
		Scopes: data.Scopes.ValueString(),
		Expiry: expiry,
	}
```

lgarber-akamai (changes requested):
Just one small change, otherwise this PR looks good!

linode/token/framework_model.go
zliang-akamai: Thanks for catching that issue!

linode/token/framework_resource_schema.go
```diff
@@ -74,15 +75,15 @@ var frameworkResourceSchema = schema.Schema{
 					RequireReplacementWhenScopesChangedDescription,
 				),
 			},
-			CustomType: customtypes.RFC3339TimeStringType{},
+			CustomType: timetypes.RFC3339Type{},
```
lgarber-akamai, line 78: One more small change, expiry will also need to be computed since the API populates it with the date `2999-12-12T05:00:00Z` if null.

lgarber-akamai (approved):
Looks good!

yec-akamai (approved):

PR #1075, opened 2023-10-12
fix: Revert integer ID attributes to strings

lgarber-akamai:
## 📝 Description

This change reverts integer ID attributes to strings to unblock provider upgrades in the [Linode Crossplane Provider](https://github.com/linode/provider-linode), which relies on the Terraform provider under the hood. This change is needed because Crossplane pre-populates ID attributes in the state with empty strings, which causes Terraform to raise an error on integer IDs. 

NOTE: Since Terraform already implicitly stores IDs as strings, we should not need any form of state migrations to maintain backwards compatibility.

## ✔️ How to Test

**Integration Tests**

```
make unittest
make PKG_NAME=linode/objkey testacc
make PKG_NAME=linode/sshkey testacc
make PKG_NAME=linode/nb testacc
```

**Manual Testing**

In a dx-devenv sandbox environment:

1. Add the following configuration to `main.tf`

```terraform
resource "linode_sshkey" "foo" {
  label = "foo"
  ssh_key = "ssh-rsa foobar foobar"
}

resource "linode_object_storage_key" "foo" {
  label = "foobar"
}

  label = "foobar"
  region = "us-east"
}
```

2. Apply the configuration (`make apply`).
3. Observe the ID attributes being stored as strings in `terraform.tfstate`

### DIFF
--- a/linode/nb/framework_models.go
+++ b/linode/nb/framework_models.go
@@ -16,7 +16,7 @@ import (
 // NodeBalancerModel describes the Terraform resource data model to match the
 // resource schema.
 type NodeBalancerModel struct {
-	ID                 types.Int64       `tfsdk:"id"`
+	ID                 types.String      `tfsdk:"id"`
 	Label              types.String      `tfsdk:"label"`
 	Region             types.String      `tfsdk:"region"`
 	ClientConnThrottle types.Int64       `tfsdk:"client_conn_throttle"`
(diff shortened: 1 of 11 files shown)


yec-akamai (approved):
Nice work, all tests passed! Thanks for putting the details for manual test.

zliang-akamai (approved):

PR #1078, opened 2023-10-15
Fix `nosec` Comment for `gosec` and remove deprecated `io/ioutil` package

zliang-akamai:
## 📝 Description

- The newer version of `gosec` requires the `nosec` comment to be inline.
- A linter doesn't allow `Write([]byte(content))` and recommended `WriteString(content)` instead.
- Removed usage of deprecated `io/ioutil` package.

### DIFF
--- a/linode/acceptance/util.go
+++ b/linode/acceptance/util.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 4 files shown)


lgarber-akamai (approved):
All tests in `linode/image `and `linode/obj` pass locally, nice work!

jriddle-linode (approved):

PR #1081, opened 2023-10-16
new: Add custom type for fields that accept IPv4/IPv6 addresses

lgarber-akamai:
## 📝 Description

This change adds a custom type for fields that accept either an IPv4 or an IPv6 address (e.g. `linode_rdns.address`).

NOTE: This type differs from the types exposed in the Hashicorp-maintained [nettypes package](https://github.com/hashicorp/terraform-plugin-framework-nettypes) in that it supports both IPv4 and IPv6 fields, rather than just one or the other.

We may want to consider making a follow up PR for a new type that supports CIDR ranges.

## ✔️ How to Test

**Acceptance Testing:**

```bash
make unittest
make PKG_NAME=linode/rdns testacc
```

**Manual Testing for Semantic Equality:**

1. Pull down this change.
2. Create a configuration using a sandbox environment (e.g. dx-devenv)

```terraform
locals {
    # The address of the newly allocated range with an appended 0 (semantically equal)
    ipv6_semantically_equal = "${linode_ipv6_range.foobar.range}0"
}

    region = "us-southeast"
    type = "g6-nanode-1"
}

  prefix_length = 64
}

resource "linode_rdns" "foobar" {
    address = local.ipv6_semantically_equal
    rdns = "${replace(local.ipv6_semantically_equal, ":", "-")}.sslip.io"
}
```

3. Apply the configuration
4. Apply the configuration again. Observe no proposed changes.

**Manual Testing for Validation:**

1. Pull down this change.
2. Create a configuration using a sandbox environment (e.g. dx-devenv)

```terraform
resource "linode_rdns" "foobar" {
    address = "foo.bar.foobar"
    rdns = "linode.com"
}
```

3. Apply the configuration. Observe the validation error.


zliang-akamai (approved):

amisiorek-akamai (approved):
Works as expected on my end

Issue #1088, opened 2023-10-19
[Bug]: Timeout in replacing instance boot disk


lucaferocino:
### Terraform Version

1.4.2

### Linode Provider Version

2.9.0

### Effected Terraform Resources

linode_instance_disk

### Panic Output

No panic, just an error output:
```terraform
linode_instance_disk.boot[0]: Still creating... [6m0s elapsed]
|
| Error: failed to create linode instance disk: [002] unexpected EOF
| 
|   with linode_instance_disk.boot[0],
|   on main.tf line 77, in resource "linode_instance_disk" "boot":
|   77: resource "linode_instance_disk" "boot" {
```

### Expected Behavior

The replacement boot disk should have been created and attached to the instance.

### Actual Behavior

The existing boot disk was destroyed, but the new replacement disk failed to be created. In fact, by inspecting the events api, there seems to be no record at all of the request to create a new boot disk.

### Steps to Reproduce

In some specific circumstances (although I haven't been able to exactly pinpoint what factors play a role in this), it seems that sending a call to create a boot disk immediately after a call to destroy an existing disk results in the creation call to hang until it hits a timeout (after 6 minutes).

I was able to reproduce the issue I experienced in `terraform` in one of my stacks by just calling the linode API as follows:
```bash
curl -H "Authorization: Bearer <token>" -X DELETE https://api.linode.com/v4beta/linode/instances/1234/disks/5678 &&  curl -v -X POST -H "Content-Type: application/json" -H "Authorization: Bearer <token>" -d '{"label":"my-boot-disk-1","size":51200,"filesystem":"ext4","root_pass":<random_pwd>}' https://api.linode.com/v4/linode/instances/1234/disks | jq 
```

Typically, re-running `terraform apply` a second time after the initial failure results in the instance disk creation call to be executed correctly and for the new boot disk to be created.

lucaferocino:
I can confirm that the issue has indeed been fixed! 👍

PR #1089, opened 2023-10-19
Wait for disk deletion to complete

lucaferocino:
## 📝 Description

This PR addresses the issues with disk replacement described in https://github.com/linode/terraform-provider-linode/issues/1088.

From the events API it appears that instance disk deletion is an asynchronous process. At the moment, however, the `terraform` provider does not seem to wait for its completion before sending a request to create a new disk. This may occasionally result in the instance disk creation call to hang indefinitely (until a timeout is reached).

This PR modifies the provider so that the `deleteResource` function for the `instancedisk` resource does in fact wait for disk deletion to complete before returning.

## ✔️ How to Test

Issue typically surfaces when it takes longer than a few seconds for the instance disk deletion to complete.

```bash
make PKG_NAME="linode/instancedisk" testacc 
```

### DIFF
--- a/linode/instancedisk/resource.go
+++ b/linode/instancedisk/resource.go
@@ -278,7 +278,13 @@ func deleteResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	tflog.Info(ctx, "Deleting instance disk")
-	p, err := client.NewEventPoller(ctx, linodeID, linodego.EntityLinode, linodego.ActionDiskDelete)
+	p, err := client.NewEventPollerWithSecondary(
+		ctx,
+		linodeID,
+		linodego.EntityLinode,
+		id,
+		linodego.ActionDiskDelete,
+	)
 	if err != nil {
 		return diag.Errorf("failed to initialize event poller: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


linode/instancedisk/resource.go
```diff
@@ -279,10 +279,21 @@ func deleteResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 	tflog.Info(ctx, "Deleting instance disk")
 	err = client.DeleteInstanceDisk(ctx, linodeID, id)
```
lgarber-akamai, line 281: Looks like there's a redundant disk deletion here

jriddle-linode (approved):

jriddle-linode (approved):

linode/instancedisk/resource.go
```diff
@@ -278,11 +278,20 @@ func deleteResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	}
 
 	tflog.Info(ctx, "Deleting instance disk")
-	err = client.DeleteInstanceDisk(ctx, linodeID, id)
+	p, err := client.NewEventPoller(ctx, linodeID, linodego.EntityLinode, linodego.ActionDiskDelete)
```
zliang-akamai, line 281:
```suggestion
	p, err := client.NewEventPollerWithSecondary(
		ctx,
		linodeID,
		linodego.EntityLinode,
		id,
		linodego.ActionDiskDelete,
	)
```
Let's also check the secondary entity, which is the disk.

lgarber-akamai (approved):
LGTM, thanks for your contribution!

Issue #1056, opened 2023-10-06
[Feature]: Linode Object Endpoint URL Change


luthermonson:
### Description
The object storage team is making changes to support multiple obj clusters per region. We need to migrate resource code to rely on the bucket API returning a hostname for the bucket to generate the endpoint URLs for the S3 calls. This replaces logic relying on the `cluster` field to generate the 

Proposal:
Update all obj resources to use the hostname returned for the bucket instead of the DNS generated via the cluster field.

PR #1090, opened 2023-10-20
Adding S3 Endpoint Computed Field

luthermonson:
## 📝 Description

Adding an additional computed field to obj and objbucket to maintain their endpoint for the s3 client connection in preparation for the multi-cluster obj rollout. Today this relies on the `cluster` but that should probably be deprecated in favor for `region`. If we wish that deprecation to happen in this PR let me know.

## ✔️ How to Test
create an objbucket and an obj item and confirm the endpoint is calculated and the item is in obj

## 📷 Preview


linode/obj/resource.go
```diff
@@ -120,6 +123,11 @@ func diffResource(
 // specified bucket via the *schema.ResourceData, then it calls
 // readResource.
 func putObject(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	// put happen before reads so add an s3 endpoint compute so it's not accidently empty
```
zliang-akamai, line 126:
```suggestion
	// put happen before reads so add an s3 endpoint compute so it's not accidentally empty
```
Suggested by the linter

yec-akamai:
Looks good! Do you mind pointing the destination branch to be `dev` instead of `main`? We usually merge change into `dev` first and cut release regularly from main, unless it's a hot fix.

Do we expect to see the endpoint from data source as well? If so, we probably need to add it in the data source schema and Read function, i.e. for the obj bucket https://github.com/linode/terraform-provider-linode/blob/dev/linode/objbucket/framework_datasource_schema.go

linode/helper/objects.go
```diff
@@ -20,14 +22,37 @@ func S3ConnFromResourceData(d *schema.ResourceData) *s3.S3 {
...
-	return s3.New(sess)
+		Endpoint:    aws.String(endpoint),
+	}))
+}
+
+// ComputeS3Endpoint sets endpoint in data for the s3 connection via the cluster and bucket combination
+func ComputeS3Endpoint(ctx context.Context, d *schema.ResourceData, meta interface{}) error {
+	if d.Get("endpoint").(string) != "" {
+		return nil
+	}
```
yec-akamai, line 44: We probably can considering removing this check. Usually terraform needs to refresh the data every time when it's read by user.

linode/helper/objects.go
```diff
@@ -20,14 +22,37 @@ func S3ConnFromResourceData(d *schema.ResourceData) *s3.S3 {
...
+	cluster := d.Get("cluster").(string)
+	bucket := d.Get("bucket").(string)
+
+	linodeClient := meta.(*ProviderMeta).Client
+	b, err := linodeClient.GetObjectStorageBucket(ctx, cluster, bucket)
+	if err != nil {
+		return fmt.Errorf("failed to find the specified Linode ObjectStorageBucket: %s", err)
+	}
+
+	return d.Set("endpoint", strings.TrimPrefix(b.Hostname, fmt.Sprintf("%s.", bucket)))
```
lgarber-akamai, line 55: Just to make the code a bit more clear, I wonder if it'd make sense to return the endpoint from this function and set it in state in the `readResource` function.

linode/obj/schema_resource.go
```diff
@@ -82,6 +82,11 @@ var resourceSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Computed:    true,
 	},
+	"endpoint": {
+		Type:        schema.TypeString,
+		Description: "The endpoint for the bucket used for s3 connections.",
+		Computed:    true,
```
lgarber-akamai, line 88: We might want to also make this field optional so users can override the target endpoint.

linode/helper/objects.go
luthermonson: @lgarber-akamai and i chatted and there are some more fixes for how this is being done and ill have an updated PR soon. agree the check can be removed... this is ack'ed and resolving the convo for now and my new changes will come today

linode/obj/schema_resource.go
luthermonson: this would mean you could set the field in hcl and it would show up in the docs?

linode/helper/objects.go
luthermonson: per a conversation this is being done and the utils funcs will be rewritten a bit to return endpoints and the data sets will be done in the callers

linode/obj/schema_resource.go
lgarber-akamai: You'll still need to manually add it to the docs, but making it both optional and computed will allow the endpoint to be overridden if specified.

linode/helper/objects.go
```diff
@@ -14,20 +16,32 @@ const (
...
-// S3ConnFromResourceData builds an S3 client from the linode_object_storage_object
-// resource's access_key, secret_key, and cluster fields.
-func S3ConnFromResourceData(d *schema.ResourceData) *s3.S3 {
-	accessKey := d.Get("access_key").(string)
-	secretKey := d.Get("secret_key").(string)
-	cluster := d.Get("cluster").(string)
-
-	sess := session.New(&aws.Config{
+// S3Connection create a client for s3 from an endpoint and keys
+func S3Connection(endpoint, accessKey, secretKey string) (*s3.S3, error) {
```
lgarber-akamai, line 20: This looks much better :rocket:

linode/objbucket/resource.go
```diff
@@ -83,13 +84,23 @@ func readResource(
 			return diag.Errorf("access_key and secret_key are required to get versioning and lifecycle info")
 		}
 
-		conn := helper.S3ConnFromResourceData(d)
+		if endpoint == "" {
+			var err error
+			if endpoint, err = helper.ComputeS3Endpoint(ctx, d, meta); err != nil {
+				return diag.FromErr(err)
+			}
+		}
```
lgarber-akamai, line 92: We don't need to compute the endpoint here since it's already available in the `bucket` object above.

e.g.
```go
endpoint := strings.TrimPrefix(bucket.Hostname, fmt.Sprintf("%s.", bucket.Label))
```

It might also make sense to move this out of the if block to make sure `endpoint` is always populated :+1:

linode/objbucket/resource.go
```diff
@@ -157,19 +165,28 @@ func updateResource(
 	lifecycleChanged := d.HasChange("lifecycle_rule")
 
 	if versioningChanged || lifecycleChanged {
+		endpoint := d.Get("endpoint").(string)
```
lgarber-akamai, line 168: `endpoint` might not be populated on resource creation because `updateResource` is called directly from `createResource` before `readResource` is called. To work around this, you could set `endpoint` in the state [here](https://github.com/linode/terraform-provider-linode/blob/ad33b5b1ea82cdbc23506ee6aa99406ba55d2e52/linode/objbucket/resource.go#L141) to ensure it's populated at creation.

FWIW this resource's implementation is super hacky as a whole, so it'd probably be worth refactoring in the future :slightly_smiling_face:

lgarber-akamai (changes requested):
Nice work so far! Just a few more changes

linode/objbucket/resource.go
luthermonson: great feedback, ill change it

linode/objbucket/resource.go
```diff
@@ -74,6 +74,7 @@ func readResource(
 	// Functionality requiring direct S3 API access
 	accessKey := d.Get("access_key").(string)
 	secretKey := d.Get("secret_key").(string)
+	endpoint := d.Get("endpoint").(string)
```
lgarber-akamai, line 77: We can resolve the endpoint from the bucket hostname here:

```suggestion
	endpoint := strings.TrimPrefix(bucket.Hostname, fmt.Sprintf("%s.", bucket.Label))
```

linode/objbucket/resource.go
```diff
@@ -83,13 +84,20 @@ func readResource(
 			return diag.Errorf("access_key and secret_key are required to get versioning and lifecycle info")
 		}
 
-		conn := helper.S3ConnFromResourceData(d)
+		if endpoint == "" {
+			endpoint = strings.TrimPrefix(bucket.Hostname, fmt.Sprintf("%s.", bucket.Label))
+		}
```
lgarber-akamai, line 89: Related to the above change, this logic can be dropped 
```suggestion
```

lgarber-akamai (changes requested):
Thanks for addressing my feedback, just one more change

lgarber-akamai (approved):

linode/obj/resource.go
```diff
@@ -120,16 +154,31 @@ func diffResource(
...
+
+	bucket := d.Get("bucket").(string)
+	accessKey := d.Get("access_key").(string)
+	secretKey := d.Get("secret_key").(string)
+	key := d.Get("key").(string)
+
+	client, err := helper.S3Connection(endpoint, accessKey, secretKey)
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
yec-akamai, line 174: nit: This logic to compute endpoint seems being used in read, delete, and put. We may consider putting it into a helper function to reuse.

yec-akamai (approved):
Great work! Just an optional comment to simplify the code.

linode/obj/resource.go
luthermonson: @yec-akamai updated to go back to the connection form data concept, only tricky part is pulling the endpoint from the s3 client config instead of the local string

zliang-akamai (approved):

zliang-akamai:
Test runs:
- https://github.com/linode/terraform-provider-linode/actions/runs/6706062631
- https://github.com/linode/terraform-provider-linode/actions/runs/6706060047
- https://github.com/linode/terraform-provider-linode/actions/runs/6706058577
- https://github.com/linode/terraform-provider-linode/actions/runs/6706055641

All passed ✅ 🎉

jriddle-linode:
resolved #1056

PR #1093, opened 2023-10-24
new: Expose `id` field in nested `linode_instance` configuration profiles

lgarber-akamai:
## 📝 Description

This change adds a new `id` field to nested configuration profiles in the `linode_instance` resource, which allows users to easily reference their config by ID in other resources and outputs.

## ✔️ How to Test

Acceptance Tests:

```
```

Manual Testing:

Inside of a provider development environment (e.g. dx-devenv):

1. Add the following configuration inside of `main.tf`:

```terraform
resource "random_password" "password" {
  length  = 16
  special = true
}

  label      = "foobar"
  region     = "us-east"
  type       = "g6-nanode-1"
  image = "linode/ubuntu22.04"
  root_pass = random_password.password.result
}

output "config_id" {
  value = linode_instance.foobar.config.0.id
}
```

2. Run `terraform init -upgrade` to install the random provider.
3. Run `make apply` to apply the configuration.
4. Observe the implicit config ID in the output.


zliang-akamai (approved):
Tested locally and worked well!

amisiorek-akamai (approved):
Tested locally and works as expected!

PR #1118, opened 2023-11-07
Fix: remove unnecessary TESTARGS in integrations-test.yaml file 

ykim-1:
## 📝 Description

Unnecessary TESTARGS was causing problems in the xml file that was generated which resulted in failure in upload to TOD

## ✔️ How to Test

Need to run the workflow after and monitor TOD

## 📷 Preview


.github/workflows/integration_tests.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           timestamp=$(date +'%Y%m%d%H%M')
           report_filename="${timestamp}_terraform_test_report.log"        
           
-          if ! make TESTARGS="2>&1" testacc > "$report_filename"; then
+          if ! make TESTARGS="" testacc > "$report_filename"; then
```
zliang-akamai, line 47:
```suggestion
          if ! make testacc > "$report_filename"; then
```
Can we remove it completely like this?

vshanthe (approved):
LGTM

PR #1135, opened 2023-11-13
Adding Framework validator for label attributes

amisiorek-akamai:
## 📝 Description

Adds length and content validators for label attributes for Framework resources. Content validators are created using a new RegEx validator. All information regarding label restrictions are pulled from API code. This PR also adds PKG_NAME to `make unittest` such that we can run unit tests by folder.

## ✔️ How to Test

`make unittest PKG_NAME=linode/helper`

`make testacc`

Manual Testing:

Inside of a provider development environment (e.g. dx-devenv):

1. Add the following configuration inside of `main.tf`:

// =============================================================================
// MySQL resource
// =============================================================================
  label  = "much_much_much_much_much_too_long_of_a_label"
  region = "us-east"
  client_conn_throttle = 20
}

2. Run `terraform init -upgrade` to install the random provider.
3. Run `make apply` to apply the configuration.
5. Observe the error.


linode/databasemysql/framework_datasource_schema.go
```diff
@@ -24,6 +25,9 @@ var frameworkDatasourceSchema = schema.Schema{
 		"label": schema.StringAttribute{
 			Description: "A unique, user-defined string referring to the Managed Database.",
 			Computed:    true,
+			Validators: []validator.String{
+				stringvalidator.LengthBetween(3, 32),
```
lgarber-akamai, line 29: optional: It might also make sense to ensure that the label contains only valid characters :+1:

zliang-akamai, line 29: I think we don't have to apply the validator to computed-only attribute because the user is not allowed to config these attributes.

amisiorek-akamai: Wait does this schema not correspond to [https://www.linode.com/docs/api/databases/#managed-mysql-database-create](https://www.linode.com/docs/api/databases/#managed-mysql-database-create) ?

amisiorek-akamai: I think RegExp is the way to go, that way we won't need to adjust the function in the future, just the RegExp used. Thoughts on the current implementation?

zliang-akamai: It's a data source schema, so it correspond to the GET endpoints. In the Terraform Framework, data source schema can't be shared with resources (at least for now). 

And additionally, user can't configured a computed-only attribute (i.e. `Computed: true` and doesn't contain `Optional` or `Required` field), even in a resource schema, so the validator is not needed on these attributes.

zliang-akamai: There are multiple set of requirements of label for different product, and a single regex is impossible to serve all of them...

amisiorek-akamai: Ahhh I see, that kicks out quite a few of these no? Actually I think that leaves only two where the label is actually manually configured?

linode/helper/regex_test.go
```diff
@@ -0,0 +1,42 @@
...
+package helper_test
+
+import (
+	"testing"
+
+	"github.com/linode/terraform-provider-linode/linode/helper"
+)
+
+func TestCheckRegex(t *testing.T) {
+	pattern := "^[a-zA-Z0-9]([-_.]?[a-zA-Z0-9]+)*[a-zA-Z0-9]$"
```
lgarber-akamai, line 12: This regex pattern looks correct and should be good to implement in a validator 🚀

linode/databasemysql/framework_datasource_schema.go
zliang-akamai: Note: we have discussed and resolved this in a private chat. Resolving this issue.

linode/nb/framework_resource_schema.go
```diff
@@ -106,6 +106,9 @@ var resourceNodebalancerV0 = schema.Schema{
 		"label": schema.StringAttribute{
 			Description: "The label of the Linode NodeBalancer.",
 			Optional:    true,
+			Validators: []validator.String{
+				stringvalidator.LengthBetween(3, 32),
+			},
```
zliang-akamai, line 111: For resource like NodeBalancer, the label requirement is not specified in the doc, but it does have some limits, and Cloud Manager doesn't allow any label with special characters. Can we dig into the API's constrains and add a regex validator here?

linode/sshkey/framework_datasource_schema.go
```diff
@@ -2,13 +2,24 @@ package sshkey
...
 import (
 	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/linode/terraform-provider-linode/linode/helper"
+)
+
+const (
+	sskKeyLabelRegex = "^[a-zA-Z0-9_-]*$"
```
amisiorek-akamai, line 12: Note: should we store RegEx consts in a different file?

linode/helper/regex_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for _, str := range testInvalidStrings {
+		if regExp.MatchString(str) {
+			t.Fatal("Should not match regex")
+		}
+	}
+}
+
+func TestCheckSuccess_nbLabel(t *testing.T) {
+	pattern := "^[a-zA-Z0-9_-]*$"
```
lgarber-akamai, line 49: These regex strings might be able to be pulled from the consts defined in each relevant resource

lgarber-akamai:
E2E test run just to make sure all validation passes: https://github.com/linode/terraform-provider-linode/actions/runs/7064032235

linode/stackscript/framework_resource_schema.go
```diff
@@ -37,6 +39,9 @@ var frameworkResourceSchema = schema.Schema{
 		"label": schema.StringAttribute{
 			Description: "The StackScript's label is for display purposes only.",
 			Required:    true,
+			Validators: []validator.String{
+				stringvalidator.LengthBetween(3, 128),
+			},
```
lgarber-akamai, line 44: Just double checking that there are no special validation rules for StackScript and Token labels?

amisiorek-akamai: Not that I can see, [stackscripts](https://bits.linode.com/LinodeAPI/apinext/blob/d425f85958d5077f4890b53a64cdfe083c6b1c2b/apinext/maps/stackscript.py) only have a lengthValidator and same goes for [tokens](https://bits.linode.com/LinodeAPI/apinext/blob/d425f85958d5077f4890b53a64cdfe083c6b1c2b/apinext/maps/oauthtoken.py)

lgarber-akamai:
Acceptance test run after formatting change: https://github.com/linode/terraform-provider-linode/actions/runs/7092062739

zliang-akamai:
Looks good to me! I'm still performing some testing locally.

zliang-akamai (approved):
Tested locally and passed.

lgarber-akamai (approved):
Looks great! All failing tests aren't related to this change, so this should be good to go 🎉

PR #1137, opened 2023-11-14
VPC interface `ip_ranges` field and vpc interface support in instance resource

zliang-akamai:
## 📝 Description

- Added `ip_ranges` field.
- Supported VPC interfaces in `linode_instance` resource.
- Centralized interface expanding and flattening logic to helper package.
- Defined SSOT interface schema cross resources.

## ✔️ How to Test
make testacc PKG_NAME="linode/instance"
make testacc PKG_NAME="linode/instanceconfig"
make testacc PKG_NAME="linode/helper"


jriddle-linode (approved):
LGTM tested local

linode/instance/schema_resource.go
```diff
@@ -67,26 +67,109 @@ func resourceDeviceDisk() *schema.Resource {
...
+		"ip_ranges": {
+			Type:        schema.TypeList,
+			Description: "List of VPC IPs or IP ranges inside the VPC subnet.",
+			Optional:    true,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+				ValidateDiagFunc: validation.AnyDiag(
+					helper.SDKv2ValidateIPv4Range,
+					helper.SDKv2ValidateIPv6Range,
+				),
```
lgarber-akamai, line 138: It looks like this field isn't being returned without the extra VPC features being enabled on your account. Do we want to include it anyway or would it make sense to split this out into a separate post-beta PR?

zliang-akamai: I think we already include many things that needs extra beta in different repo, so I guess this is fine? Like this field is in Ansible, lol

makefile
```diff
@@ -7,7 +7,7 @@ MARKDOWNLINT_IMG := 06kellyjac/markdownlint-cli
 MARKDOWNLINT_TAG := 0.19.0
 
 ACCTEST_COUNT?=1
-ACCTEST_PARALLELISM?=20
+ACCTEST_PARALLELISM?=10
```
lgarber-akamai, line 10: Nice, I think this makes sense now that the DBaaS tests have been disabled 👍

lgarber-akamai (approved):
All integration tests are passing locally and all changes look good, nice work!

PR #1143, opened 2023-11-16
fix: Resolve LKE cluster creation/deletion polling errors

lgarber-akamai:
## 📝 Description

This change adds graceful handling for polling operations that are causing LKE-related tests to fail.

## ✔️ How to Test

`make PKG_NAME=linode/lke testacc`


zliang-akamai (approved):
Tested locally and passed. Nice work!

vshanthe (approved):

PR #1164, opened 2023-11-29
Use `NewEventPollerWithSecondary` for linode instance disk event polling

yec-akamai:
## 📝 Description

This PR uses `NewEventPollerWithSecondary` for event polling in linode instance disk resource to ensure that operations filter down on secondary entity ID and prevent potential race conditions. 

## ✔️ How to Test

Run the integration tests: 
`make PKG_NAME="linode/instancedisk" testacc`

### DIFF
--- a/linode/instancedisk/resource.go
+++ b/linode/instancedisk/resource.go
@@ -261,7 +261,13 @@ func deleteResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
 	if shouldShutdown {
 		tflog.Info(ctx, "Shutting down instance for disk deletion")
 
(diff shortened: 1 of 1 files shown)


zliang-akamai (approved):
Tested locally and worked well! Nice work!

lgarber-akamai (changes requested):
Could this change also be applied to some of the polling operations in `instancedisk/helper.go`?

ykim-1 (approved):
Looks good everything worked as expected locally :)

lgarber-akamai (approved):
Thanks for addressing my change requests! All tests are passing and I manually tested resize, update, and delete operations will no issues 🙂

PR #1167, opened 2023-12-01
Implement `FrameworkRebootInstance`

zliang-akamai:
## 📝 Description

Implemented a helper function for framework resource work in the near future.

## ✔️ How to Test

We can test it by running the existent resources that uses `RebootInstance` function, like `instance`.

`make testacc PKG_NAME="linode/instance"`

### DIFF
--- a/linode/helper/instance.go
+++ b/linode/helper/instance.go
@@ -22,18 +22,46 @@ const (
 var bootEvents = []linodego.EventAction{linodego.ActionLinodeBoot, linodego.ActionLinodeReboot}
 
 // set bootConfig = 0 if using existing boot config
-func RebootInstance(ctx context.Context, d *schema.ResourceData, entityID int,
+func RebootInstance(ctx context.Context, d *schema.ResourceData, linodeID int,
 	meta interface{}, bootConfig int,
 ) diag.Diagnostics {
+	client := meta.(*ProviderMeta).Client
+	ctx, cancel := context.WithTimeout(
+		ctx,
+		time.Duration(GetDeadlineSeconds(ctx, d))*time.Second,
+	)
+	defer cancel()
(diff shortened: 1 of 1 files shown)


linode/helper/instance.go
```diff
@@ -22,18 +22,46 @@ const (
...
+	defer cancel()
+	return diag.FromErr(rebootInstance(ctx, linodeID, &client, bootConfig))
+}
+
+func FrameworkRebootInstance(
+	linodeID int,
+	client *linodego.Client,
+	bootConfig int,
+	timeoutSeconds int,
+) error {
```
lgarber-akamai, line 42: Given this function is Framework-specific, it might make sense to return (or accept) the Framework diag.Diagnostics

linode/helper/instance.go
```diff
@@ -22,18 +22,46 @@ const (
...
+	ctx, cancel := context.WithTimeout(
+		ctx,
+		time.Duration(GetDeadlineSeconds(ctx, d))*time.Second,
+	)
+	defer cancel()
+	return diag.FromErr(rebootInstance(ctx, linodeID, &client, bootConfig))
+}
+
+func FrameworkRebootInstance(
+	linodeID int,
```
lgarber-akamai, line 38: Could this function also accept a context that is passed into `context.WithTimeout(...)` so log context, etc. will be inherited?

lgarber-akamai (changes requested):
Great work! Just a few small things

linode/helper/instance.go
zliang-akamai: Thanks, done!

linode/helper/instance.go
lgarber-akamai: Good point! I think inheriting the timeout from the context and removing the timeout argument might be the better pattern here, although that's definitely up for discussion

PR #1173, opened 2023-12-05
new: Add `linode_account_availabilities` data source

lgarber-akamai:
## 📝 Description

TODO

## ✔️ How to Test

## 📷 Preview


linode/accountavailabilities/datasource_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/linode/terraform-provider-linode/v2/linode/acceptance"
+	"github.com/linode/terraform-provider-linode/v2/linode/accountavailabilities/tmpl"
+)
+
+func TestAccDataSourceAccountAvailabilities_basic(t *testing.T) {
```
lgarber-akamai, line 13: I originally wanted to add a test case for filtering on the `unavailable` field, but as far as I could tell there isn't a reliable way to guarantee a service type would be unavailable in the production environment.

jriddle-linode (approved):
LGTM. Just need to figure out which package naming to go with between here and #1174

and friendly reminder to remove the temporary stuff!

zliang-akamai (approved):
Tested locally and passed!

linode/accountavailabilities/datasource_test.go
yec-akamai: Yes AFAIK it shouldn't be a service that unavailable to a customer permanently in a region 😃 And currently there isn't `unavailable` data backfilled.

PR #1174, opened 2023-12-05
new: Support `accountavailability` Data Source

yec-akamai:
## 📝 Description

Implement a new data source `accountavailability` to get resource availability to a customer in a region. It returns unavailable resources in a region to a specific customer depends on region capability.

## ✔️ How to Test

To pass the test locally, we need to set linode url and token to Alpha. Also we will need to replace the linodego to [proj/dc-get-well](https://github.com/linode/linodego/tree/proj/dc-get-well) branch.

`make PKG_NAME="linode/accountavailability" testacc`

### DIFF
--- a/linode/accountavailability/framework_datasource_schema.go
+++ b/linode/accountavailability/framework_datasource_schema.go
@@ -1,0 +1,22 @@
+package accountavailability
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var frameworkDataSourceSchema = schema.Schema{
+	Attributes: Attributes,
+}
+
+var Attributes = map[string]schema.Attribute{
+	"region": schema.StringAttribute{
+		Description: "The id of a region.",
+		Required:    true,
+	},
+	"unavailable": schema.ListAttribute{
+		Description: "The unavailable resources in a region to a specific customer.",
+		ElementType: types.StringType,
+		Computed:    true,
+	},
+}
--- a/linode/accountavailability/framework_models.go
+++ b/linode/accountavailability/framework_models.go
@@ -1,0 +1,29 @@
+package accountavailability
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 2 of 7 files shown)


linode/accountavailability/framework_datasource_schema.go
```diff
@@ -0,0 +1,22 @@
...
+var frameworkDataSourceSchema = schema.Schema{
+	Attributes: Attributes,
+}
+
+var Attributes = map[string]schema.Attribute{
+	"region": schema.StringAttribute{
+		Description: "The id of a region.",
+		Required:    true,
+	},
+	"unavailable": schema.ListAttribute{
```
lgarber-akamai, line 17: It might make sense to make this a `SetAttribute` because the API does not guarantee the order of this list 👍

linode/accountavailabilities/framework_datasource_schema.go
```diff
@@ -11,21 +11,6 @@ var filterConfig = frameworkfilter.Config{
...
-			Description: "The region of this availability entry.",
-			Computed:    true,
-		},
-		"unavailable": schema.SetAttribute{
-			ElementType: types.StringType,
-			Description: "A list of unavailable services for the current account in this region.",
-			Computed:    true,
-		},
-	},
-}
```
lgarber-akamai, line 27: Thanks for handling this!

linode/accountavailability/framework_models.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/linode/linodego"
+)
+
+type DataSourceModel struct {
+	Region      types.String `tfsdk:"region"`
+	Unavailable types.Set    `tfsdk:"unavailable"`
+}
+
+func (d *DataSourceModel) parseAvailability(
```
lgarber-akamai, line 16: If it's not too much trouble, could you also update [the filter data source model](https://github.com/linode/terraform-provider-linode/blob/proj/dc-get-well/linode/accountavailabilities/framework_models.go#L20) to consume this struct and parser?

lgarber-akamai (approved):
Works well in my manual testing and all acceptance tests seem to pass as expected!

lgarber-akamai (approved):
LGTM, thanks for updating it!

PR #1175, opened 2023-12-05
Framework migration for instance IP

zliang-akamai:
## 📝 Description

This is to migrate `linode_instance_ip` resource to be with Terraform Framework. The newer pattern of model struct and the helper functions implemented would be helpful to solve the state consistent pain we encountered in the earlier phase of the migration work.

There is only one computed and non-configurable block in the schema so [the block migration issue](https://discuss.hashicorp.com/t/set-list-attribute-migration-from-sdkv2-to-framework/56472) we had before won't applied to this case. 

## ✔️ How to Test
`make testacc PKG_NAME=linode/instanceip`


linode/helper/framework_data.go
```diff
@@ -0,0 +1,55 @@
...
+) types.Set {
+	return KeepOrUpdateValue(original, updated, preserveKnown).(types.Set)
+}
+
+func KeepOrUpdateValue(original attr.Value, updated attr.Value, preserveKnown bool) attr.Value {
+	if preserveKnown && !original.IsUnknown() {
+		return original
+	}
+	return updated
+}
```
lgarber-akamai, line 55: We can simplify this down by making `KeepOrUpdateValue` generic using `attr.Type` as a constraint 👍 
```suggestion
func KeepOrUpdateString(original types.String, updated string, preserveKnown bool) types.String {
	return KeepOrUpdateValue(original, types.StringValue(updated), preserveKnown)
}

func KeepOrUpdateInt64(original types.Int64, updated int64, preserveKnown bool) types.Int64 {
	return KeepOrUpdateValue(original, types.Int64Value(updated), preserveKnown)
}

func KeepOrUpdateBool(original types.Bool, updated bool, preserveKnown bool) types.Bool {
	return KeepOrUpdateValue(original, types.BoolValue(updated), preserveKnown)
}

func KeepOrUpdateValue[T attr.Value](original T, updated T, preserveKnown bool) T {
	if preserveKnown && !original.IsUnknown() {
		return original
	}
	return updated
}
```

linode/instanceip/framework_models.go
```diff
@@ -0,0 +1,111 @@
...
+	m.RDNS = helper.KeepOrUpdateValue(m.RDNS, other.RDNS, preserveKnown)
+
+	m.Region = helper.KeepOrUpdateValue(m.Region, other.Region, preserveKnown)
+	m.SubnetMask = helper.KeepOrUpdateValue(m.SubnetMask, other.SubnetMask, preserveKnown)
+	m.Type = helper.KeepOrUpdateValue(m.Type, other.Type, preserveKnown)
+	m.IPVPCNAT1To1 = helper.KeepOrUpdateValue(
+		m.IPVPCNAT1To1,
+		other.IPVPCNAT1To1,
+		preserveKnown,
+	)
```
lgarber-akamai, line 108: Just a thought, I wonder if we can implement this automatically using reflection 👀

lgarber-akamai (approved):
LGTM! The new model system is much more predictable and everything works as expected in my manual testing.

ykim-1 (approved):
Looks good, everything pass locally

linode/instanceip/framework_models.go
zliang-akamai: May I know some more details about this?

lgarber-akamai: Sure! I was thinking we could [iterate through all struct fields](https://stackoverflow.com/a/18927729), get the value of each field in other, pass it through `helper.KeepOrUpdateValue`, and set the corresponding value in `m`. There might be some implementation details I'm missing (especially accounting for types) but it might be worth looking into 👀

PR #1177, opened 2023-12-06
test: Add retries to `linode_object_storage_bucket` tests

lgarber-akamai:
## 📝 Description

This change adds automatic retries to all `linode_object_storage_bucket` tests in an effort to reduce the impact of flapping test cases. 

OBJ tests can be particularly problematic due to desyncs between the API and the S3 backend. Some of these tests may not be susceptible to this issue but I figured there wouldn't be harm in applying retries regardless.

## ✔️ How to Test

Automated Testing:

```
make PKG_NAME=linode/objbucket testacc
```

Manual testing is not relevant to this PR.


ykim-1 (approved):
Looks good! Thanks for adding this

yec-akamai (approved):
Tests passed locally. Nice work!

Issue #1138, opened 2023-11-14
[Feature]: Docs for VLANs


adrianteri:
### Description

Can docs in provider have cautions/warnings when creating VLANs under *interface*?

I've battled close to a week what I thought was a bug but support indicated I didn't declare a **public** interface in my resource blocks.

Today, getting some background from issues, found out at some point the **order** mattered with which interfaces were declared and created #508. Confirmed it's not the case today...

Also recently from #1062 it seems there are inconsistencies on the order you declare interfaces and how they are created and/or associated.
From testing #508 above I've declared a public interface as **second**(just like the author of #1062) however the interface is assigned to/gets `eth0` instead of `eth1`. The author of the issue indicates it's the direct opposite for them!

From [API docs](https://www.linode.com/docs/api/linode-instances/#configuration-profile-create) it's documented that for instances to be reached **publicly** `eth0` must be a public interface.

![image](https://github.com/linode/terraform-provider-linode/assets/10687877/413eeea3-79c7-408f-b3ec-a4c55496d806)

### New or Affected Terraform Resources

linode_instance

### Potential Terraform Configuration

N/A

lgarber-akamai:
Thanks for the suggestion! I've added a ticket to our internal tracker to improve the documentation around interface ordering.

zliang-akamai:
Fixed in https://github.com/linode/terraform-provider-linode/pull/1194

PR #1194, opened 2023-12-13
docs: Improve documentation around legacy-style primary interfaces

yec-akamai:
## 📝 Description

This change improves the documentation around the ordered interface system. User needs to specify their public interface as eth0 to let Linode be reachable via public Internet. 

Addressed issue #1138.


zliang-akamai:
Hi @yec-akamai, I think we now have `primary` attribute with VPC going to be public beta soon?

yec-akamai:
> Hi @yec-akamai, I think we now have `primary` attribute with VPC going to be public beta soon?

@zliang-akamai We can add it once it's officially released. Currently we may want to make the documentation clear to user and let them know that they will need to make the public interface as eth0.

zliang-akamai (approved):

amisiorek-akamai (approved):
LGTM!

PR #1217, opened 2024-01-03
Fix test by ignoring the failure and waiting for upstream fix

zliang-akamai:
## 📝 Description

This is to fix the test failure by ignoring the state different on attribute `device` during importing test. We will have to wait for an upstream fix to fix the root cause of this issue.

Upstream issue: https://github.com/hashicorp/terraform-plugin-sdk/issues/792

Thanks @lgarber-akamai for discovering the root cause of the issue!

## ✔️ How to Test
GHA run:
https://github.com/linode/terraform-provider-linode/actions/runs/7400441105

`make testacc PKG_NAME="linode/instanceconfig"`


jriddle-linode (approved):

lgarber-akamai (approved):
All relevant tests are now passing locally, nice work!

PR #1223, opened 2024-01-05
doc: Improved Instance Resource Docs

zliang-akamai:
## 📝 Description

- Add an example of using VPC directly in instance resource.
- Format HCL (terraform) code.
- Update imagesMark  to Ubuntu 22.04.
- Remove usage of `group` and mark it as deprecated, aligned with [the API doc](https://www.linode.com/docs/api/linode-instances/#linode-create__request-body-schema).
- Replace deprecated attribute `devices` with `device`.

## ✔️ How to Test

You can try to deploy these examples in the doc.


yec-akamai (approved):
Nice work on the docs!

ykim-1 (approved):

PR #1235, opened 2024-01-12
new: Migrate `lke` data source to terraform framework

yec-akamai:
## 📝 Description

Migrate `lke` data source from SDKv2 to terraform framework. Also add some fields which are missed from the old schema. 

## ✔️ How to Test

`make testacc PKG_NAME="linode/lke"`

**Manual Test:** 
1. In a sandbox environment (i.e. dx-devenv), run the following terraform script to create a LKE cluster with 3 node pools:
```
    label       = "test-lke-cluster"
    region      = "br-gru"
    k8s_version = "1.28"
    tags        = ["test"]

    pool {
        type  = "g6-standard-2"
        count = 3
    }
}
```
2. Create a `linode_lke_cluster` data source. Assume that we wan to know the status of the above resource.
```
}

output "linode_lke_cluster" {
    value = data.linode_lke_cluster.test.status
}
```
3. Observe the data source can be created and read successfully. The status value is outputed.


jriddle-linode (approved):
LGTM passes locally, however, [parseLKEPools](https://github.com/linode/terraform-provider-linode/pull/1235/files#diff-97ca544f763b8993fee82c2df2b3d2bc71e1efe0d19f74340209531f145d54a3R94) could probably be its own function

yec-akamai:
> LGTM passes locally, however, [parseLKEPools](https://github.com/linode/terraform-provider-linode/pull/1235/files#diff-97ca544f763b8993fee82c2df2b3d2bc71e1efe0d19f74340209531f145d54a3R94) could probably be its own function

The `pareseLKEPools` has a workaround when parsing autoscaler, in order to return the same result as we did in SDKv2 without introducing breaking change. So this function is probably not reusable in other models or data sources. But we can always optimize it when we think it's reusable and build it into a separated function.

vshanthe (approved):

PR #1240, opened 2024-01-18
Migrate volume resource to TF plugin framework

zliang-akamai:
## 📝 Description

- Migrate TF SDKv2 to TF Framework for `linode_volume` resource.
- Fix a test issue by ignoring `linode_id` in import verification.
- Add pointer value support in `linode/helper/framework_data.go`.
- Add timeout support to the framework base resource.
~~- Possible breaking change:~~
  ~~- `linode_id` is now `null` instead of `0` when empty.~~
    ~~- Tests have been adjusted for this change.~~
    ~~- I wonder if this is fine to have this breaking change if we clearly note it in the release note.~~

## ✔️ How to Test
This is indeed a complicated resource to test. I really appreciate your effort for reviewing and testing it.

GHA run: https://github.com/linode/terraform-provider-linode/actions/runs/7615840894

### Automated Testing

`make int-test PKG_NAME="linode/volume"`

### Manual Testing

Here are some samples of TF config. You can try attach, detach, or re-attach volume to the Linode by adding, removing, or modifying these resources around in different order.

#### Attach from Linode
```terraform
  type   = "g6-nanode-1"
  region = "us-mia"
}

  label = "boot"
  size = linode_instance.foobar.specs.0.disk

  image = "linode/ubuntu22.04"
  root_pass = "this-is-not-a-safe-p@ssword"
}

resource "linode_instance_config" "foo" {
  label = "attach-volume"
  kernel = "linode/grub2"

  device {
    device_name = "sda"
  }

  device {
    device_name = "sdb"
    volume_id = linode_volume.foobar.id
  }
}

resource "linode_volume" "foobar" {
  label  = "tf_test-4919949863847642487"
  region = "us-mia"
  tags   = ["tf_test"]
}
```

#### Attach from Volume
```terraform
  type   = "g6-nanode-1"
  region = "us-mia"
}

  label = "boot"
  size = linode_instance.foobar.specs.0.disk

  image = "linode/ubuntu22.04"
  root_pass = "this-is-not-a-safe-p@ssword"
}

resource "linode_instance_config" "foo" {
  label = "attach-volume"
  kernel = "linode/grub2"

  device {
    device_name = "sda"
  }
}
```

After Linode creation, add volume resource (adding the volume below to the first step above will cause an API 400 error, this behavior is same as the older SDKv2 provider, and we can consider the way to fix it in the next PR maybe?)
```terraform
resource "linode_volume" "foobar" {
  label  = "tf_test-4919949863847642487"
  region = "us-mia"
  tags   = ["tf_test"]
}
```

#### Unattached Volume

```terraform
resource "linode_volume" "foobar" {
  label  = "tf_test-4919949863847642487"
  region = "us-mia"
  tags   = ["tf_test"]
}
```

#### Create Volume from Existing Volume

```terraform
resource "linode_volume" "foo" {
  label  = "tf_test-volume1"
  region = "us-mia"
  tags   = ["tf_test"]
}


resource "linode_volume" "bar" {
  label  = "tf_test-volume2"
  region = "us-mia"
  source_volume_id = linode_volume.foo.id
}

```


lgarber-akamai:
@zliang-akamai Is there any chance we could use `0` rather than `null` for an empty `linode_id`. In a vacuum I think `null` is the better option but I am a bit concerned that this breaking change could cause some difficult to diagnose bugs for users (e.g. logic checking `linode_volume.foobar.linode_id == 0`).

Maybe this is a breaking change we could schedule for the next major version?

linode/helper/framework_data.go
```diff
@@ -17,6 +17,25 @@ func KeepOrUpdateBool(original types.Bool, updated bool, preserveKnown bool) typ
...
+	return KeepOrUpdateValue(original, types.StringPointerValue(updated), preserveKnown)
+}
+
+func KeepOrUpdateInt64Pointer(original types.Int64, updated *int64, preserveKnown bool) types.Int64 {
+	return KeepOrUpdateValue(original, types.Int64PointerValue(updated), preserveKnown)
+}
+
+func KeepOrUpdateIntPointer(original types.Int64, updated *int, preserveKnown bool) types.Int64 {
+	if updated == nil {
+		return KeepOrUpdateValue(original, types.Int64Null(), preserveKnown)
```
lgarber-akamai, line 30: nitpick: Just for clarity, could we have a comment here explaining that ints do not have a corresponding `types.XPointerValue`?

jriddle-linode (approved):
LGTM everything passing locally.

zliang-akamai:
@lgarber-akamai I think the breaking change is now temporarily prevented. May I get your re-review on this? Thanks!

lgarber-akamai (approved):
LGTM, thanks for addressing the breaking change!

PR #1244, opened 2024-01-22
Misc Fixes

zliang-akamai:
## 📝 Description

- Fix a typo in the description of an attribute of the instance resource schema.
- Always copy from state in framework update function to prevent any unknown values left in plan.

## ✔️ How to Test

`make int-test PKG_NAME="linode/instance"`
`make int-test PKG_NAME="linode/instanceip"`
`make int-test PKG_NAME="linode/vpc"`

It might be a little bit... to manually test all of them, but here is a sample configs:

```terraform
    image = "linode/debian12"
    label = "foo"
    type = "g6-nanode-1"
    region = "us-mia"
}

resource "linode_instance_ip" "foo" {
    linode_id = linode_instance.foo.id
    public = true
}

resource "linode_vpc" "foo" {
    label = "my-vpc"
    region = "us-mia"
}
```


lgarber-akamai (approved):
LGTM

ykim-1 (approved):
LGTM, tests all pass locally

PR #1248, opened 2024-01-25
new: Improve logging on LKE-related resources

lgarber-akamai:
## 📝 Description

This PR makes various logging improvement to all LKE-related resources.

NOTE: This pull request adds trace statements before all API requests. This may be unnecessary because of opt-in API request logs but it does make logs a good bit more readable. Any feedback would be much appreciated!

## ✔️ How to Test

The following test steps assume you have pulled this code locally.

### Manual Testing

**NOTE: You can filter the output to only include Linode provider logs using the following: `terraform apply 2> >(grep '@module=linode' >&2)`**

1. Enable trace logging for the Linode provider:

```
export TF_LOG=TRACE
export TF_LOG_PROVIDER_LINODE=TRACE
```

2. Inside of a Terraform provider sandbox environment (e.g. dx-devenv), apply the following configuration:

```terraform
# ...

  label       = "my-cluster"
  k8s_version = "1.27"
  region      = "us-mia"

  pool {
    type  = "g6-standard-1"
    count = 1
  }

  pool {
    type = "g6-standard-1"
    count = 1
  }
}
```

3. Observe the new log statements in the stderr output.
4. Apply the following configuration to make various updates to the cluster:
 
```terraform
# ...

  label       = "my-cluster-updated"
  k8s_version = "1.28"
  region      = "us-mia"

  pool {
    type = "g6-standard-2"
    count = 1
  }
}
```

5. Observe the new log statements in the stderr output.


linode/lke/framework_datasource.go
```diff
@@ -28,6 +29,7 @@ func (r *DataSource) Read(
 	req datasource.ReadRequest,
 	resp *datasource.ReadResponse,
 ) {
+
```
yec-akamai, line 32: nit: could be nice to clean it up :)

linode/lke/resource.go
yec-akamai, line 1: Should we also consider adding logging before each API call? It might be a bit redundant because we will return diagnostics when error raised from API. But also could be useful for client who wants to trace down everything. What do you think?

lgarber-akamai: Good idea! I'll get on that shortly

lgarber-akamai: @yec-akamai Just pushed up a few commits to add trace logs before API requests, let me know what you think

yec-akamai (approved):
LGTM! Thanks for improving the change 🎉

zliang-akamai (approved):
LGTM, log shows up locally. Nice work!

PR #1252, opened 2024-01-26
fix: Restore `format` target to Makefile; make minor consistency improvements

lgarber-akamai:
## 📝 Description

This change restores the `make format` target to the Makefile and makes a few other naming changes for consistency with other targets.

## ✔️ How to Test

1. Pull down this PR.
2. Run `make format`.
3. Observe the `gofumpt` has run.


Makefile
```diff
@@ -40,23 +40,24 @@ deps:
 	go generate -tags tools tools/tools.go
 
 .PHONY: format
-format: fmt vet errcheck imports
+format:
```
jriddle-linode, line 43: The idea behind format was to handle all formatting with `make format` if you want to reduce the commands run with int-test down to just `gofumpt` then what if we just changed the `int-test` and `unit-test` target to point at `fmt-check` instead of `format`?

jriddle-linode (approved):
Everything else looks good, and the change is optional.

yec-akamai (approved):
LGTM

PR #1254, opened 2024-01-27
Support Nullable VPC Fields; Fix VPC Interface Update Error in `linode_instance` Resource

zliang-akamai:
## 📝 Description

- Bump linodego version that contains the breaking change
- Handle VPC interface add/remove/swap issue in `linode_instance` resource.

### Automated testing

`make PKG_NAME=linode/instance int-test`

### Manual testing

You can try removing, swapping (order), adding VPC interface to the Linode instance to verify if the fix works. 

```terraform
resource "linode_instance" "simple_instance" {
  label           = "simple_instance1441"
  image           = "linode/ubuntu22.04"
  region          = "us-mia"
  type            = "g6-nanode-1"
  
  interface {
    purpose   = "vpc"
    subnet_id = linode_vpc_subnet.foobar.id
    ipv4 {
      vpc = "10.0.4.250"
    }
  }
  
  interface {
    purpose = "public"
  }
}

resource "linode_vpc" "foobar" {
    label = "my-vpc"
    region = "us-mia"
    description = "test description"
}

resource "linode_vpc_subnet" "foobar" {
    vpc_id = linode_vpc.foobar.id
    label = "my-subnet"
    ipv4 = "10.0.4.0/24"
}
```


linode/helper/instance.go
```diff
@@ -199,34 +199,58 @@ func ExpandInterfaceIPv4(ipv4 any) *linodego.VPCIPv4 {
...
+		}
+		if ifaceMap["ip_ranges"] != nil {
+			ranges := ifaceMap["ip_ranges"].([]interface{})
+
+			// this is for keep result.IPRanges as a nil value rather than a value of empty slice
+			// when there is not a range.
+			if len(ranges) > 0 {
+				result.IPRanges = ExpandStringList(ranges)
+			}
+		}
```
lgarber-akamai, line 253: optional: These can be condensed down a bit

```suggestion
		if ifaceMap["subnet_id"] != nil {
			if subnetId := ifaceMap["subnetId"].(int); subnetId != 0 {
				result.SubnetID = &subnetId
			}
		}

		if ifaceMap["ipv4"] != nil {
			if ipv4 := ifaceMap["ipv4"].([]any); len(ipv4) > 0 {
				result.IPv4 = ExpandInterfaceIPv4(ipv4[0])
			}
		}
		if ifaceMap["ip_ranges"] != nil {
			// this is for keep result.IPRanges as a nil value rather than a value of empty slice
			// when there is not a range.
			if ranges := ifaceMap["ip_ranges"].([]interface{}); len(ranges) > 0 {
				result.IPRanges = ExpandStringList(ranges)
			}
		}
```

linode/instance/helpers.go
```diff
@@ -239,6 +243,7 @@ func updateInstanceConfigs(
 				"body":      configUpdateOpts,
 				"config_id": existingConfig.ID,
 			})
+			tflog.Debug(ctx, "zhiwei debug", map[string]any{"value": configUpdateOpts.Interfaces})
```
lgarber-akamai, line 246: It might make sense to convert this to an actual trace statement 👍

linode/instance/resource.go
```diff
@@ -688,25 +688,110 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
...
-		}); err != nil {
+		instance, err := client.GetInstance(ctx, id)
+		if err != nil {
+			return diag.Errorf("Error fetching data about the current linode: %s", err)
+		}
+
+		if powerOffRequired && meta.(*helper.ProviderMeta).Config.SkipImplicitReboots {
+			return diag.Errorf(
+				"add, remove, and reorder a Linode VPC interface requires implicit " +
+					"reboot of the Linode, please consider set 'skip_implicit_reboots' " +
```
lgarber-akamai, line 734: nitpick:
```suggestion
					"reboot of the Linode, please consider set 'skip_implicit_reboots' " +
```

linode/instance/resource.go
```diff
@@ -688,25 +688,110 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
...
+		}
+
+		if powerOffRequired {
+			if instance.Status != linodego.InstanceOffline {
+				if err := client.ShutdownInstance(ctx, id); err != nil {
+					return diag.Errorf(
+						"failed to shutdown instance: %s", err,
+					)
+				}
+			}
```
lgarber-akamai, line 746: I don't think this will handle transient states (e.g. shutting down), but I think the `applyBootStatus` helper in `instanceconfig` should cover those cases: https://github.com/linode/terraform-provider-linode/blob/dev/linode/instanceconfig/helper.go#L180

linode/instance/resource.go
```diff
@@ -688,25 +688,110 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
...
+		// we should power on Linode after updating of the interfaces if
+		// it's currently on and booted attribute is unset by the user.
+		// Otherwise, it will stay off (if it's already off) or be handled by
+		// `handleBootedUpdate` (if booted is set to an explicit value)
+		shouldPowerOn := bootedNull && powerOffRequired && instance.Status == linodego.InstanceRunning
+
+		// reboot won't be needed if we power off the Linode during update
+		rebootInstance = !powerOffRequired
+
+		tflog.Debug(ctx, "call update config API")
```
lgarber-akamai, line 766: Could the updateOpts be included here?

linode/instance/helpers.go
zliang-akamai: Oops, this was a temp debug output, will remove

linode/helper/instance.go
```diff
@@ -199,34 +199,54 @@ func ExpandInterfaceIPv4(ipv4 any) *linodego.VPCIPv4 {
...
+				result.IPv4 = ExpandInterfaceIPv4(ipv4[0])
+			}
+		}
+		if ifaceMap["ip_ranges"] != nil {
+			// this is for keep result.IPRanges as a nil value rather than a value of empty slice
+			// when there is not a range.
+			if ranges := ifaceMap["ip_ranges"].([]interface{}); len(ranges) > 0 {
+				result.IPRanges = ExpandStringList(ranges)
+			}
+		}
```
lgarber-akamai, line 249: Sorry for the redundant review, I ran into a few cases where these type casts raise an error on nil values:

```suggestion
		if ifaceMap["subnet_id"] != nil {
			if subnetId, ok := ifaceMap["subnetId"].(int); ok && subnetId != 0 {
				result.SubnetID = &subnetId
			}
		}

		if ifaceMap["ipv4"] != nil {
			if ipv4, ok := ifaceMap["ipv4"].([]any); ok && len(ipv4) > 0 {
				result.IPv4 = ExpandInterfaceIPv4(ipv4[0])
			}
		}
		if ifaceMap["ip_ranges"] != nil {
			// this is for keep result.IPRanges as a nil value rather than a value of empty slice
			// when there is not a range.
			if ranges, ok := ifaceMap["ip_ranges"].([]interface{}); ok && len(ranges) > 0 {
				result.IPRanges = ExpandStringList(ranges)
			}
		}
```

zliang-akamai: The `nil` value was caused by incorrect key: `subnetId`. I think you can change it to `subnet_id` to fix it.

linode/instance/resource.go
zliang-akamai: I am trying to re-use some logic in `handleBootedUpdate` (by wrapping them up in a function) to handle the cases of transient state. Do you think that looks good?

linode/helper/instance.go
lgarber-akamai: Gotcha, sounds good!

linode/instance/resource.go
lgarber-akamai: Looks good to me ^^

lgarber-akamai (approved):
Works well on my end, nice work!

zliang-akamai:
Hi @lgarber-akamai, there were some last minute logic changes in the last few commits. Feel free to leave any feedback/suggestion.

PR #1260, opened 2024-01-31
new: Overhaul node pool update logic; make count optional with an autoscaler

lgarber-akamai:
## 📝 Description

This change overhauls the `linode_lke_cluster` node pool update logic to properly make changes reflected in the Terraform plan. Additionally, this PR adjusts the pool-reordering logic in the read function to only select on ID first followed by type and count. This should _hopefully_ prevent broken diffs going forward.

Previously, the node pool update logic worked by calculating diffs between the planned pools and the existing pools in the update routine. This functionality was both inconsistent and unpredictable, which has led to a lot of broken diffs and unexpected changes for users.

Unfortunately there is a caveat to the new approach: pool order is now significant because we're relying on ID values from state. This can be a bit annoying for users looking to drop a single pool (e.g. the first defined pool) because the provider will drop the last pool and resize all subsequent pools. I looked into implementing the ID-based reordering without overhauling the update logic but the unpredictable nature of the old update logic kept breaking everything.

This PR also allows users to exclude the `count` attribute from a `linode_lke_cluster.pool` if an autoscaler is defined. This change is necessary because users may not want to override their node count on every apply, and relying on the `lifecycle.ignore_changes` rule can have confusing effects on diffing.
 
Please try your hardest to break the diffing when updating node pools, we need this implementation to be as reliable as possible 🙂 

## ✔️ How to Test

The following test steps assume you have pulled down this change locally.

### E2E Testing

```bash
make PKG_NAME=linode/lke int-test
```

### Unit Testing

```bash
make unit-test
```

### Manual Testing

The following test steps assume you are using a provider sandbox environment (e.g. dx-devenv).

#### Testing Changes to Diffing and Updates

1. Apply an arbitrary LKE cluster:

```terraform

resource "linode_lke_cluster" "test-cluster" {
  label       = "my-cluster"
  k8s_version = "1.28"
  region      = "us-ord"

  pool {
    type = "g6-standard-1"
  }

  pool {
    type = "g6-standard-1"
    count = 2

    autoscaler {
      max = 3
      min = 2
    }
  }
}
```

2. Make any number of arbitrary changes to the node pool configuration.
3. Apply the configuration and ensure the planned changes are sensible.
4. Re-apply the configuration and ensure no changes are planned.

#### Testing Changes to the Count Field

1. Apply the following configuration:

```terraform
resource "linode_lke_cluster" "test-cluster" {
  label       = "my-cluster"
  k8s_version = "1.28"
  region      = "us-ord"

  pool {
    type = "g6-standard-1"
  }

  pool {
    type = "g6-standard-1"
    count = 2

    autoscaler {
      max = 3
      min = 2
    }
  }
}
```

5. Ensure the following error is raised:

```
pool.0: `count` must be defined when no autoscaler is defined
```

6. Add an autoscaler to the first node pool and apply:

```terraform
resource "linode_lke_cluster" "test-cluster" {
  label       = "my-cluster"
  k8s_version = "1.28"
  region      = "us-ord"

  pool {
    type = "g6-standard-1"

    autoscaler {
      max = 3
      min = 1
    }
  }

  pool {
    type = "g6-standard-1"
    count = 2

    autoscaler {
      max = 3
      min = 2
    }
  }
}
```

7. Observe Terraform successfully provisions the cluster as expected.
8. Navigate to the cluster in Cloud Manager. Ensure the cluster has two node pools:

- Linode 2GB; 1 node; Autoscaler 1-3
- Linode 2GB; 2 nodes; Autoscaler 2-3

9. In Cloud Manager, manually apply the following resizes:

- `Linode 2GB; 1 node; Autoscaler 1-3` => 2 nodes
- `Linode 2GB; 2 nodes; Autoscaler 2-3` => 3 nodes

10. Reapply the configuration from step 3.
11. Observe that no changes are planned for the pool without an explicit count and that the pool with a count is planned to be downsized.
12. Observe that all changes are reflected in Cloud Manager.


linode/lke/cluster.go
```diff
@@ -363,7 +375,7 @@ func matchPoolsWithSchema(pools []linodego.LKENodePool, declaredPools []interfac
 		declaredPool := declaredPool.(map[string]interface{})
 
 		for key, pool := range poolMap {
-			if pool.Count != declaredPool["count"] || pool.Type != declaredPool["type"] {
```
lgarber-akamai, line 366: Selecting on count was causing super broken diffs, ideally we should calculate similarity scores for this function.

linode/lke/cluster.go
```diff
@@ -363,7 +375,7 @@ func matchPoolsWithSchema(pools []linodego.LKENodePool, declaredPools []interfac
...
+			if pool.ID != declaredPool["id"] {
```
lgarber-akamai, line 378: Turns out we can just select on ID because `declaredPool` is pulled from state 🙂

linode/lke/cluster.go
```diff
@@ -351,28 +364,73 @@ func recycleLKECluster(ctx context.Context, meta *helper.ProviderMeta, id int, p
...
-			if pool.Count != declaredPool["count"] || pool.Type != declaredPool["type"] {
+			if declaredAutoscaler != nil && !reflect.DeepEqual(
+				*declaredAutoscaler, pool.Autoscaler,
+			) {
 				continue
 			}
 
 			result[i] = pool
-			delete(poolMap, key)
+			delete(poolMap, pool.ID)
```
lgarber-akamai, line 428: This is a bit complicated but it should fix a LOT of broken diffs on node pools by making use of the ID stored in state.

lgarber-akamai:
Integration tests run: https://github.com/linode/terraform-provider-linode/actions/runs/7732420986

docs/resources/lke_cluster.md
```diff
@@ -72,18 +66,20 @@ The following arguments are supported:
 
 ### pool
 
+~> **Notice** Due to limitations in Terraform, the order of pools in the `linode_lke_cluster` resource is treated as significant.
+For example, the removal of the first listed pool in a cluster may result in all other node pools
+being updated accordingly.
```
lgarber-akamai, line 71: I'd appreciate some feedback on the wording here, I'm not sure if it gets the point across clearly.

linode/lke/resource.go
```diff
@@ -354,3 +392,51 @@ func populateLogAttributes(ctx context.Context, d *schema.ResourceData) context.
 		"cluster_id": d.Id(),
 	})
 }
+
+// customDiffValidateOptionalCount ensures an autoscaler must be
+// defined is count is undefined.
+//
+// This validation logic is implemented as a custom diff because
+// ValidateDiagFuncs are not currently supported directly on lists.
```
lgarber-akamai, line 400: This also extends to `RequiredWith`

linode/lke/cluster_test.go
```diff
@@ -58,69 +62,93 @@ func TestReconcileLKENodePoolSpecs(t *testing.T) {
...
+				{ID: 126, Type: "g6-standard-3", Count: 8},
+				{ID: 127, Type: "g6-standard-3", Count: 2},
 			},
 			expectedToUpdate: map[int]linodego.LKENodePoolUpdateOptions{
-				123: {Count: 2}, // -1
-				124: {Count: 8}, // +1
-				126: {Count: 9}, // +5
+				123: {Count: 2},
+				124: {Count: 9},
+				126: {Count: 8},
```
lgarber-akamai, line 98: The expectation for node pool updates has changed so the counts needed to be swapped

lgarber-akamai:
Integration test run: https://github.com/linode/terraform-provider-linode/actions/runs/7746068887

docs/resources/lke_cluster.md
yec-akamai: I can get the point, but maybe adding a sample diff can help explain the situation better?

lgarber-akamai: Maybe we could add a new section explaining these limitations in detail (with examples)?

jriddle-linode (approved):
Tested manually and using int-test, every thing's running as expected. I like the idea of a example/breakdown section in the document though.

linode/lke/cluster.go
```diff
@@ -351,34 +303,94 @@ func recycleLKECluster(ctx context.Context, meta *helper.ProviderMeta, id int, p
...
+			if declaredCount != apiPool.Count {
+				continue
+			}
+
+			if (declaredAutoscaler != nil && declaredAutoscaler.Enabled) != apiPool.Autoscaler.Enabled {
+				continue
+			}
+
+			// Only compare autoscalers if the declared autoscaler is enabled
+			if declaredAutoscaler != nil && !reflect.DeepEqual(
```
yec-akamai, line 373: Should we also add `declaredAutoscaler.Enabled` to this if statement?

docs/resources/lke_cluster.md
yec-akamai: Sounds good 👍

yec-akamai (commented):
Great work!!

linode/lke/cluster.go
lgarber-akamai: I think that logic should be handled in the block above, this is just a second-pass to make sure there are no changes if an autoscaler block has been defined.

zliang-akamai (approved):
LGTM, local tests passed!

PR #1265, opened 2024-02-05
Shutdown Linode for Updating VPC Interface when Required in `linode_instance_config` Resource

zliang-akamai:
## 📝 Description

The Terraform provider doesn't currently handle VPC interfaces update on a Linode which requires a shutting down of the instance before calling the PUT API endpoint. This PR implement the logic to handle it. Some refactors of the previous VPC interfaces update logic in `linode_instance` resource is also included, for reduced repeated code.

### Automated Testing
`make PKG_NAME="linode/instanceconfig" int-test`

### Manual Testing

You can test the VPC interface by moving it around (add, remove, swap) in the following TF config.
```terraform
resource "linode_vpc" "foobar" {
    label = "test-vpc-tf"
    region = "us-mia"
}

resource "linode_vpc_subnet" "foobar" {
    vpc_id = linode_vpc.foobar.id
    label = "test-vpc-tf"
    ipv4 = "10.0.4.0/24"
}

  label = "my-config"

  device {
    device_name = "sda"
  }

  device {
    device_name = "sdb"
    disk_id = linode_instance_disk.swap.id
  }

  interface {
    purpose = "public"
  }

  interface {
    purpose = "vlan"
    label = "my-vlan"
    ipam_address = "192.0.24.11/24"
  }
  
  booted = true
  kernel = "linode/grub2"
}

  label = "boot"
  size = linode_instance.my-instance.specs.0.disk - 1024

  image = "linode/ubuntu22.04"
  root_pass = "myc00lpassegvbherhvgoewrgvol"
}

resource "linode_instance_disk" "swap" {
  label = "swap"
  size = 1024
  filesystem = "swap"
}

  label = "my-instance"
  type = "g6-nanode-1"
  region = "us-mia"
}
```

### DIFF
--- a/linode/instance/helpers.go
+++ b/linode/instance/helpers.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/linode/linodego"
 	"github.com/linode/terraform-provider-linode/v2/linode/helper"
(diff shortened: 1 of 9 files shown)


linode/instance/helpers.go
```diff
@@ -1368,3 +1369,137 @@ func instanceIPSliceToString(ips []*linodego.InstanceIP) []string {
...
+	for _, ni := range interfaces1 {
+		included = included || ni.Purpose == linodego.InterfacePurposeVPC
+	}
+
+	for _, ni := range interfaces2 {
+		included = included || ni.Purpose == linodego.InterfacePurposeVPC
+	}
+
+	return included
+}
```
lgarber-akamai, line 1387: Optional micro-optimization: 

```suggestion
// VPCInterfaceIncluded returns whether a VPC interface is included in either slice
func VPCInterfaceIncluded(
	interfaces1 []linodego.InstanceConfigInterface,
	interfaces2 []linodego.InstanceConfigInterfaceCreateOptions,
) bool {
	for _, ni := range interfaces1 {
		if ni.Purpose == linodego.InterfacePurposeVPC {
			return true
		}
	}

	for _, ni := range interfaces2 {
		if ni.Purpose == linodego.InterfacePurposeVPC {
			return true
		}
	}

	return false
}
```

linode/instance/helpers.go
```diff
@@ -1368,3 +1369,137 @@ func instanceIPSliceToString(ips []*linodego.InstanceIP) []string {
...
+
+	powerOffRequired = powerOffRequired && bootedConfigID == targetConfigID
+
+	tflog.Debug(ctx, "Updating instance config for interface changes", map[string]any{
+		"config_id": targetConfigID,
+	})
+
+	if err != nil {
+		return false, diag.Errorf("Error fetching data about the current linode: %s", err)
+	}
```
lgarber-akamai, line 1458: I think this error check might be redundant 👍

zliang-akamai: Oops, forgot to clean it up

linode/instance/helpers.go
```diff
@@ -1368,3 +1369,137 @@ func instanceIPSliceToString(ips []*linodego.InstanceIP) []string {
...
+		return diag.Errorf("failed to boot instance after VPC interface change applied: %s", err)
+	}
+	return nil
+}
+
+func ShutdownInstanceForVPCInterfaceUpdate(ctx context.Context, meta *helper.ProviderMeta, instanceID, deadlineSeconds int) diag.Diagnostics {
+	client := &meta.Client
+	if meta.Config.SkipImplicitReboots {
+		return diag.Errorf(
+			"add, remove, and reorder a Linode VPC interface requires implicit " +
```
lgarber-akamai, line 1403: nitpick:
```suggestion
			"Adding, removing, and reordering a Linode VPC interface requires the implicit " +
```

linode/instanceconfig/resource.go
```diff
@@ -271,22 +272,52 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+				ctx, meta.(*helper.ProviderMeta), linodeID, id, helper.GetDeadlineSeconds(ctx, d),
+			)
+		}
 	}
 
-	// We should not use `HasChange(...)` here because of possible mid-apply changes
-	if !d.GetRawConfig().GetAttr("booted").IsNull() {
+	if managedBoot {
 		if err := applyBootStatus(ctx, &client, inst, id, helper.GetDeadlineSeconds(ctx, d),
 			d.Get("booted").(bool)); err != nil {
```
lgarber-akamai, line 322: This seems to be hanging when `booted = true` and the interfaces are updated to add or drop a VPC. I _think_ this is because the instance is shutdown [here](https://github.com/linode/terraform-provider-linode/pull/1265/files#diff-d0c91e4235901788aabc4f154c3636efb3ad330d1091f3616df04d47b9b0c565R301) but the instance stored in `inst` isn't updated, so the `applyBootStatus` function assumes the instance is already booted and hangs [here](https://github.com/linode/terraform-provider-linode/blob/ca53a045daa502e389a990ee4f65f7627a279812/linode/instanceconfig/helper.go#L196-L201).

As a hacky fix, you could add `inst.Status = linodego.InstanceOffline` and `inst.Status = linodego.InstanceRunning` on lines 304 and 317 respectively. Alternatively you could just refresh the instance right before this call 👍

lgarber-akamai (changes requested):

linode/instanceconfig/resource.go
zliang-akamai: I think it might be better to move instance get call to `applyBootStatus` function so it can guarantee itself always work as expected. Do you think that's a good idea?

lgarber-akamai (approved):
Works perfectly on my end, nice work! Could you also add some documentation making sure users know when their instance will be rebooted?

zliang-akamai:
Sounds good, I just added it. Thanks for the suggestion! @lgarber-akamai

ykim-1 (approved):
Looks good, tested your terraform script locally without any issues

PR #1266, opened 2024-02-06
Implement Helper Fucntion for Comma Separated Import IDs

zliang-akamai:
## 📝 Description

This PR contains implementation of `BaseResourceWithMultipleIDsImport` for supporting comma separated IDs import style, e.g. `firewall_id,device_id` or `1234,4321`

### Automated Testing
`make PKG_NAME="linode/firewalldevice" int-test`

GHA Run: https://github.com/linode/terraform-provider-linode/actions/runs/7797411038

### Manual Testing
Provision a firewall with the device.

```terraform
resource "linode_firewall_device" "my_device" {
  firewall_id = linode_firewall.my_firewall.id
  entity_id = linode_instance.my_instance.id
}

resource "linode_firewall" "my_firewall" {
  label = "my_firewall"

  inbound {
    label    = "http"
    action = "ACCEPT"
    protocol  = "TCP"
    ports     = "80"
    ipv4 = ["0.0.0.0/0"]
    ipv6 = ["::/0"]
  }

  inbound_policy = "DROP"
  outbound_policy = "ACCEPT"
}

resource "linode_instance" "my_instance" {
  label      = "my_instance"
  region     = "us-mia"
  type       = "g6-nanode-1"
}
```

Then, append a empty device resource to the TF config file.

```terraform
resource "linode_firewall_device" "my_device_duplicated" {
}
```

Finally, import the resource in a shell, `terraform import linode_firewall_device.my_device_duplicated firewall_id,device_id` (replace `firewall_id` and `device_id` with real IDs, which you can find in the state file.)

It should show a success message, like:
```
linode_firewall_device.my_device_duplicated: Import prepared!
  Prepared linode_firewall_device for import
linode_firewall_device.my_device_duplicated: Refreshing state...

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```


linode/helper/framework_resource_base.go
```diff
@@ -33,6 +42,46 @@ type BaseResourceConfig struct {
 	IsEarlyAccess bool
 }
 
+type BaseResourceWithMultipleIDsImport struct {
+	BaseResource
+
+	IDNames []string
+}
```
yec-akamai: I'm wondering if it's better to create a new ImportState helper function to handle this case, instead of creating a new BaseResource. Because the ImportState func is the only difference, and we just need to maintain one function rather than a new base resource type.

 I also looked into our code base, and I think we can follow the way of `ImportStatePassthroughInt64ID` to build this func into `ImportStateWithMultipleIDs` and use it in related resources. For example, in the `firewalldevice` resource, we can do: 
```
func (r *Resource) ImportState(
	ctx context.Context,
	req resource.ImportStateRequest,
	resp *resource.ImportStateResponse,
) {
	resource.ImportStateWithMultipleIDs(ctx, []string{"firewall_id", "id"}, req, resp)
}
```

And the import function can be something like:
```
func (r *Resource) ImportStatePassthroughInt64ID(
	ctx context.Context,
	idNames []string,
	req resource.ImportStateRequest,
	resp *resource.ImportStateResponse,
) { // The logic }
```

linode/helper/framework_resource_base.go
```diff
@@ -94,8 +143,6 @@ func (r *BaseResource) Schema(
 	resp.Schema = *r.Config.Schema
 }
 
-// ImportState should be overridden for resources with
-// complex read logic (e.g. parent ID).
```
yec-akamai, line 98: May worth still keeping this comment as a reminder? Just in case we still want to override the function for other complex logic.

linode/helper/framework_resource_base.go
zliang-akamai: Good idea, and I think that works fine too. But one downside I can see is we will still have to define `ImportState` in every resource who will need the multiple IDs import. Having it in the base resource would further reduce code needed to future resources. But that's definitely open to discussion. What do you think?

linode/helper/framework_resource_base.go
zliang-akamai: Fixed, thanks!

zliang-akamai:
@yec-akamai, I added both ways in this PR and we can remove one of them once we made the decision.

linode/helper/framework_resource_base.go
lgarber-akamai: Just chiming in, I think I personally prefer the latter solution (helper function) from a readability perspective even thought it might be a bit more verbose.

yec-akamai: +1 And I feel the helper function is a bit easier to maintain. @zliang-akamai Thank you for showing us both implementation!

zliang-akamai: @lgarber-akamai @yec-akamai, thank you both! I just pushed a commit to switch to the helper function.

linode/firewalldevice/framework_resource.go
```diff
@@ -18,7 +16,7 @@ func NewResource() resource.Resource {
 		BaseResource: helper.NewBaseResource(
 			helper.BaseResourceConfig{
 				Name:   "linode_firewall_device",
-				IDType: types.StringType,
+				IDType: types.Int64Type,
```
lgarber-akamai, line 19: Unfortunately I think we need to keep this as a string for the time being due to some weird Crossplane issues (see TPT-2234). I'm not actually sure if that's still the case though 👀

zliang-akamai: As discussed, we can defer the ID type change because it's likely no longer an issue for Crossplane provider, and this is only for changing the type in the base resource which is used for the import state function, it doesn't change the type of the ID in the schema. This is actually a fix because the ID type here mismatched the ID type defined in schema, but since the import state function in the base resource is override in this resource, I think it didn't cause any bug.

linode/helper/framework_import.go
```diff
@@ -36,3 +38,74 @@ func ImportStatePassthroughInt64ID(
...
+	CastedIDs, diags := IDsTypeConverter(IDs...)
+
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	for i, id := range CastedIDs {
+		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(ImportIDNames[i]), id)...)
+	}
```
lgarber-akamai, line 110: nitpick:

```suggestion
	castedIDs, diags := IDsTypeConverter(IDs...)

	resp.Diagnostics.Append(diags...)
	if resp.Diagnostics.HasError() {
		return
	}

	for i, id := range castedIDs {
		resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root(ImportIDNames[i]), id)...)
	}
```

linode/helper/framework_import.go
```diff
@@ -36,3 +38,74 @@ func ImportStatePassthroughInt64ID(
...
+
+type IDTypeConverter func(values ...string) ([]any, diag.Diagnostics)
+
+func AllInt64(values ...string) ([]any, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	result := make([]any, len(values))
+	for i, v := range values {
+		int64Val := StringToInt64(v, &diags)
+		result[i] = int64Val
```
lgarber-akamai, line 50: nitpick:
```suggestion
		result[i] = StringToInt64(v, &diags)
```

linode/helper/framework_import.go
```diff
@@ -36,3 +38,74 @@ func ImportStatePassthroughInt64ID(
 
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, attrPath, intID)...)
 }
+
+type IDTypeConverter func(values ...string) ([]any, diag.Diagnostics)
+
+func AllInt64(values ...string) ([]any, diag.Diagnostics) {
```
lgarber-akamai, line 44: Just to make them a bit more concise, maybe it'd make sense to name the type converters something like `BulkStringToInt64`?

linode/helper/framework_import.go
```diff
@@ -36,3 +38,74 @@ func ImportStatePassthroughInt64ID(
...
+
+	var IDs []string
+
+	for _, id := range idParts {
+		if id == "" {
+			resp.Diagnostics.AddError("Unexpected Import Identifier", unexpectedIDsErrorMsg)
+			return
+		}
+		IDs = append(IDs, id)
+	}
```
lgarber-akamai, line 100: nitpick:

```suggestion
	var ids []string

	for _, id := range idParts {
		if id == "" {
			resp.Diagnostics.AddError("Unexpected Import Identifier", unexpectedIDsErrorMsg)
			return
		}
		ids = append(ids, id)
	}
```

lgarber-akamai (approved):
Tested an import on the `linode_firewall_device` resource using the following configuration and everything worked as expected, excellent work!

```terraform
resource "linode_firewall_device" "my_device" {
  firewall_id = linode_firewall.test.id
  entity_id = linode_instance.test.id
}

  label = "test"

  inbound {
    label    = "http"
    action = "ACCEPT"
    protocol  = "TCP"
    ports     = "80"
    ipv4 = ["0.0.0.0/0"]
    ipv6 = ["::/0"]
  }

  inbound_policy = "DROP"
  outbound_policy = "ACCEPT"
}

  label      = "my_instance"
  region     = "us-southeast"
  type       = "g6-standard-1"
}
```

yec-akamai (approved):
Tested and the import works successfully! Nice work 🎉

lgarber-akamai (approved):

PR #1277, opened 2024-02-08
fix: `domains` integration test

yec-akamai:
## 📝 Description

A fix to `domains` integration tests. Although we create 2 new domains for this test case, there could be more than the 2 domains existing on the test account, because we're running concurrent tests. They're created by other test cases and haven't been cleared up yet, which may result in failures in the integration test.

## Test

`make PKG_NAME="linode/domains" int-test
`


lgarber-akamai (approved):
All test are passing locally with the following:

```bash
make PKG_NAME=linode/domains int-test
```

Thanks for fixing this and great work!

jriddle-linode (approved):
`int-test` all working, nice work.

PR #1280, opened 2024-02-09
 Migrate Shared IP Resource to TF Plugin Framework

zliang-akamai:
## 📝 Description

This PR is for migrating `linode_instance_shared_ips` resource to Terraform Plugin Framework from the SDKv2.

### Automated Testing
`make PKG_NAME="linode/instancesharedips" int-test`

### Manual Testing
```terraform
resource "linode_instance_shared_ips" "share-primary" {
  linode_id = linode_instance.secondary.id
  addresses = [linode_instance_ip.primary.address]
}

resource "linode_instance_ip" "primary" {
  linode_id = linode_instance.primary.id
}

resource "linode_instance" "primary" {
  label = "node-primary"
  type = "g6-nanode-1"
  region = "us-mia"
}

resource "linode_instance" "secondary" {
  label = "node-secondary"
  type = "g6-nanode-1"
  region = "us-mia"
}
```


linode/helper/framework_conversion.go
```diff
@@ -43,6 +44,22 @@ func StringSliceToFramework(val []string) []types.String {
 	return result
 }
 
+// StringSliceToFramework converts the given string slice
+// into a framework-compatible slice of attr.Value.
+func StringSliceToFrameworkGeneric(val []string) []attr.Value {
```
yec-akamai, line 49: I'm thinking if it's better to return `[]types.String` instead of `[]attr.Value` to be more specific.

linode/helper/framework_data.go
```diff
@@ -17,6 +18,19 @@ func KeepOrUpdateBool(original types.Bool, updated bool, preserveKnown bool) typ
 	return KeepOrUpdateValue(original, types.BoolValue(updated), preserveKnown)
 }
 
+func KeepOrUpdateSet(
```
yec-akamai, line 21: Can we take in an `elementType` param to make this func more generic? Currently it only resolves a set of string and it hopefully can be used for sets with other types of elements.

linode/instancesharedips/framework_resource.go
```diff
@@ -0,0 +1,223 @@
...
+
+	var plan ResourceModel
+	client := r.Meta.Client
+
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	ctx = populateLogAttributes(ctx, plan)
```
yec-akamai, line 76: We probably can use `tflog.SetField(...)` instead of building the function `populateLogAttributes`.

linode/helper/framework_conversion.go
zliang-akamai: The function who returns `[]types.String` is [right below this one](https://github.com/linode/terraform-provider-linode/pull/1280/files#diff-6715b511a073485e1e06abd89d7985c9bd214181c6064cbb4bfb9dbd9a0a14c6R65)..

Casting to `[]attr.Value` is intentional because we sometimes need to convert `[]string` to `[]attr.Value` because `[]types.String` can't be passed as `[]attr.Value` in golang even if `types.String` implements `attr.Value`.

ykim-1 (approved):

linode/helper/framework_conversion.go
yec-akamai: The below function `FrameworkSliceToString` actually returns `[]string` which converts the framework type to the go type. I thought this function you're building is doing the revert way.

>  []types.String can't be passed as []attr.Value in Golang even if types.String implements attr.Value

I think it can work if the caller wants a `[]types.String` specifically. Just reviewed the way you called this function is from `KeepOrUpdateSet(...)`, which takes in an `[]attr.Value`. This should be the reason that this function is returning like this way.

linode/instancesharedips/framework_resource.go
yec-akamai: Got it. Looks good!

yec-akamai (approved):
LGTM. Tests passed locally!

linode/helper/framework_conversion.go
zliang-akamai: Ah, sorry, it was [the function above it](https://github.com/linode/terraform-provider-linode/pull/1280/files#diff-6715b511a073485e1e06abd89d7985c9bd214181c6064cbb4bfb9dbd9a0a14c6L32), not the one below it..

linode/instancesharedips/framework_resource.go
zliang-akamai: Actually, @yec-akamai I think you are right. Sorry about misunderstood the comment..
I will fix it in a future PR.

PR #1281, opened 2024-02-09
horrible hack to enable crossplane-provider-linode

tchinmai7:
## 📝 Description

This is required to get the crossplane provider-linode to work, we need a way to set the client&config in a framework provider 

Crossplane's upjet https://github.com/crossplane/upjet/tree/main has recently added support to allow creating resources using the terraform provider as a library. Normally, the v2 provider & framework provider's Configure() methods would be called by terraform itself, but in this case crossplane-provider-linode is the one initializing the terraform provider., and it needs to call Configure(). The pattern followed in other upjet based providers (notably AWS) is to  initialize the v2 provider, and use the same client + config to initialize the framework provider as well. The alternative is to call the framework provider's configure() method directly within upjet. I went with the `CreateFrameworkProviderWithMeta` approach to simplify things on the crossplane end (notably because i didn't know how to construct the arguments to the framework provider's configure() method) 

## ✔️ How to Test

## 📷 Preview


linode/framework_provider.go
```diff
@@ -74,6 +74,17 @@ type FrameworkProvider struct {
 	Meta            *helper.FrameworkProviderMeta
 }
 
+func CreateFrameworkProviderWithMeta(version string, meta any) provider.ProviderWithValidateConfig {
+	castMeta := meta.(*helper.ProviderMeta)
+	return &FrameworkProvider{
```
zliang-akamai, line 79: Would it be fine to type `meta` as `*helper.ProviderMeta` for the Crossplane provider?

zliang-akamai: And I think it may be more straight forward to pass in a client instead of the whole SDKv2 meta?

linode/helper/framework_provider_model.go
```diff
@@ -5,6 +5,26 @@ import (
 	"github.com/linode/linodego"
 )
 
+func GetFrameworkProviderModelFromProviderConfig(config *Config) *FrameworkProviderModel {
```
zliang-akamai, line 8:
```suggestion
func GetFrameworkProviderModelFromSDKv2ProviderConfig(config *Config) *FrameworkProviderModel {
```
Maybe more clear to state the config is from SDKv2 provider?

zliang-akamai (approved):
LGTM!

zliang-akamai (approved):
LGTM!

PR #1285, opened 2024-02-13
Migrate Parse Attrs Function to Flatten Function with Preserving Known Values Option for NodeBalancer Resource

zliang-akamai:
## 📝 Description

This PR migrates the legacy attributes parsing function to the newer flatten function to enabled support of preserving known value to better align with Terraform Framework's principles to void potential issues.

### Automated Testing
`make PKG_NAME="linode/nb" int-test`
`make PKG_NAME="linode/nb" unit-test`

Note that **you may see an irrelevant API error during the test**, for example:
```
=== NAME  TestAccResourceToken_recreative_update
    resource_test.go:113: Step 2/4 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        -       "created": "2024-02-13T01:47:45Z",
        +       "created": "2024-02-13T01:47:46Z",
          }
```
This is caused by inconsistent dates being returned from the API in GET and POST endpoints, and it's currently being fixed.

### Manual Testing
Here is a sample token resource definition you can play around with:
```terraform
    label = "mynodebalancer"
    region = "us-mia"
    client_conn_throttle = 20
}
```

### DIFF
--- a/linode/nb/framework_models.go
+++ b/linode/nb/framework_models.go
@@ -24,8 +24,8 @@ type NodeBalancerModel struct {
 	ClientConnThrottle types.Int64       `tfsdk:"client_conn_throttle"`
 	FirewallID         types.Int64       `tfsdk:"firewall_id"`
 	Hostname           types.String      `tfsdk:"hostname"`
-	Ipv4               types.String      `tfsdk:"ipv4"`
-	Ipv6               types.String      `tfsdk:"ipv6"`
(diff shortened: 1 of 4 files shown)


ykim-1 (approved):
LGTM, nb integration tests are flaky but not related to this PR

yec-akamai (approved):
Tests passed except the known issue. Nice work!

PR #1297, opened 2024-02-15
fix: Revert Framework resource ID attributes to strings

lgarber-akamai:
## 📝 Description

This change reverts the ID attributes to strings for resources and their associated data sources implemented using Framework. This should resolve various issues with downstream consumers of the Terraform provider, specifically Pulumi.

This PR also updates the shared ID import logic to support custom types per-ID rather than using a single custom type for all ID attributes.

## ✔️ How to Test

The following test steps assume you have pulled down this PR locally before running.

### Integration Tests

```bash
make PKG_NAME=linode/firewalldevice int-test
make PKG_NAME=linode/instancesharedips int-test
make PKG_NAME=linode/volume int-test
make PKG_NAME=linode/vpc int-test
make PKG_NAME=linode/vpcsubnet int-test
```

## 📷 Preview

### DIFF
--- a/linode/firewalldevice/framework_models.go
+++ b/linode/firewalldevice/framework_models.go
@@ -1,6 +1,7 @@
 package firewalldevice
 
 import (
+	"strconv"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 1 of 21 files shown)


jriddle-linode (approved):
LGTM passing locally.

zliang-akamai:
https://github.com/linode/terraform-provider-linode/pull/1297/files#diff-be3cccd401e4a31ed557fb922be10e809145e4a3c99a39b5ed83fe4afe52d9c5R21

Let's change `IDType` in `NewBaseResource(...)` call in `instancesharedips` package.

lgarber-akamai:
> https://github.com/linode/terraform-provider-linode/pull/1297/files#diff-be3cccd401e4a31ed557fb922be10e809145e4a3c99a39b5ed83fe4afe52d9c5R21
> 
> Let's change `IDType` in `NewBaseResource(...)` call in `instancesharedips` package.

Nice catch! Fixing that now

yec-akamai (approved):
LGTM! Tested import works locally 👍

PR #1300, opened 2024-02-16
Migrate deprecated `SDKVersionString()` to `runtime/debug` in standard library

zliang-akamai:
## 📝 Description

`meta.SDKVersionString()` has been deprecated, and we can use `debug.ReadBuildInfo()` from the standard library `runtime/debug` instead for getting the version of the Terraform SDK.

## ✔️ How to Test

- Enable logging
```bash
export TF_LOG=DEBUG
export TF_LOG_PATH=$PWD/log.txt 
export TF_LOG_PROVIDER_LINODE_REQUESTS=DEBUG
```

- Create some resources: one is with the Framework, and another is with the SDKv2

```terraform
resource "linode_instance" "example" {
  image     = "linode/debian12"
  region    = "us-mia"
  type      = "g6-nanode-1"
  root_pass = "rgvergnbkerognoregnore3vewvrfgbbgefdbnh"
}

resource "linode_token" "name" {
  label = "testtoken"
  scopes = "linodes:read_only"
}
```

- Inspect the `log.txt` file, and ensure both `Terraform-Plugin-Framework/1.5.0` and `Terraform-Plugin-SDK/2.32.0` are in the user agents in the log.

image

image


jriddle-linode (approved):
LGTM, works locally.

linode/helper/version.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	return ""
+}
+
+func GetSDKv2Version() string {
+	return getDepVersion("github.com/hashicorp/terraform-plugin-sdk/v2")
+}
+
+func GetFrameworkVersion() string {
```
lgarber-akamai, line 29: @zliang-akamai Could the Framework UA also be updated to consume this function? 

https://github.com/linode/terraform-provider-linode/blob/dev/linode/framework_provider_config.go#L293

lgarber-akamai (approved):
LGTM, thanks for addressing my feedback!

PR #1302, opened 2024-02-17
linode_image upload: avoid buffering the image content in memory

kierdavis:
## 📝 Description

The current implementation of creating a `linode_image` by uploading a local file involves accumulating the file's content in a `bytes.Buffer` during the upload, then computing the MD5 checksum of the buffer's content. I believe this buffer is unnecessary - the file's content can be streamed into the hasher as well as the uploader.

This drastically reduces memory consumption. Before this change I was unable to upload a ~800MB image without my computer (8GB total) running out of memory almost immediately; after this change I could successfully upload it (with about 4GB peak memory usage by terraform + terraform-provider-linode together).

## ✔️ How to Test

```
# Create an "image" to upload:
$ dd if=/dev/random of=img bs=1M count=200
$ gzip --fast img
$ du -hs img.gz
201M	img.gz

$ cat main.tf
terraform {
  required_providers {
    linode = {
      source = "linode/linode"
      version = "2.14.0"
    }
  }
}
resource "linode_image" "main" {
  label = "no-image-buffering-test"
  file_path = "./img.gz"
  file_hash = filemd5("./img.gz")
  region = "eu-west"
}

$ export LINODE_TOKEN=...
$ terraform init
$ terraform apply -auto-approve
$ terraform destroy -auto-approve
```

Observe the peak resident set size of the terraform-provider-linode process (e.g. by watching `top`/`htop`).

Without this patch => 1286M
With this patch => 973M

### DIFF
--- a/linode/image/resource.go
+++ b/linode/image/resource.go
@@ -1,7 +1,6 @@
 package image
 
 import (
-	"bytes"
 	"context"
 	"crypto/md5" // #nosec G501 -- endpoint expecting md5
 	"encoding/hex"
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):
Everything seems to work as expected in my local testing, thanks for discovering and resolving this!

PR #1307, opened 2024-02-20
new: Add CustomDiff and PlanModifier for case-insensitive sets

lgarber-akamai:
## 📝 Description

This change alters all Framework and SDKv2 resources with tags to ignore the case during diffs on the `tags` field. This is achieved through the new `CaseInsensitiveSet` CustomDiff and PlanModifier added in this PR.

These work by modifying the plan to match values in the state regardless if they are equivalent ignoring case. This allows `terraform plan` to accurately report changes certain values in the set while not reporting case changes.

Additionally, this PR resolves an issue that prevented tags from being updated to an empty set in the `linode_domain`, `linode_instance`, and `linode_lke_cluster` resources.

This PR also adds notes to the documentation making it clear that tags are fully case-insensitive.

## ✔️ How to Test

The following test steps assume you have pulled this PR down locally.

### Unit Testing

```bash
make unit-test
```

### Integration Testing

```bash
make int-test PKG_NAME=linode/domain
make int-test PKG_NAME=linode/firewall
make int-test PKG_NAME=linode/lke
make int-test PKG_NAME=linode/nb
make int-test PKG_NAME=linode/volume
```

### Manual Testing

1. In a Terraform provider sandbox (e.g. dx-devenv), apply the following configuration:

```terraform
# ...

locals {
  tags = ["foo", "bar"]
}

  label = "test"
  region = "us-mia"
  type = "g6-nanode-1"

  tags = local.tags
}

resource "linode_volume" "test" {
  label = "test"
  region = "us-mia"
  size = 20
  tags = local.tags
}

  label = "test"
  region = "us-mia"
  tags = local.tags
  k8s_version = "1.28"

  pool {
    type = "g6-standard-1"
    count = 1
  }
}

resource "linode_domain" "test" {
  domain = "${random_string.domain_name.result}.com"
  soa_email = "test@${random_string.domain_name.result}.com"
  type = "master"
  tags = local.tags
}

resource "random_string" "domain_name" {
  length = 32
  special = false
  numeric = false
}

  label = "test"
  tags = local.tags

  inbound_policy = "DROP"
  outbound_policy = "DROP"
}

resource "linode_nodebalancer" "test" {
  region = "us-mia"
  label = "test"
  tags = local.tags
}
```

2. Observe the configuration applies successfully.
3. Update the `tags` local to change the case without any other changes:

```terraform
locals {
  tags = ["fOO", "Bar"]
}
```
4. Apply the configuration and ensure no changes were planned/made.
5. Update the `tags` local to alter the tags:

```terraform
locals {
  tags = ["wow", "bar"]
}
```

6. Apply the configuration and ensure the new tags are reflected in the plan.
7. Apply the configuration again and ensure no changes were planned/made.


linode/volume/framework_schema_resource.go
```diff
@@ -88,6 +89,9 @@ var frameworkResourceSchema = schema.Schema{
 			Description: "The Linode ID where the Volume should be attached.",
 			Optional:    true,
 			Computed:    true,
+			PlanModifiers: []planmodifier.Int64{
+				int64planmodifier.UseStateForUnknown(),
```
lgarber-akamai, line 93: I was getting some weird infinite diffs & apply errors without this, figured I'd sneak it in here 🙂

lgarber-akamai:
Integration test run: https://github.com/linode/terraform-provider-linode/actions/runs/7979214236

linode/helper/customdiffs/caseinsensitiveset.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// CaseInsensitiveSet allows us to ignore case diffs on case-insensitive
+// set values (e.g. tags).
+//
+// This function could not be implemented as a DiffSuppressFunc because DiffSuppressFuncs
+// are run per-entry rather than on the set as a whole.
+//
+// NOTE: The target field must be marked as computed.
```
lgarber-akamai, line 16: This is because `diff.SetNew` only works on computed fields.

linode/nb/framework_resource_schema.go
```diff
@@ -104,23 +106,35 @@ var frameworkResourceSchema = schema.Schema{
...
+			},
 			Description: "An array of tags applied to this object. Tags are for organizational purposes only.",
 		},
 		"transfer": schema.ListAttribute{
 			Description: "Information about the amount of transfer this NodeBalancer has had so far this month.",
 			Computed:    true,
 			ElementType: TransferObjectType,
+			PlanModifiers: []planmodifier.List{
+				listplanmodifier.UseStateForUnknown(),
+			},
```
lgarber-akamai, line 129: I'm aware that we can't guarantee the values of these attributes during the plan but this prevents an infinite diff that this resource was running into (see the above comments).

We _might_ be able to implement a custom type diff suppressor to work around this but it would require a good amount of additional boilerplate and would be a bit hacky. I'd definitely appreciate any thoughts on whether this is an acceptable workaround.

Potentially related to https://github.com/hashicorp/terraform-plugin-framework/issues/898

linode/helper/planmodifiers/caseinsensitiveset.go
```diff
@@ -0,0 +1,58 @@
...
+	return "Ensures a set does not trigger diffs on planned values with different cases."
+}
+
+func (d *caseInsensitiveSetPlanModifier) MarkdownDescription(ctx context.Context) string {
+	return d.Description(ctx)
+}
+
+func (d *caseInsensitiveSetPlanModifier) PlanModifySet(ctx context.Context, req planmodifier.SetRequest, resp *planmodifier.SetResponse) {
+	oldEntryMap := make(map[string]string)
+	resultList := make([]string, 0)
```
zliang-akamai, line 32:
```suggestion
resultList := make([]string)
```
We can ignore the `0` to make it simpler

zliang-akamai: Never mind, my memory was incorrect...

zliang-akamai (approved):
LGTM, tests passed locally. Nice work!

lgarber-akamai:
> Thank you @lgarber-akamai! It's a surprise for me that plan modifier can actually modify a known value if it matches either the configured value **or the prior state**. This may help me to proceed with some discarded migration works because I previously thought a known value couldn't be modified in the plan modifier.

Nice, glad this helped! There are definitely still some caveats to modifying the state with Framework but hopefully those will disappear with time 🙂

yec-akamai (approved):
Nice work! All tests passed locally 🎉

PR #1314, opened 2024-02-21
fix: Fill gaps for Crossplane empty ID workaround

lgarber-akamai:
## 📝 Description

This PR applies the Crossplane empty ID bug workaround from #1085 to Framework resources that were missing it.

This should resolve the following issue when trying to provision certain resources (e.g. VPCs) through Crossplane:

```
read resource request failed: Failed to convert string to int: strconv.Atoi: parsing "": invalid syntax
```

Additionally, this PR adds logic to override the ID of a resource during creation to resolve an edge case where the ID would not be stored in state because `preserveKnown` treats an empty ID in state as known.

### Integration Testing

```
make int-test PKG_NAME=linode/accountsettings 
make int-test PKG_NAME=linode/firewalldevice 
make int-test PKG_NAME=linode/instanceip 
make int-test PKG_NAME=linode/instancesharedips 
make int-test PKG_NAME=linode/volume 
make int-test PKG_NAME=linode/vpc
make int-test PKG_NAME=linode/vpcsubnet
```


linode/vpc/framework_resource.go
```diff
@@ -56,6 +57,10 @@ func (r *Resource) Create(
 	}
 
 	data.FlattenVPC(ctx, vpc, true)
+
+	// IDs should always be overridden during creation (see #1085)
+	data.ID = types.StringValue(strconv.Itoa(vpc.ID))
```
lgarber-akamai, line 62: @zliang-akamai Would it make sense to instead remove the preserveKnown logic for the ID field in the `data.FlattenX(...)` methods? I'm hesitant because it might introduce some unintuitive behavior, but might simplify this workaround a bit.

zliang-akamai: Yep, both ways seem to be fine to me

zliang-akamai (approved):
LGTM! Tests passed!

yec-akamai (approved):
LGTM! Tests passed locally 🎉

PR #1320, opened 2024-02-22
Only set linode_id in log; append unhandled diags

zliang-akamai:
## 📝 Description

2 small fixes to the shared IP resource.

- `id` and `linode_id` is the same thing in this resource, thus only one is needed in the logs.
- `diags` returned from `plan.Addresses.ElementsAs` wasn't handled, it should be appended into the `diags` and checked.

## ✔️ How to Test

```
make PKG_NAME=linode/instancesharedips int-test
```

```terraform
resource "linode_instance_shared_ips" "share-primary" {
  linode_id = linode_instance.secondary.id
  addresses = [linode_instance_ip.primary.address]
}

resource "linode_instance_ip" "primary" {
  linode_id = linode_instance.primary.id
}

resource "linode_instance" "primary" {
  label = "node-primary"
  type = "g6-nanode-1"
  region = "us-mia"
}

resource "linode_instance" "secondary" {
  label = "node-secondary"
  type = "g6-nanode-1"
  region = "us-mia"
}
```

### DIFF
--- a/linode/instancesharedips/framework_resource.go
+++ b/linode/instancesharedips/framework_resource.go
@@ -41,7 +41,10 @@ func CreateOrUpdateSharedIPs(
 		LinodeID: linodeID,
 	}
 
-	plan.Addresses.ElementsAs(ctx, &createOpts.IPs, false)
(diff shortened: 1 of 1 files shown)


yec-akamai (approved):
LGTM!

PR #1331, opened 2024-02-28
new: Populate route_target in linode_ipv6_range

yec-akamai:
## 📝 Description

When user configured `linode_id` to create a ipv6 range, populate the value of `route_target` implicitly. 

## ✔️ How to Test

`make unit-test`

`make int-test PKG_NAME="linode/ipv6range"`

**Manual test:**
1. In a sandbox environment (i.e. dx-devenv), put the following configuration to create ipv6 range under a new Linode using linode_id:
```
    label = "tf-test-instance"
    type = "g6-nanode-1"
    region = "us-mia"
    booted = false
}

    prefix_length = 64
}

output "route_target" {
    value = linode_ipv6_range.foobar.route_target
}

output "linode_ipv6" {
    value = linode_instance.foobar.ipv6
}
```
2. Apply the configuration. Observe that the `route_target` is correctly populated. Its value without the prefix length equals to `linode_ipv6` the Linode’s IPv6 SLAAC address.
3. Destroy the resources for clean up.

### DIFF
--- a/linode/ipv6range/framework_resource.go
+++ b/linode/ipv6range/framework_resource.go
@@ -116,23 +116,9 @@ func (r *Resource) Create(
 		return
 	}
 
-	// when user configured linode_id, populate route_target implicitly
+	// If user configured linode_id, populate route_target implicitly.
 	if linodeIdConfigured {
-		ranges, err := client.ListIPv6Ranges(ctx, nil)
-		if err != nil {
-			resp.Diagnostics.AddError(
-				"Failed to list ipv6 ranges when create.",
-				err.Error(),
-			)
-			return
-		}
-
-		for _, r := range ranges {
-			if r.Range == data.ID.ValueString() {
-				data.RouteTarget = types.StringValue(r.RouteTarget)
-				break
-			}
-		}
+		data.RouteTarget = types.StringValue(ipv6range.RouteTarget)
 	}
 
 	resp.Diagnostics.Append(data.FlattenIPv6Range(ctx, ipv6rangeR, true)...)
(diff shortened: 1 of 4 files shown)


linode/ipv6range/framework_resource.go
```diff
@@ -116,6 +116,11 @@ func (r *Resource) Create(
 		return
 	}
 
+	// If user configured linode_id, populate route_target implicitly.
+	if linodeIdConfigured {
+		data.RouteTarget = types.StringValue(ipv6range.RouteTarget)
+	}
+
```
zliang-akamai, line 123: I think this part is not necessary because it will be populated in `FlattenIPv6Range` function, which is called below?

yec-akamai: We probably don't want `FlattenIPv6Range` to take care of RouteTarget, and we only want to compute it when linode_id is configured by user to create ipv6_range. If user create a ipv6_range using route_target, we won't need to compute it.

zliang-akamai: Hi @yec-akamai, I think `FlattenIPv6Range` was designed to handle these cases automatically. So when user create a IPv6 Range using `route_target`, `FlattenIPv6Range` will preserve the user configured value (a known value) as `true` is passed as the `preserveKnown` parameter. If the user create a IPv6 Range via Linode ID, the `route_target` will become an unknown value and will be overridden in `FlattenIPv6Range`.

zliang-akamai: Ah okay, I see the problem there... `route_target` won't be in the response of get IPv6 range API call

zliang-akamai (approved):
Works for me, looks good to me

lgarber-akamai (approved):
Integration tests are passing locally and everything seems to work as expected in my manual testing, great work!

Issue #1172, opened 2023-12-05
[Bug]: Circular dependecy for buckets and limited access keys with lifecycle policy or versioning


wbollock:
### Terraform Version

v0.14.9

### Linode Provider Version

1.29.0

### Effected Terraform Resources

linode_object_storage_bucket
linode_object_storage_key
lifecycle_rule

### Terraform Config Files

```hcl
resource "linode_object_storage_bucket" "test-bucket" {
  cluster  = "us-iad-1"
  label    = "test-bucket"

  # need access/secret key to make lifecycle policy (or to configure versioning)
  access_key = linode_object_storage_key.test-key.access_key
  secret_key = linode_object_storage_key.test-key.secret_key

  lifecycle_rule {
    id      = "thirty-days-retention"
    enabled = true
    expiration {
      days = 30
    }
  }
}

resource "linode_object_storage_key" "test-key" {
  label    = "test-key"
  bucket_access {
    # creates circular dependency as a limited access key
    bucket_name = linode_object_storage_bucket.test-buckey.label
    cluster     = "us-iad-1"
    permissions = "read_write"
  }
}
```

### Expected Behavior

I should be able to have a limited access key for my bucket that can also create lifecycle rules

### Actual Behavior

I must create a second non-limited access key that is used to just create the lifecycle versioning. The key has very broad access and I'd rather use the limited access, bucket-scoped keys instead

### Steps to Reproduce

```hcl
resource "linode_object_storage_bucket" "test-bucket" {
  cluster  = "us-iad-1"
  label    = "test-bucket"

  # need access/secret key to make lifecycle policy (or to configure versioning)
  access_key = linode_object_storage_key.test-key.access_key
  secret_key = linode_object_storage_key.test-key.secret_key

  lifecycle_rule {
    id      = "thirty-days-retention"
    enabled = true
    expiration {
      days = 30
    }
  }
}

resource "linode_object_storage_key" "test-key" {
  label    = "test-key"
  bucket_access {
    # creates circular dependency as a limited access key
    bucket_name = linode_object_storage_bucket.test-buckey.label
    cluster     = "us-iad-1"
    permissions = "read_write"
  }
}
```

yec-akamai:
Hi @wbollock ,

Thanks for bringing up this issue. It's an intended behavior for now, however, we're considering a possible way to simplify generating the key to the obj bucket. We're going to have an internal discussion on it, and will let you know if there is any update!

[...]

PR #1337, opened 2024-03-04
new: Provider-level Object Storage Keys

yec-akamai:
## 📝 Description

This change allows user to specify object storage keys at provider-level. The access and secret keys can be used in `linode_object_storage_bucket` and `linode_object_storage_object` resource. We aim to simplify the object key management and avoid issue like circular dependency. The keys then can be specified in the provider configuration, or can be exported as an environment variable. 

Addressed issue #1172.

## ✔️ How to Test

`make int-test PKG_NAME="linode/obj"`
`make int-test PKG_NAME="linode/objbucket"`

**Manual Test:**
1. In a sandbox environment, i.e. dx-devenv, use the following configuration to create an object bucket with lifecycle policy:
```
provider "linode" {
    obj_access_key = ${your-access-key}
    obj_secret_key = ${your-secret-key}
}

    cluster = "us-mia-1"
    label = "test-obj-bucket"

    lifecycle_rule {
        prefix = "tf"
        enabled = true

        abort_incomplete_multipart_upload_days = 5

        expiration {
            date = "2024-03-21"
        }
    }
}
```
2. Observe that object bucket is created successfully without any error raised.
3. Destroy the resource to clean up.
4. Test that object credentials by exporting from environment
`export LINODE_OBJ_ACCESS_KEY=$YOUR_ACCESS_KEY`
`export LINODE_OBJ_SECRET_KEY=$YOUR_SECRET_KEY`
5. Put the configuration without provider metadata
```
    cluster = "us-mia-1"
    label = "test-obj-bucket"

    lifecycle_rule {
        prefix = "tf"
        enabled = true

        abort_incomplete_multipart_upload_days = 5

        expiration {
            date = "2024-03-21"
        }
    }
}
```
6. Observe that object bucket is created successfully without any error raised.
7. Destroy the resource to clean up.

### DIFF
--- a/linode/obj/resource.go
+++ b/linode/obj/resource.go
@@ -91,16 +91,6 @@ func readResource(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 		d.Set("endpoint", endpoint)
 	}
 
-	// Compute object access credentials from provider metadata when they're not configured explicitly
-	config := meta.(*helper.ProviderMeta).Config
-	if _, ok := d.GetOk("access_key"); !ok {
-		d.Set("access_key", config.ObjAccessKey)
-	}
-
-	if _, ok := d.GetOk("secret_key"); !ok {
-		d.Set("secret_key", config.ObjSecretKey)
-	}
-
 	return nil
 }
 
--- a/linode/obj/resource.go
+++ b/linode/obj/resource.go
@@ -91,6 +91,16 @@ func readResource(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 		d.Set("endpoint", endpoint)
 	}
 
+	// Compute object access credentials from provider metadata when they're not configured explicitly
+	config := meta.(*helper.ProviderMeta).Config
+	if _, ok := d.GetOk("access_key"); !ok {
+		d.Set("access_key", config.ObjAccessKey)
+	}
+
+	if _, ok := d.GetOk("secret_key"); !ok {
+		d.Set("secret_key", config.ObjSecretKey)
+	}
+
 	return nil
 }
 
--- a/linode/objbucket/resource.go
+++ b/linode/objbucket/resource.go
@@ -54,6 +54,7 @@ func readResource(
 	populateLogAttributes(ctx, d)
 	tflog.Debug(ctx, "reading linode_object_storage_bucket")
 	client := meta.(*helper.ProviderMeta).Client
+	config := meta.(*helper.ProviderMeta).Config
 
 	cluster, label, err := DecodeBucketID(ctx, d.Id())
 	if err != nil {
@@ -84,8 +85,20 @@ func readResource(
 	}
 
 	// Functionality requiring direct S3 API access
-	accessKey := d.Get("access_key").(string)
-	secretKey := d.Get("secret_key").(string)
+	var accessKey, secretKey string
+
+	if v, ok := d.GetOk("access_key"); ok {
+		accessKey = v.(string)
+	} else {
+		accessKey = config.ObjAccessKey
+	}
+
+	if v, ok := d.GetOk("secret_key"); ok {
+		secretKey = v.(string)
+	} else {
+		secretKey = config.ObjSecretKey
+	}
+
 	endpoint := helper.ComputeS3EndpointFromBucket(ctx, *bucket)
 
 	_, versioningPresent := d.GetOk("versioning")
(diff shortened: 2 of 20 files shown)


linode/objbucket/schema_resource.go
```diff
@@ -4,14 +4,18 @@ import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
...
 	},
 	"access_key": {
-		Type:        schema.TypeString,
-		Description: "The S3 access key to use for this resource. (Required for lifecycle_rule and versioning)",
-		Optional:    true,
+		Type: schema.TypeString,
+		Description: "The S3 access key to use for this resource. (Required for lifecycle_rule and versioning). " +
+			"If not specified with the resource, the value will be read from provider-level obj_access_key.",
+		Optional: true,
+		Computed: true,
```
lgarber-akamai, line 18: I wonder if it'd make sense to hash these in state like we do for root_pass: https://github.com/linode/terraform-provider-linode/blob/dev/linode/instance/schema_resource.go#L340

yec-akamai: I changed a bit around with the hashing and find it will add a lot complexity to the obj keys management. The use case for root_pass is a bit simpler, because it's only used while creating and its value in the state is never reused/read. Also it could be more difficult when we're going to migrate these to framework, since the config value doesn't match the state. I've read some issue with StateFunc migration in the Hashicorp discussion.

Instead of hashing these value, how about marking the secret_key as `Sensitive`? It's also what we do in the `objkey` schema: https://github.com/linode/terraform-provider-linode/blob/dev/linode/objkey/framework_resource_schema.go#L32

lgarber-akamai: Marking them as sensitive sounds perfect 👍

lgarber-akamai (approved):
Great work! E2E tests are all passing locally and the following configuration is working well:

```terraform
provider "linode" {
  obj_access_key = "..."
  obj_secret_key ="..."
}

  cluster = "us-mia-1"
  label = "lgarber-test"

  lifecycle_rule {
    prefix = "tf"
    enabled = true

    abort_incomplete_multipart_upload_days = 5

    expiration {
      date = "2024-03-21"
    }
  }
}

resource "linode_object_storage_object" "test" {
  bucket     = linode_object_storage_bucket.test.label
  cluster    = "us-mia-1"
  key        = "foobar"
  content = "cool"
}
```

(issue #1337 :trollface:)

linode/obj/resource.go
```diff
@@ -91,6 +91,16 @@ func readResource(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
...
+	// Compute object access credentials from provider metadata when they're not configured explicitly
+	config := meta.(*helper.ProviderMeta).Config
+	if _, ok := d.GetOk("access_key"); !ok {
+		d.Set("access_key", config.ObjAccessKey)
+	}
+
+	if _, ok := d.GetOk("secret_key"); !ok {
+		d.Set("secret_key", config.ObjSecretKey)
+	}
+
```
zliang-akamai, line 103: Can we just use the keys in S3 API calls without saving it? When a user config a provider level key, if the key is saved into the state of this resource, later when the user revoke the key and regenerate a new key, then they won't be able set it to provider again because this resource will continue used the saved old key.

Another concern is that by saving the provider level key into the state, the user won't be able to easily remove the key. If they just remove the key from provider without inspecting the state, they might left sensitive key in the state but think they already cleaned up everything sensitive.

linode/objbucket/resource.go
```diff
@@ -84,8 +85,20 @@ func readResource(
...
+	} else {
+		accessKey = config.ObjAccessKey
+	}
+
+	if v, ok := d.GetOk("secret_key"); ok {
+		secretKey = v.(string)
+	} else {
+		secretKey = config.ObjSecretKey
+	}
+
```
zliang-akamai, line 101: The concerns may also be applied to here.

linode/obj/resource.go
yec-akamai: It's a good point. I was hesitating whether to save the keys into the resource state. I eventually saved it because the `getObject()` in the resource_test needs to get the access and secret keys from the state, and it seems no way from there to the provider config. But now I think I probably can generate temporary keys to get around it.

zliang-akamai (approved):
Everything works as expected. Nice work!

PR #1344, opened 2024-03-06
fix: Check for errors before setting log fields in `linode_object_storage_key`

lgarber-akamai:
## 📝 Description

This change resolves an issue that would result in a panic when attempting to provision a `linode_object_storage_key` resource with invalid options. This is because a `helper.SetLogFieldBulk(...)` call was accidentally inserted before the OBJ key creation error check.

Resolves #1334 

## ✔️ How to Test

The following test steps assume you have pulled down this PR locally.

### Integration Testing

```
make PKG_NAME=linode/objkey int-test
```

### Manual Testing

1. In a Terraform provider sandbox environment (e.g. dx-devenv), apply the following configuration:

```terraform
# ...

resource "linode_object_storage_key" "test" {
  label = "extremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabelextremelylonglabel"
}
```

2. Ensure the provider does not panic and instead a human-readable error is raised.

PR #1346, opened 2024-03-10
fix: Check for errors before setting log fields in `linode_nodebalancer_node`, `linode_nodebalancer_config`

junho100:
## 📝 Description

Like PR #1344,  same potential issues exist in `linode_nodebalancer_node`, `linode_nodebalancer_config`. `tflog.SetField(...)` call was inserted before error check.

## ✔️ How to Test

```
make PKG_NAME=linode/nbconfig int-test
make PKG_NAME=linode/nbnode int-test
```


lgarber-akamai (approved):
LGTM, thank you for the contribution!

PR #1347, opened 2024-03-11
fix: Reboot the linode when the active config is updated

jriddle-linode:
## 📝 Description

Updates the logic to reboot the Linode when the active Instance Config is updated. Will only reboot if:
- The config is already booted
- The user has not set SkipImplicitReboots in their config
- The Booted flag for this config is still True

## ✔️ How to Test

```bash
```

## Manual Testing

```terraform
# ...

    label = "rebooter_instance"
    type = "g6-nanode-1"
    region = "us-mia"
}

resource "linode_instance_disk" "foobar" {
  label = "boot"
  size = linode_instance.foobar.specs.0.disk

  image = "linode/alpine3.18"
}

resource "linode_instance_config" "foobar" {
  label = "my-config-updated"
  comments = "cool-updated"

  devices {
    sdb {
      disk_id = linode_instance_disk.foobar.id
    }
  }

  helpers {
    devtmpfs_automount = false
    distro = false
    modules_dep = false
    network = false
    updatedb_disabled = false
  }

  interface {
    purpose = "public"
  }

  kernel = "linode/latest-32bit"
  memory_limit = 513
  root_device = "/dev/sdb"
  virt_mode = "fullvirt"

  booted = true
}
```

```bash
make apply
```

Then change something about the config like the Interface

```terraform
  interface {
    purpose = "vlan"
    label = "cooler"
    ipam_address = "10.0.0.3/24"
  }
```

Then apply again and see that the instance Reboots.


linode/instanceconfig/resource.go
```diff
@@ -270,26 +285,14 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
 
 	// We should not use `HasChange(...)` here because of possible mid-apply changes
 	managedBoot := !d.GetRawConfig().GetAttr("booted").IsNull()
 
-	inst, err := client.GetInstance(ctx, linodeID)
-	if err != nil {
-		return diag.Errorf("Error finding the specified Linode Instance: %s", err)
-	}
-
-	shouldPowerBackOn := !managedBoot && powerOffRequired && inst.Status == linodego.InstanceRunning
```
zliang-akamai, line 292: I think we should only power back on the instances with current status of `running` before we powering it off for VPC update?

zliang-akamai: Oh, never mind, it was already included when `powerOffRequired` is created.

zliang-akamai (approved):
Tried multiple use cases in manual testing, and all worked well! Everything looks great other than Ye's suggestions above. Nice work!

ykim-1 (approved):
LGTM, works locally

PR #1351, opened 2024-03-12
new: Generate temporary object storage keys to use

yec-akamai:
## 📝 Description

This change introduces an implicit object key generation in object storage bucket and object. If opt-in the temporary key generation, user don't have to provide their own obj access keys, and the keys will be created at apply-time with limited access to target bucket and cluster. The keys will be cleaned up automatically after use. User can only opt-in the temp key generation at provider-level.

This PR also combines #1337 which enables provider-level obj keys. With both change, now when object storage keys are required in the resources, we will look for them in the following order:
1) Whether the keys are specified in the resource configuration;
2) Whether the provider-level object keys exist;
3) Whether user opts-in temporary keys generation.

If none of these is satisfied, an error will be raised.

Addressed issue #1172.

## ✔️ How to Test

`make int-test PKG_NAME="linode/objbucket"`
`make int-test PKG_NAME="linode/obj"`

**Manual Test:**
1. In a sandbox environment, i.e. dx-devenv, run the following configuration to enable temporary keys generation when creating object storage bucket and an object in it
```terraform
provider "linode" {
    obj_use_temp_keys = true
}

    cluster = "us-mia-1"
    label = "test-obj-bucket"

    lifecycle_rule {
        prefix = "tf"
        enabled = true

        abort_incomplete_multipart_upload_days = 5

        expiration {
            date = "2024-03-21"
        }
    }
}

resource "linode_object_storage_object" "test" {
    bucket     = linode_object_storage_bucket.test.label
    cluster    = "us-mia-1"
    key        = "foobar"
    content = "cool"
}
```
2. Plan and apply the change. Observe the resources are created successfully.


zliang-akamai:
Do you think it would be nicer to return the temp keys from the utility function rather than setting it into the schema? I feel it is not very safe to set it into the schema and then cleanup later. Imagine that somehow a bug or something causes the cleaning up failed..

yec-akamai:
> Do you think it would be nicer to return the temp keys from the utility function rather than setting it into the schema? I feel it is not very safe to set it into the schema and then cleanup later. Imagine that somehow a bug or something causes the cleaning up failed..

I did try to implement that way, but it turns out kind of redundant that we only return the temp keys value. Because there are 3 different ways to grab the keys now, making use of the schema can help us unify the code. I also think about if we migrate the resource to framework in the future, we will be able to make use of the resource model instead, which is clearer and safer.

Regarding to clean up, I don't think it's very likely to have an issue there unless it's a real bug. We won't write the change into state and it doesn't change anything with user config.

linode/obj/resource.go
```diff
@@ -104,12 +134,42 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+
+				defer CleanUpTempKeys(ctx, client, keys.ID)
+			}
+		}
+
+		if !CheckObjKeysConfiged(objKeys) {
+			return diag.Errorf("access_key and secret_key are required to update linode_object_storage_object")
+		}
+
+		s3client, err := helper.S3ConnectionFromData(ctx, d, meta, objKeys.AccessKey, objKeys.SecretKey)
```
jriddle-linode, line 172: Since this section of checks appears a few times perhaps it'd be better to put in a helper function like `S3ConnectionFromData`?

jriddle-linode (approved):
Other than the suggestion everything looks good and passes locally.

linode/obj/resource.go
zliang-akamai: Or we might put a helper function to get keys in the `obj/helper`
Here is a sample code I wrote. I didn't test it though.
```go
func GetObjKeys(
	ctx context.Context,
	d *schema.ResourceData,
	config *helper.Config,
	client linodego.Client,
	bucket, cluster string,
) (ObjectKeys, diag.Diagnostics, func()) {
	objKeys := ObjectKeys{
		AccessKey: d.Get("access_key").(string),
		SecretKey: d.Get("secret_key").(string),
	}

	var teardownObjTempKeys func() = nil

	if !CheckObjKeysConfiged(objKeys) {
		// If object keys don't exist in the resource configuration, firstly look for the keys from provider configuration
		if providerKeys, ok := GetObjKeysFromProvider(objKeys, config); ok {
			objKeys = providerKeys
		} else if config.ObjUseTempKeys {
			// Implicitly create temporary object storage keys
			keys, diag := CreateTempKeys(ctx, client, bucket, cluster, "read_write")
			if diag != nil {
				return objKeys, diag, nil
			}

			objKeys.AccessKey = keys.AccessKey
			objKeys.SecretKey = keys.SecretKey
			teardownObjTempKeys = func() {CleanUpTempKeys(ctx, client, keys.ID)}
		}
	}

	if !CheckObjKeysConfiged(objKeys) {
		return objKeys, diag.Errorf(
			"access_key and secret_key are required to update linode_object_storage_object",
		), nil
	}
	return objKeys, nil, teardownObjTempKeys
}
```

yec-akamai: Thank you both for the suggestion! Just made this into one helper function.

zliang-akamai (approved):
There are some intermittent errors, very likely caused by Terraform testing framework. We should be fine to proceed to merging this PR.

Everything else looks good and works well. Multiple manual tests worked perfectly.

```
=== NAME  TestAccResourceObject_tempKeys
    test_retry.go:25: Retrying on test failure: Step 1/1 error: Error running apply: exit status 1
        
        Error: access_key and secret_key are required.
        
          with linode_object_storage_object.temp_keys,
          on terraform_plugin_test.tf line 26, in resource "linode_object_storage_object" "temp_keys":
          26: resource "linode_object_storage_object" "temp_keys" {
```

lgarber-akamai (approved):
Looks great! After looking into the weird validation error in the E2E tests I think it's safe to say it's a weird TF testing framework related to overriding provider configurations, so we can probably just ignore it for now 👍

lgarber-akamai (commented):
Just a heads up, I'm occasionally getting this error when attempting to provision a bunch of objects at once: 
```
failed to put Bucket (fewijgioerjgioejrng) Object (test_temp_keys): operation error S3: PutObject, https response error StatusCode: 403, RequestID: tx0000090627edba3655d97-0065fdcf32-1bad7e04-default, HostID: 1bad7e04-default-default, api error AccessDenied: UnknownError
```

I'm guessing it's intermittent so it might make sense to add some retry logic [here](https://github.com/linode/terraform-provider-linode/blob/TPT-2583/opt-in-temp-obj-keys/linode/obj/resource.go#L271) to mitigate this issue.

linode/obj/helpers.go
```diff
@@ -2,15 +2,134 @@ package obj
...
+			objKeys.SecretKey = keys.SecretKey
+			teardownTempKeysCleanUp = func() { cleanUpTempKeys(ctx, client, keys.ID) }
+		}
+	}
+
+	if !checkObjKeysConfiged(objKeys) {
+		return objKeys, diag.Errorf(
+			"access_key and secret_key are required.",
+		), nil
+	}
```
lgarber-akamai, line 132: nitpick:
```suggestion

	if !checkObjKeysConfiged(objKeys) {
		// If object keys don't exist in the resource configuration, firstly look for the keys from provider configuration
		if providerKeys, ok := getObjKeysFromProvider(objKeys, config); ok {
			objKeys = providerKeys
		} else if config.ObjUseTempKeys {
			// Implicitly create temporary object storage keys
			keys, diag := createTempKeys(ctx, client, bucket, cluster, permission)
			if diag != nil {
				return objKeys, diag, nil
			}

			objKeys.AccessKey = keys.AccessKey
			objKeys.SecretKey = keys.SecretKey
			teardownTempKeysCleanUp = func() { cleanUpTempKeys(ctx, client, keys.ID) }
		} else {
			return objKeys, diag.Errorf(
				"access_key and secret_key are required.",
			), nil
		}
```

lgarber-akamai, line 132: nitpick:
```suggestion

	if !checkObjKeysConfiged(objKeys) {
		// If object keys don't exist in the resource configuration, firstly look for the keys from provider configuration
		if providerKeys, ok := getObjKeysFromProvider(objKeys, config); ok {
			objKeys = providerKeys
		} else if config.ObjUseTempKeys {
			// Implicitly create temporary object storage keys
			keys, diag := createTempKeys(ctx, client, bucket, cluster, permission)
			if diag != nil {
				return objKeys, diag, nil
			}

			objKeys.AccessKey = keys.AccessKey
			objKeys.SecretKey = keys.SecretKey
			teardownTempKeysCleanUp = func() { cleanUpTempKeys(ctx, client, keys.ID) }
		} else {
			return objKeys, diag.Errorf(
				"access_key and secret_key are required.",
			), nil
		}
```

yec-akamai:
@zliang-akamai @lgarber-akamai Thank you both for the careful review!

> Just a heads up, I'm occasionally getting this error when attempting to provision a bunch of objects at once:
> 
> ```
> failed to put Bucket (fewijgioerjgioejrng) Object (test_temp_keys): operation error S3: PutObject, https response error StatusCode: 403, RequestID: tx0000090627edba3655d97-0065fdcf32-1bad7e04-default, HostID: 1bad7e04-default-default, api error AccessDenied: UnknownError
> ```
> 
> I'm guessing it's intermittent so it might make sense to add some retry logic [here](https://github.com/linode/terraform-provider-linode/blob/TPT-2583/opt-in-temp-obj-keys/linode/obj/resource.go#L271) to mitigate this issue.

Good catch! I added a new helper `putObjectWithRetries` to handle the retries. Let me know if it looks good to you!

lgarber-akamai (approved):
Looks perfect, nice work!

PR #1361, opened 2024-03-20
fix: Update context when populating log attributes with correct types

yec-akamai:
## 📝 Description

Some of our code is missing updating the context when populating log attributes. Update `ctx` variable for these.

Also fix the type issue in `populateLogAttribute` in several resources.

## ✔️ How to Test

1. Enable trace logging for the Linode provider:
```
export TF_LOG=TRACE
export TF_LOG_PROVIDER_LINODE=TRACE
```

2. In a sandbox environment, i.e. dx-devenv, run the following configuration to create a objbucket:
```
    cluster = "us-mia-1"
    label = "test-obj-bucket"
}
```

3. Make sure that log attributes (`bucket`, `cluster`) are populated correctly. Destroy the resource for clean up.

4. Run the configuration to create a ipv6range:
```
  label = "my-linode"
  image = "linode/alpine3.19"
  type = "g6-nanode-1"
  region = "us-southeast"
}

  prefix_length = 64
}
```
5. Destroy the resource, and make sure that log attributes (`ipv6_id`, `range`) for deleting it are populated correctly.

### DIFF
--- a/linode/instancedisk/framework_resource.go
+++ b/linode/instancedisk/framework_resource.go
@@ -471,7 +471,7 @@ func (r *Resource) ImportState(
 
 func populateLogAttributes(ctx context.Context, model ResourceModel) context.Context {
 	return helper.SetLogFieldBulk(ctx, map[string]any{
-		"linode_id": model.LinodeID,
-		"disk_id":   model.ID,
+		"linode_id": model.LinodeID.ValueInt64(),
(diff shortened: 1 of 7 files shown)


zliang-akamai (approved):
Thanks for fixing it!

lgarber-akamai (approved):
LGTM, thanks for addressing this!

PR #1366, opened 2024-03-22
test: migrate to feature-specific tagging for tests instead of opt in system

ykim-1:
## 📝 Description

Migrating from opt in system to using feature specific go:build tags for integration test. 

make command updates:

- `make int-test` : will run all integration tests

there is a new variable `TEST_TAGS` where we can specify a module name and run it separately. If no `TEST_TAGS` is specified it will default to `integration` tag

e.g. to run from `domainrecord`
- `make TEST_TAGS="domainrecord" int-test`

Note: if both tags are specified it will run every integration test since the logic is that each package name is `OR` with `integration` tag
 
All Database tests have `dbaas` tag now, but the long running tests (e.g. `databasemysql`) will still require `RUN_LONG_TESTS` flag to be True

e.g. 
`make TEST_TAGS="dbaas_tests" RUN_LONG_TESTS=True int-test` will run every database tests

`make TEST_TAGS="dbaas_tests" int-test` will exclude the long running database tests

Similarly, for running all account related tests:

`make TEST_TAGS="act_tests" ACC_OPT_IN_TESTS="test" int-test` will run account related tests

## ✔️ How to Test

`make TEST_TAGS="domainrecord" int-test`
`make int-test`

## 📷 Preview


ykim-1:
Tested on my forked repo, failures are not due to this PR 
http://198.19.5.79:1212/builds/65fda0e2b1b33f00014d2750?team=DX&buildName=linode-terraform&bld_id=65fda0e2b1b33f00014d2750&testcaseStatus=failure

linode/accountavailabilities/datasource_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build (integration || accountavailabilities) && !optional && !long_running
```
lgarber-akamai, line 1: Just to make sure I'm understanding this fully, is the intention here to disable this test if long-running tests or optional tests are enabled?

ykim-1: Yeah exactly, and I wanted to give an option to run each package separately if we needed to. 
E.g. 
- `accountavailabilities` will run just tests in the package
- `integration` will run all tests excluding tests marked with long_running and optional
- specifying either optional or long_running will run exclude everything except that is marked those tags

yec-akamai (commented):
Nice work separating the tests! Can we add the new test instructions from description in README?

jriddle-linode (approved):
Works locally, too bad it adds all these `[no test files]` lines, if we could remove those that would be nice.. unless they're being used for TOD or something.

ykim-1:
> Nice work separating the tests! Can we add the new test instructions from description in README?

Yup will do! changing the README file slipped through my mind

.github/workflows/integration_tests_pr.yml
```diff
@@ -49,15 +46,56 @@ jobs:
...
+          --xmlfile "${filename}"
+
+      - name: Upload test results to bucket
+        run: |
+          report_filename=$(ls | grep -E '^[0-9]{12}_terraform_test_report\.xml$')
+          python3 tod_scripts/test_report_upload_script.py "${report_filename}"
+        env:
+          LINODE_CLI_OBJ_ACCESS_KEY: ${{ secrets.LINODE_CLI_OBJ_ACCESS_KEY }}
+          LINODE_CLI_OBJ_SECRET_KEY: ${{ secrets.LINODE_CLI_OBJ_SECRET_KEY }}
+
```
zliang-akamai, line 98: Maybe we can leave this workflow simpler by not uploading to TOD and just showing the result on GitHub?

ykim-1: It already reports the report in GitHub in the `Run integration Tests` step
`          cat "$report_filename"`

ykim-1:
The change is a little bit messy and unclear when running the test, will try to make it cleaner. Putting the PR back to draft

linode/accountavailabilities/datasource_test.go
ykim-1: I changed the logic a bit for better ease of use so scratch everything I said above. Updated the description

lgarber-akamai:
@ykim-1 Just a heads up, it looks like a bunch of commits from other PRs somehow ended up in the diff for this PR 👀

linode/databaseaccesscontrols/resource_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration || databaseaccesscontrols
```
lgarber-akamai, line 1: Would it make sense to add a `dbaas` build flag that is required to run all of the DBaaS tests?

linode/lkenodepool/framework_resource_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build (integration || lkenodepool) && !optional && !long_running
```
yec-akamai, line 1: Do we still want to have the `long_running` tag? It seems not being addressed/described after the clean up.

linode/objcluster/datasource_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build (integration || objcluster) && !optional && !long_running
```
yec-akamai, line 1: Same here^

yec-akamai (commented):
Overall looks good! It's easier to understand how to run the integration tests after the clean up 👍

linode/accountsettings/resource_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration || accountsettings
```
lgarber-akamai, line 1: I forgot to include this in my previous feedback but could we put all account tests behind an additional flag like we did for DBaaS? Account tests can be a bit finicky and may not always work as expected depending on the target account 🙂

linode/objcluster/datasource_test.go
ykim-1: Thanks for catching these two.. I must have missed them 👍

linode/accountsettings/resource_test.go
ykim-1: Yeah we can but I think we should still keep `OptInTest()` and `LongRunningTest()` functions since it will make the solution a lot simpler to use with `make` commands 
e.g.
```
make int-test    // runs all integration tests excluding long running (dbaas) and optin (account) tests
make TEST_TAGS="domain"  int-test // runs domain tests  
make TEST_TAGS="dbaas_tests" RUN_LONG_TESTS=True  int-test    // runs all dbaas tests
make TEST_TAGS="act_tests" ACC_OPT_IN_TESTS=True  int-test  // runs all account related tests
make RUN_LONG_TESTS=True ACC_OPT_IN_TESTS=True  int-test  // runs everything
```

yec-akamai (approved):
Nice work!

jriddle-linode (approved):
Works great, thanks for cleaning up the output.

lgarber-akamai (approved):
LGTM, excellent work!

PR #1367, opened 2024-03-25
new: Add support for batch API requests

lgarber-akamai:
## 📝 Description

This change adds support for concisely running non-sequential API requests concurrently. This can considerably speed up plans for configurations that contain a large number of resources and for users with high-latency connections.

This pull request only parallelizes GET requests in instance-related resources, we can prioritize moving other resources over in the future.

The performance gains may seem small (~1 second) on a minimal configuration or on a low-latency connection, but the performance improvement should scale considerably with less ideal conditions.

## ✔️ How to Test

### Integration Testing

```
```

### Unit Testing

```
make unit-test
```

### Manual Testing

```terraform
  label = "my-config"

  device {
    device_name = "sda"
  }

  booted = true
}

  label = "boot"
  size = linode_instance.my-instance.specs.0.disk

  image = "linode/ubuntu22.04"
  root_pass = "myc00lpa$$w0rd!!!!!!"
}

  label = "my-instance"
  type = "g6-nanode-1"
  region = "us-mia"
}
```

2. Ensure the configuration applies successfully.


zliang-akamai (approved):
Tests passed. Everything looks good to me!

ykim-1 (approved):
LGTM

PR #1369, opened 2024-03-26
Add Framework List Value Helpers

zliang-akamai:
## 📝 Description

I tried to write some helpers for converting a slice of items to a TF framework `types.List` value. My thought was naive and final implementation (as in this PR) is much more complicated than I imagined originally, but I still think it may be worth to add since it can still make list value creation easier on the data source and resource CURD code.

## ✔️ How to Test
I tried migrating the data source in `linode/instancenetworking` to use these helpers, so you can do this for integration testing:

```bash
make PKG_NAME="linode/instancenetworking" int-test
```

Or manual testing

```terraform
data "linode_instance_networking" "example" {
    linode_id = 123
}
```


lgarber-akamai (approved):
LGTM! Could you drop the `log.txt` and `main.tf` before merging?

zliang-akamai:
@lgarber-akamai oops, accidentally included it in the fix commit.. Removed now.

jriddle-linode (approved):
LGTM

PR #1383, opened 2024-04-05
Add time format upgrade logic for stackscript resource

### LINKED ISSUE #1372: [Bug]: Stackscript RFC3339 error
### Terraform Version

Terraform v1.3.7 on linux_amd64

### Linode Provider Version

v2.18.0

### Effected Terraform Resources

linode_stackscript

### Terraform Config Files

```terraform
resource "linode_stackscript" "test" {
    label = "test"
    description = "Test stackscript"
    script = <<EOF
#!/bin/bash
echo "test"
EOF
    images = ["any/all"]
    rev_note = "v1"
}
```

### Debug Output

Error: Invalid RFC3339 String Value 
│   with linode_stackscript.test,
│   on test.tf line 1, in resource "linode_stackscript" "test":
│    1: resource "linode_stackscript" "test" {
│ 
│ A string value was provided that is not valid RFC3339 string format.
│ 
│ Given Value: 2023-08-12 14:41:04 +0000 UTC
│ Error: parsing time "2023-08-12 14:41:04 +0000 UTC" as "2006-01-02T15:04:05Z07:00": cannot parse " 14:41:04 +0000 UTC" as "T"

### Panic Output

_No response_

### Expected Behavior

Should correctly interpret the timestamps

### Actual Behavior

After upgrading from v1 to v2.18.0, I'm getting this error on plan/apply for an existing state file. There is no date/time included in the resource config, so this appears to be a problem with parsing the created/updated timestamp in the existing state file. Manually updating timestamps in the state file to match RFC3339 appears to resolve the error, but I would have expected this to either be handled automatically or documented somewhere.

For now, the workaround of manually editing the state file appears to be sufficient, so feel free to close this if that's the recommended approach to a v1 -> v2 upgrade.

### Steps to Reproduce

Use existing state file for a stackscript resource created before v2, using a non-RFC3339 timestamp in the "created" and "updated" fields.

Apologies if there are upgrade/migration instructions somewhere for v2 of the provider that we didn't follow, I couldn't find anything in the docs.

zliang-akamai:
## 📝 Description

Implement time format upgrade logic to seamlessly upgrade the time format from something like`2024-04-05 21:42:22 +0000 UTC` to RFC3339, while preventing the error reported by RFC3339 format validation in the `timetypes.RFC3339Type` when the old format has been written into the state by a v1 Linode provider.

closes #1372 

### Automated Testing

```bash
make PKG_NAME="linode/stackscript" unit-test
```

```bash
make PKG_NAME="linode/stackscript" int-test
```

### Manual Testing

- Turn off the [Terraform provider development overrides](https://developer.hashicorp.com/terraform/plugin/debugging#terraform-cli-development-overrides)
- Apply this Terraform config below with a v1 provider
```terraform
terraform {
  required_providers {
    linode = {
      source = "linode/linode"
      version = "1.30.0"
    }
  }
}
provider "linode" {
}

resource "linode_stackscript" "test" {
    label = "test"
    description = "Test stackscript"
    script = <<EOF
#!/bin/bash
echo "test"
EOF
    images = ["any/all"]
    rev_note = "v1"
}
```
- When changing provider version, you may see `Error: Inconsistent dependency lock file`. You can do `terraform init -upgrade` to update the lock file to fix it.
- After applying it with v1 provider, turn the [Terraform provider development overrides](https://developer.hashicorp.com/terraform/plugin/debugging#terraform-cli-development-overrides) back on.
- Do `go build` and then `terraform plan` with the provider in this branch to verify that no error is showing up.


lgarber-akamai:
Just a thought, have you looked into whether this same issue affects resources other then `linode_stackscript`?

zliang-akamai:
Hi @lgarber-akamai, there was a similar issue there that causes RFC3339 custom type couldn't be applied to the firewall device resource.
https://github.com/linode/terraform-provider-linode/blob/3be9fbe1f9e5c71d2ecce368500550e7fea8f48b/linode/firewalldevice/framework_resource_schema.go#L43-L56

But I can also implement a similar state upgrader for it, maybe in another PR?

lgarber-akamai (approved):

ykim-1 (approved):
Tests and manual steps work well locally

PR #1396, opened 2024-04-19
Fix Sent to Closed Channel Panic in `helper.RunBatch`

zliang-akamai:
## 📝 Description

~~This is to fix an issue that errors can be sent to a closed channel in `helper.RunBatch` function.~~

Instead of fixing existing code, the better solution is to migrate to `sync/errgroup`. Thanks @lgarber-akamai for the idea!

Other MISC changes bundled:
- Migrate `log` to `tflog` in `instanceconfig` package.
- Change `ErrHasStatus(err, 404)` to `IsNotFound(err)`
- ~~Joining all errors in via `errors.Join`~~ No longer needed after migrating to `sync/errgroup`

### Automated Testing

```
```

### Manual Testing

```terraform
  label = "my-config"

  device {
    device_name = "sda"
  }

  booted = true
}

  label = "boot"
  size = linode_instance.my-instance.specs.0.disk

  image = "linode/ubuntu22.04"
  root_pass = "myc00lpa$$w0rd!!!!!!"
}

  label = "my-instance"
  type = "g6-nanode-1"
  region = "us-mia"
}
```

2. Remove the Linode instance on Cloud Manager to force a 404 error returned from API to the Terraform Provider.
3. `terrafrom refresh`
4. Observe no panic (i.e. no provider crash error).

### DIFF
--- a/linode/helper/batch.go
+++ b/linode/helper/batch.go
@@ -1,7 +1,7 @@
 package helper
 
 import (
-	"fmt"
+	"errors"
 	"sync"
 )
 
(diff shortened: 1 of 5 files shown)


linode/helper/batch.go
```diff
@@ -11,35 +11,38 @@ type BatchFunction func() error
 // This is handy for running certain non-sequential API requests in parallel.
 func RunBatch(toExecute ...BatchFunction) error {
 	var wg sync.WaitGroup
+	wg.Add(len(toExecute))
```
lgarber-akamai, line 14: @zliang-akamai We have a ticket up to move this over to errgroups (TPT-2829), would it make sense to just do that as a part of this PR?

lgarber-akamai (approved):
All tests are passing on my end and the implementation looks great, nice work!

yec-akamai (approved):
Works well on my end, nice work!

PR #1399, opened 2024-04-22
Fix `kernels` integration test with updated filter

yec-akamai:
## 📝 Description

The kernels test was failing because of outdated filter. Updated the filter content and removed some filters to reduce the chance of failures due to updated information from API response.

## ✔️ How to Test

`make int-test PKG_NAME=linode/kernels`


lgarber-akamai (approved):
LGTM, thanks for fixing this!

ykim-1 (approved):
thanks for fixing this

zliang-akamai (approved):

PR #1401, opened 2024-04-24
new: Add `linode_placement_group` resource

lgarber-akamai:
## 📝 Description

This pull request adds a new `linode_placement_group` resource that allows users to manage Linode Placement Groups through Terraform.

**NOTE: This pull request does NOT include the logic to manage Placement Group assignment and unassignment. That functionality will be added in TPT-2626 and TPT-2859 as discussed in private.**

## ✔️ How to Test

The following test steps require that you have pulled down this PR locally and have pointed your environment variables at an API instance where placement groups are accessible:

```
export LINODE_URL=...
export LINODE_TOKEN=...
```

### Unit Testing

```
make PKG_NAME=linode/placementgroup unit-test
```

### Integration Testing

```
make PKG_NAME=linode/placementgroup int-test
```

### Manual Testing

```terraform
resource "linode_placement_group" "group" {
  label = "test"
  region = "us-east"
  affinity_type = "anti_affinity:local"
}
```

3. Re-plan the configuration and ensure no changes are planned.
4. Navigate to the new PG in Cloud Manager and ensure all fields match the configuration.
5. Apply the following configuration to update the PG's label:

```terraform
resource "linode_placement_group" "group" {
  label = "test-updated"
  region = "us-east"
  affinity_type = "anti_affinity:local"
}
```

6. Refresh the PG in Cloud Manager and ensure the label has been updated
7. Run `terraform destroy` on the configuration and ensure it completes successfully.
8. Ensure the PG is no longer present in Cloud Manager.


docs/resources/placement_group.md
```diff
@@ -0,0 +1,51 @@
...
+
+The following arguments are supported:
+
+* `label` - (Required) The label of the Placement Group. This field can only contain ASCII letters, digits and dashes.
+
+* `region` - (Required) The region of the Placement Group.
+
+* `affinity_type` - (Required) The affinity policy to use when placing Linodes in this group.
+
+* `is_strict` - (Optional) Whether Linodes must be able to become compliant during assignment. (Default `true`)
```
lgarber-akamai, line 33: The default in Cloud Manager is true but the field is required by the API. Would it make sense to make the field required in TF too?

linode/placementgroup/framework_models.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type PlacementGroupMemberModel struct {
+	LinodeID    types.Int64 `tfsdk:"linode_id"`
+	IsCompliant types.Bool  `tfsdk:"is_compliant"`
+}
+
+func (m *PlacementGroupMemberModel) FlattenPlacementGroupMember(member *linodego.PlacementGroupMember) {
+	// This object is always computed so we don't need to respect preserveKnown here.
+	// It may still be good practice to respect preserveKnown here though, so definitely
```
zliang-akamai, line 18: While I think it's completely fine to not respect `preserveKnown` for a fully computed attribute, I recently realize that when we have to determine whether it's an unknown, we can do conversions between a slice of model structs and a Framework's `types.List`:
```go
list := types.ListValueFrom(ctx, someObjectType, sliceOfModels)

list.ElementsAs(ctx, &sliceOfModels, false)
```

https://github.com/linode/terraform-provider-linode/pull/1405/files#diff-b51120d512a9186537c26eccf08af488e064b5ddb2ee15e98f01627c6317739d

linode/placementgroup/framework_resource.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	id := helper.FrameworkSafeStringToInt(data.ID.ValueString(), &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	pg, err := client.GetPlacementGroup(ctx, id)
+	if err != nil {
+		if lerr, ok := err.(*linodego.Error); ok && lerr.Code == 404 {
```
zliang-akamai, line 103: I think we may use the `IsNotFound` helper here.

linode/placementgroup/framework_resource.go
```diff
@@ -0,0 +1,213 @@
...
+
+	id := helper.FrameworkSafeStringToInt(data.ID.ValueString(), &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	tflog.Debug(ctx, "client.DeletePlacementGroup(...)")
+	err := client.DeletePlacementGroup(ctx, id)
+	if err != nil {
+		if lerr, ok := err.(*linodego.Error); (ok && lerr.Code != 404) || !ok {
```
zliang-akamai, line 201: Same here

linode/placementgroup/framework_schema_resource.go
```diff
@@ -0,0 +1,72 @@
...
+					"is_compliant": schema.BoolAttribute{
+						Computed: true,
+						Description: "Whether this member is currently compliant with the " +
+							"Placement Group's affinity policy.",
+					},
+				},
+			},
+		},
+	},
+}
```
zliang-akamai, line 72: I think we can't use block for computed list or set. A `SetAttribute` would be needed.

Issue can be reproduced by:
- Create a placement group with Terraform
- Assign it a Linode on Cloud Manager
- `terraform refresh`
- `terraform apply` will error out.

zliang-akamai: You can still use model slice with a SetAttribute in schema though.

lgarber-akamai: Ah nice find! In this case I guess we'll need to use a SetAttribute with a custom ObjectType like we do in [`linode_stackscript`](https://github.com/linode/terraform-provider-linode/blob/dev/linode/stackscript/framework_resource_schema.go#L141)?

linode/placementgroup/framework_models.go
lgarber-akamai: Good to know!

zliang-akamai (approved):
Everything works well, nice work!

ykim-1 (approved):
LGTM! Tests pass locally

PR #1402, opened 2024-04-25
Include resource and data source name as a comment in user-agent

zliang-akamai:
## 📝 Description

This is to include resource or data source name in the user agent of the API calls.

### Automated Testing

Which runs very looooong and may have intermittent errors.
```bash
make test
```

### Manual Testing

- Enable logging:
```bash
export TF_LOG=DEBUG
export TF_LOG_PATH=$PWD/log.txt 
export TF_LOG_PROVIDER_LINODE_REQUESTS=DEBUG
```
- Deploy some resources, like:
```terraform
resource "linode_vpc" "test" {
    label = "test-vpc${count.index}"
    region = "us-mia"
    description = "My first VPC."
    count = 3
}

    label = "mynodebalancer${count.index}"
    region = "us-mia"
    client_conn_throttle = 20
    count = 3
}
```
- Inspect `log.txt` for user agents recorded, they should be something like `User-Agent="HashiCorp Terraform/1.8.1 (+https://www.terraform.io) Terraform-Plugin-Framework/1.8.0 terraform-provider-linode/dev (linode_vpc resource)"`

### DIFF
--- a/linode/framework_provider_config.go
+++ b/linode/framework_provider_config.go
@@ -290,7 +290,7 @@ func (fp *FrameworkProvider) InitProvider(
 		client.SetRetryMaxWaitTime(time.Duration(maxRetryDelayMilliseconds) * time.Millisecond)
 	}
 
-	userAgent := fp.fwProviderUserAgent(tfVersion, UAPrefix)
+	userAgent := fp.userAgent(tfVersion, UAPrefix)
 	client.SetUserAgent(userAgent)
 	fp.Meta.ProviderUserAgent = userAgent
 
@@ -300,7 +300,7 @@ func (fp *FrameworkProvider) InitProvider(
 	fp.Meta.Client = client
 }
 
-func (fp *FrameworkProvider) fwProviderUserAgent(
+func (fp *FrameworkProvider) userAgent(
 	tfVersion string,
 	UAPrefix string,
 ) string {
--- a/linode/framework_provider_config.go
+++ b/linode/framework_provider_config.go
@@ -27,7 +27,10 @@ func (fp *FrameworkProvider) Configure(
 	resp *provider.ConfigureResponse,
 ) {
 	var data helper.FrameworkProviderModel
-	var meta helper.FrameworkProviderMeta
+
+	if fp.Meta == nil {
+		fp.Meta = &helper.FrameworkProviderMeta{}
+	}
 
 	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
 	if resp.Diagnostics.HasError() {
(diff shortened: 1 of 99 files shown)


linode/helper/framework_resource_base.go
```diff
@@ -55,6 +59,8 @@ func (r *BaseResource) Configure(
 		return
 	}
 
+	r.Client = GetFwClientWithUserAgent(r.Config.Name+" resource", r.Meta)
```
lgarber-akamai, line 62: Optional nitpick: Would it make sense to distinguish resources/data sources in the UA using the `data.` prefix?

e.g.
`linode_image` vs `data.linode_image`?

linode/helper/provider_meta.go
```diff
@@ -0,0 +1,35 @@
...
+	meta *FrameworkProviderMeta,
+) *linodego.Client {
+	client := meta.Client
+	client.SetUserAgent(generate(meta.ProviderUserAgent, resourceOrDataSourceUAComment))
+
+	return &client
+}
+
+func GetSDKClientWithUserAgent(
+	resourceOrDataSourceUAComment string,
```
lgarber-akamai, line 20: Just to simplify and unify these helpers a bit, would it make sense for them to accept the resource name and whether it's a resource or data source (enum?) as separate arguments?

linode/framework_provider_config.go
```diff
@@ -290,16 +290,17 @@ func (fp *FrameworkProvider) InitProvider(
...
 	helper.ApplyAllRetryConditions(&client)
 
-	meta.Config = lpm
-	meta.Client = &client
+	fp.Meta.Config = lpm
+	fp.Meta.Client = client
 }
 
-func (fp *FrameworkProvider) terraformUserAgent(
+func (fp *FrameworkProvider) fwProviderUserAgent(
```
lgarber-akamai, line 303: nitpick: I think this can just become `userAgent(...)` since fwProvider is implied from the struct?

jriddle-linode (approved):
LGTM, works locally, approved pending Lena's recommendations.

linode/helper/provider_meta.go
zliang-akamai: They takes different struct as parameters (our Framework provider metadata and SDKv2 provider metadata are separate structs), so I think it wouldn't look great if we combine them. What do you think?

linode/helper/framework_resource_base.go
zliang-akamai: Thanks for the suggestion! Just added it.

linode/helper/provider_meta.go
lgarber-akamai: Good point, I think it makes sense to leave it as-is in that case 👍

lgarber-akamai (approved):
LGTM, great work!

PR #1403, opened 2024-04-26
fix: Revert RunBatch(...) usage

lgarber-akamai:
## 📝 Description

This change reverts all usage of `helper.RunBatch(...)` in response to a thread-safety issue in the underlying HTTP client used by linodego. We have plans to resolve this issue in the future but have decided to drop concurrency from API requests in TF for the time being.

Test runs:
* https://github.com/linode/terraform-provider-linode/actions/runs/8850661690
* https://github.com/linode/terraform-provider-linode/actions/runs/8850663250

## ✔️ How to Test

### Integration Testing

```
make PKG_NAME=linode/instance int-test
make PKG_NAME=linode/instanceconfig int-test
```

### DIFF
--- a/linode/instance/resource.go
+++ b/linode/instance/resource.go
@@ -53,50 +53,30 @@ func readResource(ctx context.Context, d *schema.ResourceData, meta interface{})
 		return diag.Errorf("Error parsing Linode instance ID %s as int: %s", d.Id(), err)
 	}
 
-	var instance *linodego.Instance
-	var instanceNetwork *linodego.InstanceIPAddressResponse
-	var instanceDisks []linodego.InstanceDisk
-	var instanceConfigs []linodego.InstanceConfig
+	instance, err := client.GetInstance(ctx, id)
+	if linodego.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


linode/instance/resource.go
```diff
@@ -53,50 +53,30 @@ func readResource(ctx context.Context, d *schema.ResourceData, meta interface{})
...
-		func(ctx context.Context) (err error) {
-			instanceConfigs, err = client.ListInstanceConfigs(ctx, id, nil)
-			if err != nil {
-				err = fmt.Errorf("failed to get instance configs: %w", err)
-			}
-			return
-		},
-	)
+	instance, err := client.GetInstance(ctx, id)
+	if linodego.IsNotFound(err) {
```
lgarber-akamai, line 57: linodego.IsNotFound(...) has handling for nil errors so I figured I'd leave it out of the `if err != nil` block for readability 👍

ykim-1 (approved):
Thanks for addressing this, hopefully we get more consistent test runs

jriddle-linode (approved):
LGTM works locally.

PR #1404, opened 2024-04-30
fix: Apply Crossplane plan ID workaround across all Framework resources

lgarber-akamai:
## 📝 Description

This change resolves an issue that caused resources managed through Crossplane to have their IDs dropped during update operations. In short, this issue occurred because the plan's ID is populated with an empty string (`UseStateForUnknown` isn't respected), so `preserveKnown` will not implicitly update the ID when applying the state to the plan.

## ✔️ How to Test

N/A

### DIFF
--- a/linode/accountsettings/framework_resource.go
+++ b/linode/accountsettings/framework_resource.go
@@ -128,6 +128,13 @@ func (r *Resource) Update(
 
 	plan.CopyFrom(state, true)
 
+	// Workaround for Crossplane issue where ID is not
+	// properly populated in plan
+	// See TPT-2865 for more details
+	if plan.ID.ValueString() == "" {
+		plan.ID = state.ID
+	}
+
 	// Apply the state changes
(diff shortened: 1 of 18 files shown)


tchinmai7 (approved):
Works great! Seeing exactly one create and update call, and no duplicate resources.

```
Events:
  Type     Reason                           Age                   From                                                  Message
  ----     ------                           ----                  ----                                                  -------
  Warning  CannotResolveResourceReferences  8m6s (x5 over 8m21s)  managed/instance.linode.upbound.io/v1alpha1, kind=ip  cannot resolve references: mg.Spec.ForProvider.LinodeID: referenced field was empty (referenced resource may not yet be ready)
  Normal   CreatedExternalResource          7m50s                 managed/instance.linode.upbound.io/v1alpha1, kind=ip  Successfully requested creation of external resource
  Normal   UpdatedExternalResource          7m49s                 managed/instance.linode.upbound.io/v1alpha1, kind=ip  Successfully requested update of external resource
```

zliang-akamai (approved):

PR #1406, opened 2024-04-30
new: Add Placement Group fields to `linode_instance` resource

lgarber-akamai:
## 📝 Description

This pull request adds support for assigning a Linode to a Placement Group through a new `placement_group` field in the `linode_instance` resource, including the ability to reassign a Linode to a different PG. 

~~This PR relies on #1401 and will need to be rebased before being merged. In the meantime, relevant commits can be viewed here: https://github.com/linode/terraform-provider-linode/pull/1406/files/2a9f67928ea55f1c110ebc422259d5528a0358b0..HEAD~~

## ✔️ How to Test

The following test steps assume you have pulled down this PR locally and configured your environment to point to an API instance where Placement Groups are accessible:

```
export LINODE_URL=...
export LINODE_TOKEN=...
```

### Integration Testing

```
make PKG_NAME=linode/instance ARGS="-run TestAccResourceInstance_withPG" int-test
make PKG_NAME=linode/instance ARGS="-run TestAccResourceInstance_pgAssignment" int-test
```

### Manual Testing

```terraform
  label           = "my-instance"
  region          = linode_placement_group.test.region
  type            = "g6-standard-1"

  placement_group {
    id = linode_placement_group.test.id
  }
}

resource "linode_placement_group" "test" {
  label = "my-placement-group"
  region = "us-east"
  affinity_type = "anti_affinity:local"
}
```

3. Navigate to the new instance & PG in Cloud Manager and ensure everything is configured as described in the configuration.
4. Re-plan the configuration to ensure the `label`, `is_strict`, and `affinity_type` fields have been implicitly computed/populated.
5. Apply the following configuration to unassign the Linode from the PG:

```terraform
  label           = "my-instance"
  region          = linode_placement_group.test.region
  type            = "g6-standard-1"
}

resource "linode_placement_group" "test" {
  label = "my-placement-group"
  region = "us-east"
  affinity_type = "anti_affinity:local"
}
```

6. Ensure the apply finishes successfully.
7. Navigate to Cloud Manager and ensure the Linode has been unassigned from the PG.


linode/instance/resource_test.go
```diff
@@ -2327,6 +2327,152 @@ func TestAccResourceInstance_migration(t *testing.T) {
...
+	pgIDs := []string{"g1", "g2"}
+
+	// Resolve a region with support for PGs
+	// TODO: Uncomment once availability issues have been resolved
+	//testRegion, err := acceptance.GetRandomRegionWithCaps(
+	//	[]string{"Linodes", "Placement Group"},
+	//)
+	//if err != nil {
+	//	t.Fatal(err)
+	//}
```
lgarber-akamai, line 2398: I think we'll need to uncomment this in the final project PR (once everything's in prod) 👍

linode/instance/schema_resource.go
```diff
@@ -29,6 +29,38 @@ You may need to switch to an explicit disk configuration.
...
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeInt,
+				Required:    true,
+				Description: "The ID of the Placement Group to assign this Linode to.",
+			},
+			"compliant_only": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
```
lgarber-akamai, line 43: I'm not sure if mixing computed & non-computed fields in a nested block is something we want to do here given the potential complications in framework 🤔

docs/resources/instance.md
```diff
@@ -142,6 +161,8 @@ The following arguments are supported:
 
 * `metadata.0.user_data` - (Optional) The base64-encoded user-defined data exposed to this instance through the Linode Metadata service. Refer to the base64encode(...) function for information on encoding content for this field.
 
+* `placement_group.0.id` - (Optional) The ID of the Placement Group to assign this Linode to.
```
lgarber-akamai, line 164: `compliant_only` is intentionally undocumented here due to an internal decision

linode/instance/helpers.go
```diff
@@ -1476,3 +1476,97 @@ func SafeShutdownInstance(ctx context.Context, client *linodego.Client, instance
...
+
+		return err
+	}
+
+	if err := attemptUnassign(); err != nil {
+		return fmt.Errorf("failed to unassign Linode from Placement Group: %w", err)
+	}
+
+	if err := attemptAssign(); err != nil {
+		return fmt.Errorf("failed to unassign Linode from Placement Group: %w", err)
```
zliang-akamai, line 1548:
```suggestion
		return fmt.Errorf("failed to assign Linode to Placement Group: %w", err)
```

lgarber-akamai: Oops, nice catch!

linode/instance/schema_resource.go
zliang-akamai: I think it will trouble future framework work but I don't think there is a better way to implement it in SDKv2. So, let keep it?

zliang-akamai, line 43: I think this can be computed when not configured by the user?

zliang-akamai: I think both are fine as we might have to migrate to nested attribute at the very very end.

zliang-akamai: Or default to API's default value?

lgarber-akamai: Nice observation! I'm not 100% sure if it'd make sense to make this field computed given it is only specified in the request body for instance creation/assignment to bypass strictness requirements and isn't returned by the API.

lgarber-akamai: Sounds good to me! In that case I think it makes sense to keep things as-is 👍

lgarber-akamai: Hmm, I wonder if we could just inherit from state in the read flatten function, or maybe we could just disable diffs on this field altogether? I think the solution will need to be hacky one way or another given this value isn't stateful on the API 🤔

zliang-akamai (approved):
Looks great! Thanks for addressing them!

zliang-akamai (approved):

linode/helper/sdkv2_cty.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// If the value was defined explicitly as nil, return nil
+	value := d.Get(path)
+	if value == nil {
+		return nil
+	}
+
+	castedValue := value.(T)
+	return &castedValue
+}
```
lgarber-akamai, line 48: We can apply this to any user-defined SDKv2 bool field that we want to differentiate between false and undefined values on (e.g. booted). To keep the scope of this PR simple, I only made use of it for compliant_only 👍

linode/instance/resource.go
```diff
@@ -583,6 +587,26 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
...
+			"new_pg_id": newPGID,
+		})
+
+		if err := reassignPlacementGroup(
+			ctx,
+			client,
+			instance.ID,
+			oldPGID.(int),
+			newPGID.(int),
+			helper.SDKv2UnwrapOptionalConfigAttr[bool](ctx, d, "placement_group.0.compliant_only"),
```
lgarber-akamai, line 604: This is more or less the same thing as how we check that the user has defined the `booted` attribute, but it needs a bit more complexity since `compliant_only` is deeply nested.

linode/helper/sdkv2_cty.go
```diff
@@ -0,0 +1,72 @@
...
+
+// SDKv2PathToCtyPath converts an SDKv2-style path (e.g. foo.0.bar)
+// to a cty-style path object.
+func SDKv2PathToCtyPath(path string) cty.Path {
```
lgarber-akamai, line 52: I was thinking SDKv2 would expose a helper for this somewhere but I couldn't find one :thinking:

ykim-1 (approved):
Tests and manual steps work as expected. Good job!

PR #1410, opened 2024-05-03
Revert "Include resource and data source name as a comment in user-agent"

zliang-akamai:
Reverts linode/terraform-provider-linode#1402

This is causing concurrent map writing issue.

The linodego client has a pointer point to a resty client, and I copied linodego client but the underlying resty client was never copied, and the same resty client is referred by multiple linodego clients. When SetHeader are called from multiple linodego clients, there will be a concurrent map write issue.


ykim-1:
Thanks for addressing this. I will try to monitor the Test run on this PR - https://github.com/linode/terraform-provider-linode/actions/runs/8943255175

ykim-1 (approved):

lgarber-akamai (approved):
LGTM, thanks for handling this!

PR #1417, opened 2024-05-07
Fix concurrent map write on `Schema.Blocks` in the fw base resource

zliang-akamai:
## 📝 Description

This may be better than https://github.com/linode/terraform-provider-linode/pull/1415 because it introduces less file changes.

I am opening this PR for running tests, will fill up more description later.

Test run round 1: https://github.com/linode/terraform-provider-linode/actions/runs/8992245971
Test run round 2: https://github.com/linode/terraform-provider-linode/actions/runs/9005016326
Test run round 3: TBD

## ✔️ How to Test

## 📷 Preview

### DIFF
--- a/linode/helper/framework_resource_base.go
+++ b/linode/helper/framework_resource_base.go
@@ -84,14 +84,14 @@ func (r *BaseResource) Schema(
 		return
 	}
 
-	if r.Config.TimeoutOpts != nil {
-		if r.Config.Schema.Blocks == nil {
-			r.Config.Schema.Blocks = make(map[string]schema.Block)
-		}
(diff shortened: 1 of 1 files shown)


lgarber-akamai:
@zliang-akamai Great work on this! In my opinion, this would be preferable to the first solution since we treat `BaseResourceConfig.Schema` as optional for resources that override the Schema(...) method 👍

lgarber-akamai (approved):

ykim-1 (approved):
This also looks like a cleaner change compare to the first solution. Excellent work and thank you for debugging this. Test runs also look very stable 🚀

PR #1426, opened 2024-05-15
Framework Migration: Firewall Resource

zliang-akamai:
## Description

This is to migrate the firewall resource to use Terraform Framework.

## Test

```bash
make int-test PKG_NAME="linode/firewall" PARALLEL=5
```

```bash
make int-test PKG_NAME="linode/nb" PARALLEL=5
```

```bash
make int-test PKG_NAME="linode/firewalls" PARALLEL=5
```

And you can manually play around with the whole resource, like:
```hcl
  label = "myfirewall"
  tags  = ["test"]

  inbound {
    label    = "tf-test-in"
    action   = "ACCEPT"
    protocol = "TCP"
    ipv4     = ["0.0.0.0/0"]
  }
  inbound_policy  = "DROP"
  outbound_policy = "DROP"
}
```


ezilber-akamai (approved):
Nice work! Tests are all passing locally

jriddle-linode (approved):
Everything's passing locally and it all looks good! nice work.

linode/firewall/framework_resource_test.go
```diff
@@ -420,40 +424,6 @@ func TestAccLinodeFirewall_externalDelete(t *testing.T) {
 	})
 }
 
-func TestAccLinodeFirewall_emptyIPv6(t *testing.T) {
```
yec-akamai, line 423: Just curious, why're we removing this test case?

linode/firewall/framework_resource_test.go
```diff
@@ -126,7 +127,7 @@ func TestAccLinodeFirewall_minimum(t *testing.T) {
 					resource.TestCheckResourceAttr(testFirewallResName, "disabled", "false"),
 					resource.TestCheckResourceAttr(testFirewallResName, "inbound.#", "1"),
 					resource.TestCheckResourceAttr(testFirewallResName, "inbound.0.protocol", "TCP"),
-					resource.TestCheckResourceAttr(testFirewallResName, "inbound.0.ports", ""),
+					// resource.TestCheckResourceAttr(testFirewallResName, "inbound.0.ports", ""),
```
yec-akamai, line 130: probably uncomment this check?

linode/firewall/framework_resource_test.go
zliang-akamai: An empty list of IP is not allowed by the API. You can try this sample call:
```bash
curl --request POST \
     --url https://api.linode.com/v4/networking/firewalls \
     --header 'accept: application/json' \
     --header "authorization: Bearer $TOKEN" \
     --header 'content-type: application/json' \
     --data '
{
  "rules": {
    "inbound_policy": "DROP",
    "outbound_policy": "ACCEPT",
    "inbound": [
      {
        "addresses": {
          "ipv4": [
            "74.102.235.249/32"
          ],
          "ipv6": []
        },
        "action": "DROP",
        "protocol": "UDP"
      }
    ]
  },
  "label": "myfw123"
}
'
```

so I added validators to prevent any empty list input on the `ipv6` and `ipv4` attributes and retired this test case.

linode/firewall/framework_resource_test.go
zliang-akamai: An empty string of `ports` is no longer a valid value in a recent API change. So it would become a `null` now in Terraform state.

linode/firewall/framework_resource_test.go
zliang-akamai: Actually, I can bring this test case back, but the `ipv6` attribute will be changed from `[]` to a `null` lol

yec-akamai: Sounds good!

PR #1428, opened 2024-05-16
new: Add information about Placement Group limits to the `linode_region` and `linode_regions` data sources

lgarber-akamai:
## 📝 Description

This pull request adds the `placement_group_limits` field to the schema for the `linode_region` and `linode_regions` data sources.

## ✔️ How to Test

The following test steps assume you have pulled down this PR locally and pointed your local environment at an API instance where Placement Groups are enabled:

```
export LINODE_URL=...
export LINODE_TOKEN=...
```

### Unit Testing

```
make unit-test
```

### Integration Testing

```
make PKG_NAME=linode/region int-test
make PKG_NAME=linode/regions int-test
```

### Manual Testing

In a Terraform provider sandbox environment (e.g. dx-devenv), apply the following:

```terraform
output "plural-limits" {
  value = {for r in data.linode_regions.test.regions : r.id => r.placement_group_limits}
}

output "singular-limits" {
  value = {(data.linode_region.test.id): data.linode_region.test.placement_group_limits}
}

data "linode_regions" "test" {
  filter {
    name = "capabilities"
    values = ["Placement Group"]
  }
}

data "linode_region" "test" {
  id = data.linode_regions.test.regions.0.id
}
```

2. Ensure the apply runs successfully and the output looks similar to the following:

```
plural-limits = {
  "eu-west" = tolist([
    {
      "maximum_linodes_per_pg" = 5
      "maximum_pgs_per_customer" = 100
    },
  ])
  "us-east" = tolist([
    {
      "maximum_linodes_per_pg" = 5
      "maximum_pgs_per_customer" = 100
    },
  ])
}
singular-limits = {
  "us-east" = tolist([
    {
      "maximum_linodes_per_pg" = 5
      "maximum_pgs_per_customer" = 100
    },
  ])
}
```

### DIFF
--- a/linode/region/framework_datasource_schema.go
+++ b/linode/region/framework_datasource_schema.go
@@ -1,11 +1,19 @@
 package region
 
 import (
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-var frameworkDataSourceSchema = schema.Schema{
+var ObjectTypePGLimits = types.ObjectType{
+	AttrTypes: map[string]attr.Type{
+		"maximum_pgs_per_customer": types.Int64Type,
+		"maximum_linodes_per_pg":   types.Int64Type,
+	},
+}
+
+var DataSourceSchema = schema.Schema{
 	Attributes: map[string]schema.Attribute{
 		"country": schema.StringAttribute{
 			Description: "The country where this Region resides.",
@@ -31,6 +39,11 @@ var frameworkDataSourceSchema = schema.Schema{
 		"status": schema.StringAttribute{
 			Description: "This region’s current operational status.",
 			Computed:    true,
+		},
+		"placement_group_limits": schema.ListAttribute{
+			Description: "Information about placement groups limits for this region.",
+			Computed:    true,
+			ElementType: ObjectTypePGLimits,
 		},
 	},
 	Blocks: map[string]schema.Block{
--- a/linode/regions/framework_datasource_schema.go
+++ b/linode/regions/framework_datasource_schema.go
@@ -2,8 +2,8 @@ package regions
 
 import (
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
-	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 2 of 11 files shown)


linode/region/framework_datasource_schema.go
```diff
@@ -32,6 +40,11 @@ var frameworkDataSourceSchema = schema.Schema{
 			Description: "This region’s current operational status.",
 			Computed:    true,
 		},
+		"placement_group_limits": schema.ListAttribute{
+			Description: "Information about placement groups limits for this region.",
+			Computed:    true,
+			ElementType: ObjectTypePGLimits,
+		},
```
lgarber-akamai, line 47: I was super tempted to implement this as an ObjectAttribute but decided against it to stay consistent with the rest of the provider.

linode/regions/framework_datasource_schema.go
```diff
@@ -24,49 +24,8 @@ var frameworkDataSourceSchema = schema.Schema{
...
-									Computed:    true,
-								},
-								"ipv6": schema.StringAttribute{
-									Description: "The IPv6 addresses for this region’s DNS resolvers, separated by commas.",
-									Computed:    true,
-								},
-							},
-						},
-					},
-				},
```
lgarber-akamai, line 69: I took this opportunity to reuse the schema & model from the singular data source since their implementations were identical 👍

linode/region/framework_datasource_schema.go
yec-akamai: Should we build it into a nested block attribute? Because it's only for read, it doesn't seem to any blocker to do so.

yec-akamai (commented):
LGTM! Tests passed locally. Just a small comment:

linode/region/framework_datasource_schema.go
lgarber-akamai: Good question! I think this is mostly a stylistic choice so it might be worth quickly discussing at Cafe tomorrow

ykim-1 (approved):
Test and local steps pass as expected. Good job!

linode/region/framework_datasource_schema.go
lgarber-akamai: Going to mark this as resolved for now per the discussion in Slack. We can definitely revisit this in the future 👍

yec-akamai (approved):
LGTM!

PR #1436, opened 2024-05-20
new: Support control plane ACL in lke cluster resoure and datasource

yec-akamai:
## 📝 Description

Support control plane ACL CRUD operations in cluster resource and data source. 

## ✔️ How to Test

```
make PKG_NAME="linode/lke" int-test
```

Manual test:

1. In a sandbox environment, i.e. dx-dvenv, run the following tf config to create an LKE cluster with control plane ACL:
```
    label       = "my-cluster"     
    k8s_version = "1.28"           
    region      = "us-east"     
    tags        = ["prod"]         
    control_plane {
        acl {
            enabled = true
            addresses {
                ipv4 = "0.0.0.0/0"
                ipv6 = "2001:db8::/32"
            }
        }
    }
    pool {
        type  = "g6-standard-2"
        count = 1
    }
}
```
2. Run the above plan and observe the resource created successfully.
3. Disable the ACL
```
    label       = "my-cluster"     
    k8s_version = "1.28"           
    region      = "us-east"     
    tags        = ["prod"]         
    control_plane {
        acl {
            enabled = false
        }
    }
    pool {
        type  = "g6-standard-2"
        count = 1
    }
}
```
5. Read the data source from the updated LKE
```
}
```
6. Change around the configs to see if there is any edge case


go.mod
```diff
@@ -31,6 +31,8 @@ require (
 	golang.org/x/sync v0.7.0
 )
 
+replace github.com/linode/linodego => github.com/linode/linodego v1.33.2-0.20240520152533-30da20d4b936
```
yec-akamai, line 34: The LKE ACL change in linodego hasn't been released. Will make sure we've released it before merge this PR.

linode/lke/framework_datasource.go
```diff
@@ -97,7 +98,25 @@ func (r *DataSource) Read(
 		return
 	}
 
-	resp.Diagnostics.Append(data.parseLKEAttributes(ctx, cluster, pools, kubeconfig, endpoints, dashboard)...)
+	acl, err := client.GetLKEClusterControlPlaneACL(ctx, clusterId)
+	if err != nil {
+		if lerr, ok := err.(*linodego.Error); ok &&
+			lerr.Code == 400 && strings.Contains(lerr.Message, "Cluster does not support Control Plane ACL") {
```
yec-akamai, line 104: Because ACL isn't returned from the GET cluster endpoint, we have to make an extra call here to get the control plane ACL. There is a tricky case that if this cluster doesn't have an ACL, which is optional, calling GET ACL endpoint will return a 400 and a corresponding error message. I don't think this error should break the data source or resource since ACL is totally optional, so I build it into a warning and continue. Any comment is welcomed!

linode/lke/resource.go
```diff
@@ -96,7 +96,18 @@ func readResource(ctx context.Context, d *schema.ResourceData, meta interface{})
 		return diag.Errorf("failed to get API endpoints for LKE cluster %d: %s", id, err)
 	}
 
-	flattenedControlPlane := flattenLKEClusterControlPlane(cluster.ControlPlane)
+	acl, err := client.GetLKEClusterControlPlaneACL(ctx, id)
+	if err != nil {
+		if lerr, ok := err.(*linodego.Error); ok &&
+			lerr.Code == 400 && strings.Contains(lerr.Message, "Cluster does not support Control Plane ACL") {
+			// The cluster does not have a Gateway
+			tflog.Warn(ctx, fmt.Sprintf("This cluster does not have a Control Plane ACL. %v", err.Error()))
```
lgarber-akamai, line 104: Given this is in the read routine this seems to be a valid case, so we may not need to warn here 👍

linode/lke/schema_resource.go
```diff
@@ -143,8 +144,48 @@ var resourceSchema = map[string]*schema.Schema{
...
+							},
+							"addresses": {
+								Type:        schema.TypeList,
+								Description: "A list of ip addresses to allow.",
+								Optional:    true,
+								Computed:    true,
+								Elem: &schema.Resource{
+									Schema: map[string]*schema.Schema{
+										"ipv4": {
+											Type:        schema.TypeList,
```
lgarber-akamai, line 169: Would it make sense for the IPv4 and IPv6 fields to become `TypeSet` since we can't guarantee their order?

linode/lke/schema_resource.go
```diff
@@ -143,8 +144,48 @@ var resourceSchema = map[string]*schema.Schema{
 					Optional:    true,
 					Computed:    true,
 				},
+				"acl": {
```
lgarber-akamai, line 147: I noticed dropping the `acl` block after applying a configuration with a valid ACL doesn't trigger any updates. I think this behavior is probably fine/expected but I just wanted to bring attention to it just in case 🙂

linode/lkeclusters/framework_models.go
```diff
@@ -22,15 +21,19 @@ type LKEClusterFilterModel struct {
...
+	Updated      types.String    `tfsdk:"updated"`
+	Label        types.String    `tfsdk:"label"`
+	Region       types.String    `tfsdk:"region"`
+	Status       types.String    `tfsdk:"status"`
+	K8sVersion   types.String    `tfsdk:"k8s_version"`
+	Tags         types.Set       `tfsdk:"tags"`
+	ControlPlane LKEControlPlane `tfsdk:"control_plane"`
+}
+
+type LKEControlPlane struct {
```
lgarber-akamai, line 35: nitpick: 
```suggestion
type LKEControlPlaneModel struct {
```

linode/lke/framework_datasource.go
lgarber-akamai: Same as [my other comment](https://github.com/linode/terraform-provider-linode/pull/1436/files#r1608461131), since the ACL not being available on a cluster is a valid case I think it's fine to drop the warning here ^

lgarber-akamai (commented):
Just a few small changes and this should be good to go, excellent work!

lgarber-akamai (approved):
LGTM, nice work!

jriddle-linode (approved):
LGTM! working on my end!

Issue #1447, opened 2024-05-30
[Bug]: failed to get control plane ACL for LKE cluster XXXXXX: [404] Not found


idogada-akamai:
### Terraform Version

1.7.1

### Linode Provider Version

2.21.0

### Effected Terraform Resources

linode_lke_cluster

### Terraform Config Files

```hcl
resource "linode_lke_cluster" "kubernetes-cluster" {
  k8s_version = var.k8s_cluster.version
  label       = var.k8s_cluster.name
  tags        = concat([var.k8s_cluster.name], var.k8s_cluster.tags)
  region      = var.k8s_cluster.region
  dynamic "pool" {
    for_each = var.k8s_cluster.pools
    content {
      type  = pool.value["type"]
      count = pool.value["min"]
      autoscaler {
        min = pool.value["min"]
        max = pool.value["max"]
      }
    }
  }

  control_plane {
    high_availability = var.k8s_cluster.ha_control_plane
  }

  lifecycle {
    ignore_changes = [
      pool.0.count,
      pool.1.count
    ]
  }
}
```

### Expected Behavior

Linode provider not failing when ACL is not set on the cluster

### Actual Behavior

Linode provider is trying to get ACL even though it's marked as optional, and failing

### Steps to Reproduce

* Create a cluster without ACL
* Run Terraform apply a 2nd time

idogada-akamai:
@yec-akamai When is this release will marked as latest in the Hashicorp registry? It still shows `2.21.0`
https://registry.terraform.io/providers/linode/linode/latest/docs

PR #1449, opened 2024-05-30
fix: Ignore 404 when reading LKE ACL

yec-akamai:
## 📝 Description

Ignore the 404 error returned from the ACL read path because of missing customer permission.

Missing permission on the ACL write path will still return a 403 error.

Addressed #1447 

## ✔️ How to Test

```
make PKG_NAME="linode/lke" int-test
```

Manual test:

1. Remove the required tag from your account to perform this test
2. Create a LKE cluster without ACL in the control plane
```
    label       = "my-cluster"     
    k8s_version = "1.28"           
    region      = "us-east"     
    tags        = ["prod"]         
    control_plane {
        high_availability = true
    }
    pool {
        type  = "g6-standard-2"
        count = 1
    }
}
```
3. Observe no error raised
4. Run the following config to read the cluster
```
}
```
5. Obeserve no error raised during this process

### DIFF
--- a/docs/resources/lke_cluster.md
+++ b/docs/resources/lke_cluster.md
@@ -62,8 +62,8 @@ resource "linode_lke_cluster" "test" {
         acl {
             enabled = true
             addresses {
-                ipv4 = "0.0.0.0/0"
-                ipv6 = "2001:db8::/32"
+                ipv4 = ["0.0.0.0/0"]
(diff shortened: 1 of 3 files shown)


lgarber-akamai (approved):
Looks great, thanks for fixing this!

ykim-1 (approved):
Nice fix, tests passed locally

PR #1466, opened 2024-06-06
Deprecates `linode_object_storage_cluster` data source

zliang-akamai:
## 📝 Description

Deprecating it due to the deprecation of the corresponding API.

## ✔️ How to Test

Applying this data source:

`tofu apply` or `terraform apply`

```terraform
data "linode_object_storage_cluster" "primary" {
    id = "us-mia-1"
}
```

A deprecated warning would be expected:

```
OpenTofu has compared your real infrastructure against your configuration and found no
differences, so no changes are needed.
╷
│ Warning: Deprecated
│ 
│   with data.linode_object_storage_cluster.primary,
│   on main.tf line 15, in data "linode_object_storage_cluster" "primary":
│   15: data "linode_object_storage_cluster" "primary" {
│ 
│ 'linode_object_storage_cluster' data source has been deprecated because it uses a deprecated
│ View Cluster API. In the future, 'region' will be preferred over 'cluster' for the request
│ body of many other Object Storage related API endpoints because there may be multiple clusters
│ per region.
│ 
│ (and one more similar warning elsewhere)
╵
```

### DIFF
--- a/linode/objcluster/framework_datasource_schema.go
+++ b/linode/objcluster/framework_datasource_schema.go
@@ -5,6 +5,11 @@ import (
 )
 
 var frameworkDatasourceSchema = schema.Schema{
+	DeprecationMessage: "'linode_object_storage_cluster' data source " +
+		"has been deprecated because it uses a deprecated View Cluster " +
+		"API. In the future, 'region' will be preferred over 'cluster' " +
+		"for the request body of many other Object Storage related API " +
+		"endpoints because there may be multiple clusters per region.",
 	Attributes: map[string]schema.Attribute{
 		"id": schema.StringAttribute{
 			Description: "The unique ID of this Cluster.",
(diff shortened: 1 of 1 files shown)


linode/objcluster/framework_datasource_schema.go
```diff
@@ -5,6 +5,11 @@ import (
 )
 
 var frameworkDatasourceSchema = schema.Schema{
+	DeprecationMessage: "'linode_object_storage_cluster' data source " +
+		"has been deprecated because it uses a deprecated View Cluster " +
+		"API. In the future, 'region' will be preferred over 'cluster' " +
+		"for the request body of many other Object Storage related API " +
+		"endpoints because there may be multiple clusters per region.",
```
lgarber-akamai, line 12: We might be able to condense this down to something like 👍 
```
This data source has been deprecated because it relies on deprecated API endpoints. Going forward, `region` will be the preferred way to designate where a bucket should be created.
```

ykim-1 (approved):
LGTM after Lena's comment

lgarber-akamai (approved):
Looks perfect, great work!

PR #1467, opened 2024-06-07
Reuse the client if it is already configured in fwprovider

tchinmai7:
## 📝 Description

Crossplane provider-linode expects to use a single configured instance of the linode client across all invocations. However, due to how upjet operates, the configureProvider() gets invoked on every resource call. To preserve the client, see if the fp.Meta.Client is already initialized, and if so, re-use it.
## ✔️ How to Test

The main thing here should be verifying it doesn't break any usual terraform flows.

## 📷 Preview


linode/framework_provider_config.go
```diff
@@ -215,6 +215,17 @@ func (fp *FrameworkProvider) InitProvider(
...
 	meta *helper.FrameworkProviderMeta,
 ) {
+	if fp.Meta.Client != nil {
+		// Crossplane provider-linode expects to use a single configured instance of the linode client across all invocations
+		// However, due to how upjet operates, the configureProvider() gets invoked on every resource call. To preserve the client,
+		// see if the fp.Meta.Client is already initialized, and if so, re-use it.
+
+		tflog.Info(ctx, "Linode client was already configured, re-using..")
+		meta.Client = fp.Meta.Client
+		meta.Config = fp.Meta.Config
```
lgarber-akamai, line 225: I can't think of a case where this would happen, but could we raise an error here if `fp.Meta.Config` is nil just to pick up any accidental edge cases/regressions?

lgarber-akamai (approved):
Everything seems to work as expected on my local machine, thanks for the contribution!

linode/framework_provider_config.go
tchinmai7: could add a check for this too - but its a struct not a pointer

jriddle-linode (approved):
LGTM thank you!!

zliang-akamai (approved):

PR #1471, opened 2024-06-11
fix: Resolve nil pointer dereference in FrameworkProvider{}.InitProvider(...) method

lgarber-akamai:
## 📝 Description

This pull request resolves an issue in the `FrameworkProvider{}.InitProvider(...)` idempotence logic that would cause the following error to be raised:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1f86cc9]
```

## ✔️ How to Test

### Integration Testing

```
make PKG_NAME=linode/region int-test
```

### Manual Testing

N/A

### DIFF
--- a/linode/framework_provider_config.go
+++ b/linode/framework_provider_config.go
@@ -215,7 +215,7 @@ func (fp *FrameworkProvider) InitProvider(
 	diags *diag.Diagnostics,
 	meta *helper.FrameworkProviderMeta,
 ) {
-	if fp.Meta.Client != nil {
+	if fp.Meta != nil && fp.Meta.Client != nil {
 		// Crossplane provider-linode expects to use a single configured instance of the linode client across all invocations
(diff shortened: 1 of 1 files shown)


zliang-akamai (approved):

linode/framework_provider_config.go
```diff
@@ -215,7 +215,7 @@ func (fp *FrameworkProvider) InitProvider(
 	diags *diag.Diagnostics,
 	meta *helper.FrameworkProviderMeta,
 ) {
-	if fp.Meta.Client != nil {
+	if fp.Meta != nil && fp.Meta.Client != nil {
```
tchinmai7, line 218: oops 😅

tchinmai7 (approved):

ykim-1 (approved):
Thanks for the quick fix 👍

PR #1476, opened 2024-06-18
doc: Add documentation notes for Placement Groups LA

lgarber-akamai:
## 📝 Description

This change adds notes the relevant resource/data source docs informing users that PGs may not currently be available to all users.

## ✔️ How to Test

N/A

### DIFF
--- a/docs/data-sources/placement_group.md
+++ b/docs/data-sources/placement_group.md
@@ -6,6 +6,8 @@ description: |-
 
 # Data Source: linode\_placement\_group
 
+**NOTE: Placement Groups may not currently be available to all users.**
+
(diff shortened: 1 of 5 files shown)


yec-akamai (approved):
LGTM!

Just come to my mind that do we have a ticket to remove these notes when placement group goes to GA?

lgarber-akamai:
> LGTM!
> 
> Just come to my mind that do we have a ticket to remove these notes when placement group goes to GA?

@yec-akamai Good point! I'll make one as a follow up to the current ticket 👍

jriddle-linode (approved):

PR #1475, opened 2024-06-18
Release v2.22.1

lgarber-akamai:

PR #1477, opened 2024-06-18
test: add environment file to cloud firewall implementation to stabilize builds and reduce execution time

ykim-1:
## 📝 Description

Addressing TLS handshake failures - https://github.com/linode/terraform-provider-linode/pull/1475

Now executing `int-test` will create an env file in tmp directory containing ipv4 and ipv6 information which will be read in `e2e_test_firewall.gotf` script. This will greatly improve execution time of the workflow and also avoid TLS handshake failures

## ✔️ How to Test

Forked run - https://github.com/ykim-1/terraform-provider-linode/actions/runs/9571535532

## 📷 Preview

### DIFF
--- a/scripts/generate_ip_env_fw_e2e.sh
+++ b/scripts/generate_ip_env_fw_e2e.sh
@@ -4,7 +4,7 @@
 PUBLIC_IPV4=$(curl -s https://api.ipify.org)
 
 # Fetch public IPv6 address
-PUBLIC_IPV6=$(curl -s https://ifconfig.co/ip)
+PUBLIC_IPV6=$(curl -s https://api64.ipify.org)
 
 # Create .env file in /tmp directory
 ENV_FILE="/tmp/ip_vars.env"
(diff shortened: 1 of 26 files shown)


scripts/generate_ip_env_fw_e2e.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Fetch public IPv4 address
+PUBLIC_IPV4=$(curl -s https://api.ipify.org)
+
+# Fetch public IPv6 address
+PUBLIC_IPV6=$(curl -s https://api64.ipify.org)
+
+# Create .env file in /tmp directory
+ENV_FILE="/tmp/ip_vars.env"
```
lgarber-akamai, line 10: Would it make sense to store this file in something like `/tmp/linode/ip_vars.env` just to scope it a bit more tightly?

lgarber-akamai (approved):
Seems to be working well on my end other than the optional feedback above, great work!

PR #1493, opened 2024-06-26
Support Region for Object Storage Bucket Resource

zliang-akamai:
## 📝 Description

This is to support region for `linode_object_storage_bucket` resource.

### Automated Testing

```bash
make PKG_NAME=linode/objbucket int-test
```

Note that you may see error like this, which is a previously introduced and known issue due to a TF bug, which should not be relevant to the change in this PR.
```
=== NAME  TestAccResourceBucket_forceDelete
    test_retry.go:25: Retrying on test failure: Step 1/2 error: Error running apply: exit status 1
        
        Error: [400] [bucket_access[0].bucket_name] Bucket not found
        
          with linode_object_storage_bucket.foobar,
          on terraform_plugin_test.tf line 18, in resource "linode_object_storage_bucket" "foobar":
          18: resource "linode_object_storage_bucket" "foobar" {
```

#### Region Mode

```terraform
resource "linode_object_storage_key" "mykey" {
  label = "image-access"
}

resource "linode_object_storage_bucket" "mybucket" {
  access_key = linode_object_storage_key.mykey.access_key
  secret_key = linode_object_storage_key.mykey.secret_key

  region = "us-mia"
  label  = "mybucket"
}
```

#### Cluster Mode
```terraform
resource "linode_object_storage_key" "mykey" {
  label = "image-access"
}

resource "linode_object_storage_bucket" "mybucket" {
  access_key = linode_object_storage_key.mykey.access_key
  secret_key = linode_object_storage_key.mykey.secret_key

  cluster = "us-mia-1"
  label  = "mybucket"
}
```


linode/objbucket/resource_test.go
```diff
@@ -556,9 +556,16 @@ func TestAccResourceBucket_forceDelete(t *testing.T) {
 							}},
 						}
 
-						keys, _ := client.CreateObjectStorageKey(context.Background(), createOpts)
+						keys, err := client.CreateObjectStorageKey(context.Background(), createOpts)
+						if err != nil {
+							t.Errorf("error creating obj keys in PreConfig func: %v", err)
```
lgarber-akamai, line 561: I think we'll need to reference `retryT` here rather than `t` to have it properly accounted for in the retry system 👍

linode/obj/helpers.go
```diff
@@ -38,23 +39,37 @@ func getObjKeysFromProvider(
...
+	bucket, regionOrCluster, permissions string,
 ) (*linodego.ObjectStorageKey, diag.Diagnostics) {
 	tflog.Debug(ctx, "Create temporary object storage access keys implicitly.")
 
+	tempBucketAccess := linodego.ObjectStorageKeyBucketAccess{
+		BucketName:  bucket,
+		Permissions: permissions,
+	}
+
+	if isCluster(regionOrCluster) {
```
lgarber-akamai, line 63: I'm a little weary about distinguishing cluster from region using regex validation, although in theory it should be fine.

I wonder if it would make sense to maybe do something along the lines of:

```go
type BucketParent struct {
	Cluster *string
	Region *string
}

func (p BucketParent) ClusterOk() (string, bool) {
	if p.Cluster != nil {
		return *p.Cluster, true
	}

	return "", false
}

func (p BucketParent) RegionOk() (string, bool) {
	if p.Region != nil {
		return *p.Region, true
	}

	return "", false
}

// ...

func createTempKeys(
	ctx context.Context,
	client linodego.Client,
	parent BucketParent,
	bucket, permissions string,
) (*linodego.ObjectStorageKey, diag.Diagnostics) {
	// ...

	if region, ok := parent.RegionOk(); ok {
		tempBucketAccess.Region = region
	}

	if cluster, ok := parent.ClusterOk(); ok {
		tempBucketAccess.Cluster = cluster
	}
}
```
type

zliang-akamai: As discussed, this might be hard to implement due to cluster/region is read from resource ID in [some cases](https://github.com/linode/terraform-provider-linode/pull/1493/files#diff-55090792a29210600c60d24b798c34102d498a87d8b76d9e4c53098c2ad95639L60), and there isn't another obvious way to distinguish them without regex or string parsing.

lgarber-akamai (approved):
Tests are passing on my end, excellent work!

ykim-1 (approved):
LGTM, tests pass locally

PR #1496, opened 2024-06-27
Replace deprecated aws sdk `EndpointResolver`

zliang-akamai:
## 📝 Description

> config.WithEndpointResolver is deprecated: The global endpoint resolution interface is deprecated. The API for endpoint resolution is now unique to each service and is set via the EndpointResolverV2 field on service client options.

This is to replace deprecated `EndpointResolver` with the options manipulation function in S3 service client.

## ✔️ How to Test

```bash
make PKG_NAME=linode/obj int-test
```

```bash
make PKG_NAME=linode/objbucket int-test
```


ezilber-akamai (approved):

yec-akamai (approved):
Works well locally!

PR #1520, opened 2024-07-16
Change all API Doc urls to point to new Techdocs

jriddle-linode:
## 📝 Description

Updates all `developer.linode.com` urls to point to the new `https://techdocs.akamai.com/`
Also adds a lot of urls that weren't previously there.

### DIFF
--- a/docs/data-sources/account.md
+++ b/docs/data-sources/account.md
@@ -7,6 +7,7 @@ description: |-
 # Data Source: linode\_account
 
 Provides information about a Linode account.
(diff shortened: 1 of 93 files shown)


ykim-1 (approved):
LGTM, links all seem to direct to the correct pages. Nice work!

lgarber-akamai (commented):
@jriddle-linode Looks like there are still a few residual legacy API links shown by:

```
grep -rni "/docs/api" .
```

Otherwise this looks great, nice work!

zliang-akamai (approved):

PR #1530, opened 2024-07-26
Support tags in nested node pool

zliang-akamai:
## 📝 Description

This is to support tags on nested node pools in a LKE cluster resource.

### Automated Testing
```bash
make int-test PKG_NAME="linode/lke"
```

### Manual Testing

```terraform
    label       = "my-cluster"
    k8s_version = "1.30"
    region      = "us-mia"
    tags        = ["prod"]

    pool {
        type  = "g6-standard-1"
        count = 1
        tags = ["prod"]
    }
}
```


linode/lke/schema_resource.go
```diff
@@ -79,6 +83,32 @@ var resourceSchema = map[string]*schema.Schema{
 					Description: "A Linode Type for all of the nodes in the Node Pool.",
 					Required:    true,
 				},
+				"tags": {
+					Type:                  schema.TypeSet,
+					Description:           "An array of tags applied to this node pool.",
```
lgarber-akamai, line 88: Super minor nitpick: 
```suggestion
					Description:           "A set of tags applied to this node pool.",

```

lgarber-akamai:
I'm getting some weird diffs when applying the following config twice, although AFAIK fortunately the fix should just requiring adding an extra check here: https://github.com/zliang-akamai/terraform-provider-linode/blob/zhiwei/node-pool-tag/linode/lke/cluster.go#L339

Config:
```terraform
  label       = "my-cluster"
  k8s_version = "1.30"
  region      = "us-mia"
  tags        = ["prod"]

  pool {
    type  = "g6-standard-1"
    count = 1
    tags = ["prod"]
  }

  pool {
    type  = "g6-standard-1"
    count = 1
    tags = ["prod"]
  }

  pool {
    type  = "g6-standard-1"
    count = 1
    tags = ["cool", "prod"]
  }

  pool {
    type  = "g6-standard-1"
    count = 1
    tags = ["wow"]
  }
}
```

Plan:
```
# linode_lke_cluster.my-cluster will be updated in-place
  ~ resource "linode_lke_cluster" "my-cluster" {
        id            = "204853"
        tags          = [
            "prod",
        ]

      ~ pool {
            id    = 296939
          ~ tags  = [
              - "wow",
              + "prod",
            ]
            # (3 unchanged attributes hidden)
        }
      ~ pool {
            id    = 296937
          ~ tags  = [
                "prod",
              + "cool",
            ]
            # (3 unchanged attributes hidden)
        }
      ~ pool {
            id    = 296938
          ~ tags  = [
              - "prod",
              - "cool",
              + "wow",
            ]
            # (3 unchanged attributes hidden)
        }

        # (2 unchanged blocks hidden)
    }
```

lgarber-akamai:
Excellent work so far! Just a few small things and this should be good to go 🎉

zliang-akamai:
@lgarber-akamai weirdly I couldn't reproduce the issue... But I did add the check to make it logically correct. Can you help me to test it again on your end to see if the issue is still there after the fix? Thanks!

lgarber-akamai:
> @lgarber-akamai weirdly I couldn't reproduce the issue... But I did add the check to make it logically correct. Can you help me to test it again on your end to see if the issue is still there after the fix? Thanks!

@zliang-akamai Sure, I'm on it! I'm guessing it's inconsistent to reproduce because it only occurs when the API returns pools in an unexpected order

linode/lke/cluster.go
```diff
@@ -337,6 +338,10 @@ func matchPoolsWithSchema(pools []linodego.LKENodePool, declaredPools []interfac
 				continue
 			}
 
+			if !slices.Equal(helper.ExpandStringSet(declaredPool["tags"].(*schema.Set)), apiPool.Tags) {
```
lgarber-akamai, line 341: Since the tags are unordered, would it make sense to sort both tag slices before comparing here?

zliang-akamai: @lgarber-akamai good point! I checked the underlying docs and Terraform SDK already handle sorting in their set's `List()` method.

> The order of the returned elements is deterministic. Given the same set, the order of this will always be the same. 

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema#Set.List

zliang-akamai: @lgarber-akamai good point! I thought TF SDK handled it but I forgot another side is from the API..

linode/lke/cluster.go
```diff
@@ -338,7 +337,7 @@ func matchPoolsWithSchema(pools []linodego.LKENodePool, declaredPools []interfac
 				continue
 			}
 
-			if !slices.Equal(helper.ExpandStringSet(declaredPool["tags"].(*schema.Set)), apiPool.Tags) {
+			if helper.CompareStringSets(helper.ExpandStringSet(declaredPool["tags"].(*schema.Set)), apiPool.Tags) {
```
lgarber-akamai, line 340: I think there might be a missing `!` (not) here 👀

lgarber-akamai (approved):
Looks perfect after the above comment is addressed!

Tested by patching provider with logic to shuffle returned node pools:

```bash
# Apply pool shuffling patch
curl https://gist.githubusercontent.com/lgarber-akamai/8659f3556dbf27d3eb67b21cea34f3b3/raw/shuffle_pools.patch | git apply --allow-empty

# Enter the dx-devenv sandbox
cd path/to/my/dx-devenv/terraform/sandbox

# Download a `main.tf` with a bunch of pools with minor tag variations
curl https://gist.githubusercontent.com/lgarber-akamai/8659f3556dbf27d3eb67b21cea34f3b3/raw/main.tf -o main.tf

# Create the cluster
make apply 

# Ensure no pool diffs exist in output
make plan 
```

linode/lke/cluster.go
zliang-akamai: Ah sorry, just added. Thanks!

PR #1534, opened 2024-07-31
Add deprecate message to cluster attributes

zliang-akamai:
## 📝 Description

This is to deprecate `cluster` attribute in favor of `region` attribute.


ezilber-akamai (approved):

lgarber-akamai (approved):
LGTM, nice work!

PR #1543, opened 2024-08-07
Fix flatten image panic when error happens in image creation

zliang-akamai:
## 📝 Description

This is to inspect `resp.Diagnostics` and return early if there is an error.

## ✔️ How to Test
```bash
make int-test PKG_NAME="linode/image"
```


lgarber-akamai (approved):
Thanks for bringing this fix back to `dev`, looks great!

yec-akamai (approved):

ezilber-akamai (approved):
LGTM!

PR #1550, opened 2024-08-13
Implemented VPC IP List data source

ezilber-akamai:
## 📝 Description

Created a data source to support both the `/v4/vpcs/:id/ips` and /`v4/vpcs/ips` endpoints.

### Integration Tests
`make int-test PKG_NAME="linode/vpcips"`


linode/vpc_ips/datasource_test.go
```diff
@@ -0,0 +1,71 @@
+//go:build integration || vpc_ips
+
+package vpc_ips_test
```
lgarber-akamai, line 3: nitpick: Could this module be named `vpcips` to align with the naming of the other modules?

linode/vpc_ips/framework_datasource.go
```diff
@@ -0,0 +1,113 @@
...
+	data.FlattenVPCIPs(helper.AnySliceToTyped[linodego.VPCIP](result), false)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
+}
+
+func ListAllVPCIPs(ctx context.Context, client *linodego.Client, filter string) ([]any, error) {
+	tflog.Trace(ctx, "client.ListVPCIPs(...)", map[string]any{
```
lgarber-akamai, line 88:
```suggestion
	tflog.Trace(ctx, "client.ListAllVPCIPAddresses(...)", map[string]any{
```

linode/vpc_ips/framework_datasource.go
```diff
@@ -0,0 +1,113 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return helper.TypedSliceToAny(vpcIps), nil
+}
+
+func ListVPCIPs(ctx context.Context, client *linodego.Client, filter string, vpcID int) ([]any, error) {
+	tflog.Trace(ctx, "client.ListVPCIPs(...)", map[string]any{
```
lgarber-akamai, line 102:
```suggestion
	tflog.Trace(ctx, "client.ListVPCIPAddresses(...)", map[string]any{
```

lgarber-akamai (approved):
Looks great other than the nitpicks above, nice work!

linode/vpc_ips/datasource_test.go
ezilber-akamai: Good call. Fixed!

linode/vpc_ips/framework_datasource.go
ezilber-akamai: Fixed!

linode/vpc_ips/framework_datasource.go
ezilber-akamai: Fixed!

lgarber-akamai (approved):

linode/vpcips/framework_datasource.go
```diff
@@ -0,0 +1,113 @@
...
+	// Check if VPC ID is provided
+	if !data.VPCID.IsNull() {
+		tflog.Debug(ctx, "Filtering IPs for specific VPC", map[string]interface{}{
+			"vpc_id": data.VPCID.ValueInt64(),
+		})
+
+		result, d = filterConfig.GetAndFilter(
+			ctx, r.Meta.Client, data.Filters,
+			func(ctx context.Context, client *linodego.Client, filter string) ([]any, error) {
+				return ListVPCIPs(ctx, client, filter, int(data.VPCID.ValueInt64()))
```
zliang-akamai, line 60: int64 to int conversion may required `helper.FrameworkSafeInt64ToInt`

linode/vpcips/framework_models.go
```diff
@@ -0,0 +1,78 @@
...
+	m.LinodeID = helper.KeepOrUpdateValue(m.LinodeID, other.LinodeID, preserveKnown)
+	m.Region = helper.KeepOrUpdateValue(m.Region, other.Region, preserveKnown)
+	m.Active = helper.KeepOrUpdateValue(m.Active, other.Active, preserveKnown)
+	m.NAT1To1 = helper.KeepOrUpdateValue(m.NAT1To1, other.NAT1To1, preserveKnown)
+	m.VPCID = helper.KeepOrUpdateValue(m.VPCID, other.VPCID, preserveKnown)
+	m.SubnetID = helper.KeepOrUpdateValue(m.SubnetID, other.SubnetID, preserveKnown)
+	m.ConfigID = helper.KeepOrUpdateValue(m.ConfigID, other.ConfigID, preserveKnown)
+	m.InterfaceID = helper.KeepOrUpdateValue(m.InterfaceID, other.InterfaceID, preserveKnown)
+}
+
```
zliang-akamai, line 57: `CopyFrom` is actually not needed for a data source, because it's only required to eliminate remaining unknown values after flatten of an linodego object after an update API call in the `Update` function in a resource

linode/vpcips/datasource_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestAccDataSourceVPCIPs_basic(t *testing.T) {
+	t.Parallel()
+
+	resourceName_all_ips := "data.linode_vpc_ips.foobar"
+	resourceName_vpc_ips := "data.linode_vpc_ips.barfoo"
+	resourceName_filtered_ips := "data.linode_vpc_ips.foobar_filter"
+
+	vpcLabel := acctest.RandomWithPrefix("tf-test")
+	testRegion, err := acceptance.GetRandomRegionWithCaps([]string{"VPCs"})
```
zliang-akamai, line 24:
```suggestion
	testRegion, err := acceptance.GetRandomRegionWithCaps([]string{"VPCs"}, "core")
```

Don't forget this parameter updated in dev branch so we won't need another PR to fix it lol.
(only apply this change after merging dev to this branch)

linode/vpcips/datasource_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acceptance.PreCheck(t) },
+		ProtoV5ProviderFactories: acceptance.ProtoV5ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: tmpl.DataBasic(t, vpcLabel, testRegion, "10.0.0.0/24", "10.0.1.0/24"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName_all_ips, "vpc_ips.#", "2"),
+					resource.TestCheckResourceAttr(resourceName_vpc_ips, "vpc_ips.#", "1"),
```
zliang-akamai, line 37: Assertions like these may fail when there are other VPC instances live in the same account, which is very likely for our shared testing account because multiple tests may run concurrently. I would suggest adding filter to every test case to ensure only the VPC created by this test workflow are selected by the data source in the template files.

linode/vpcips/framework_schema_datasource.go
```diff
@@ -0,0 +1,91 @@
...
+		Computed:    true,
+	},
+}
+
+var filterConfig = frameworkfilter.Config{
+	"active":    {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeBool},
+	"config_id": {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeInt},
+	"linode_id": {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeInt},
+	"region":    {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeString},
+	"vpc_id":    {APIFilterable: false, TypeFunc: frameworkfilter.FilterTypeInt},
```
zliang-akamai, line 68: I think these attributes are API filterable?

linode/vpcips/framework_datasource.go
ezilber-akamai: Good to know, thanks for the tip!

linode/vpcips/framework_models.go
ezilber-akamai: Thanks for explaining! I removed it 👍

linode/vpcips/datasource_test.go
ezilber-akamai: Good catch! Fixed this.

linode/vpcips/datasource_test.go
ezilber-akamai: Good point, I didn't consider this. I updated the first test case to filter on the ids of the two newly created linodes. For the second test, I did not add a filter because the datasource is given the id of the newly created vpc, so it should only ever return the IP Addresses relevant to the test. Let me know if you think I should add a filter anyway.

linode/vpcips/framework_schema_datasource.go
ezilber-akamai: Thanks, I fixed this! Just for future reference, where can you find if a field is listed as API Filterable?

zliang-akamai: I think the API filterable fields are noted in the API docs?
image
https://techdocs.akamai.com/linode-api/reference/get-vpcs-ips

linode/vpcips/framework_datasource.go
```diff
@@ -57,7 +57,7 @@ func (r *DataSource) Read(
 		result, d = filterConfig.GetAndFilter(
 			ctx, r.Meta.Client, data.Filters,
 			func(ctx context.Context, client *linodego.Client, filter string) ([]any, error) {
-				return ListVPCIPs(ctx, client, filter, int(data.VPCID.ValueInt64()))
+				return ListVPCIPs(ctx, client, filter, helper.FrameworkSafeInt64ToInt(data.VPCID.ValueInt64(), &resp.Diagnostics))
```
zliang-akamai, line 60: Since `FrameworkSafeInt64ToInt` may place an error in `resp.Diagnostics`, I think we can convert this ID around line 53-56, directly in the if block and then we can inspect the error there and return earlier as needed. How does that sound?

linode/vpcips/framework_schema_datasource.go
ezilber-akamai: Oh that makes sense! So fields with API Filterable set to false are additional filters added to terraform that are not supported at the API level?

linode/vpcips/datasource_test.go
zliang-akamai: That looks great! Thanks for addressing it!

linode/vpcips/framework_datasource.go
ezilber-akamai: Good call, I think that is a better solution. Thanks!

linode/vpcips/framework_schema_datasource.go
zliang-akamai: Yes, I think so

linode/vpcips/framework_datasource.go
ezilber-akamai: Fixed!

zliang-akamai (approved):
Looks good to me, nice work!

PR #1555, opened 2024-08-15
test: Fix TestAccResourceInstance_privateImage integration test

yec-akamai:
## 📝 Description

Update the test case to create with booted instance and imagize it. 

## ✔️ How to Test

```
make int-test PKG_NAME="linode/instance" ARGS="-run TestAccResourceInstance_privateImage"
```


lgarber-akamai (approved):
LGTM, thanks for the quick fix!

zliang-akamai (approved):

ykim-1 (approved):
Thanks for the quick fix! Test pass locally but we should definitely document or come up with a solution for edge case like this.

PR #1560, opened 2024-08-19
Fix nil pointer issue in disk resource when failed to wait for a disk

zliang-akamai:
## 📝 Description
This is to fix a nil pointer issue in the Linode Disk resource.


jriddle-linode (approved):

yec-akamai (approved):
LGTM

Issue #1536, opened 2024-08-05
[Bug]: error in `instanceconfig.expandDevicesBlock()`


ngfeldman:
### Terraform Version

OpenTofu v1.8.0 on linux_amd64

### Linode Provider Version

2.25.0

### Effected Terraform Resources

linode_instance_config

### Terraform Config Files

```terraform
terraform {
  required_providers {
    linode = {
      source  = "linode/linode"
      version = "~> 2.25"
    }
  }
}

variable "linode_api_token" {
  type      = string
  sensitive = true
}

variable "linode_volume_id" {
  type = string
}

provider "linode" {
  token = var.linode_api_token
}

data "linode_user" "me" {
  username = "nfeldman"
}

data "linode_sshkeys" "ssh_keys" {
  filter {
    name   = "label"
    values = data.linode_user.me.ssh_keys
  }
}

locals {
  user_data = templatefile(
    "${path.module}/user-data.yaml.tftpl",
    { sshkeys = data.linode_sshkeys.ssh_keys.sshkeys },
  )
}

resource "linode_instance" "instance" {
  label            = "production-docker-host"
  image            = "linode/ubuntu24.04"
  region           = "us-west"
  type             = "g6-nanode-1"
  authorized_users = [data.linode_user.me.username]
  metadata {
    user_data = base64encode(local.user_data)
  }
}

resource "linode_instance_config" "instance_config" {
  linode_id = linode_instance.instance.id
  label     = "my-config"
  device {
    device_name = "linode-volume"
    volume_id   = var.linode_volume_id
  }
}

output "instance_ip_address" {
  value = linode_instance.instance.ip_address
}

output "user_data" {
  value = local.user_data
}
```

### Debug Output

```
Do you want to perform these actions?
  OpenTofu will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

linode_instance.instance: Creating...
linode_instance.instance: Still creating... [10s elapsed]
linode_instance.instance: Still creating... [20s elapsed]
linode_instance.instance: Still creating... [30s elapsed]
linode_instance.instance: Still creating... [40s elapsed]
linode_instance.instance: Creation complete after 50s [id=62258928]
linode_instance_config.instance_config: Creating...
╷
│ Error: Plugin did not respond
│ 
│   with linode_instance_config.instance_config,
│   on main.tf line 52, in resource "linode_instance_config" "instance_config":
│   52: resource "linode_instance_config" "instance_config" {
[...]

PR #1563, opened 2024-08-22
Add validation for the accepted values of `device_name` attribute of instance config resource

zliang-akamai:
## 📝 Description

Validate the input attribute value against accepted values before proceeding to processing it. This is necessary because the provider may panic when invalid input passed in.

Resolves https://github.com/linode/terraform-provider-linode/issues/1536 

## ✔️ How to Test
```bash
make int-test PKG_NAME="linode/instanceconfig" PARALLEL=5
```

Verifying an error will be reported when the invalid `device_name` passed in:

```hcl
  label            = "testinstance"
  image            = "linode/ubuntu24.04"
  region           = "us-mia"
  type             = "g6-nanode-1"
}

resource "linode_instance_config" "test" {
  label     = "my-config"
  device {
    device_name = "invalid-device-name"
    volume_id   = linode_volume.test.id
  }
}

resource "linode_volume" "test" {
    label = "foo-volume"
    region = linode_instance.test.region
}
```

### DIFF
--- a/linode/instanceconfig/schema_resource.go
+++ b/linode/instanceconfig/schema_resource.go
@@ -172,6 +172,15 @@ var deviceV2Schema = map[string]*schema.Schema{
 		Type:        schema.TypeString,
 		Required:    true,
 		Description: "The Disk ID to map to this disk slot",
+		ValidateDiagFunc: validation.ToDiagFunc(
+			validation.StringInSlice(
+				[]string{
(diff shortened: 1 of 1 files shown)


jriddle-linode (approved):
Looks good and works locally!

yec-akamai (approved):
Works well on my end!

PR #1564, opened 2024-08-22
Add `.terraform.lock.hcl` to `.gitignore`

zliang-akamai:
## 📝 Description

This is to add the `.terraform.lock.hcl` file to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ modules-dev/
 terraform-provider-linode
 
 dist/**
+
+.terraform.lock.hcl


ykim-1 (approved):

yec-akamai:
I'm wondering is this a necessary change? Normally we prefer to use sandbox environment to init TF, so this shouldn't be generated under the provider repo.

zliang-akamai:
> I'm wondering is this a necessary change? Normally we prefer to use sandbox environment to init TF, so this shouldn't be generated under the provider repo.

Thanks for letting me know lol. I haven't tried the sandbox so far, so let me try it.

PR #1574, opened 2024-09-09
Increase root pass length to 64 in tests

zliang-akamai:
## 📝 Description

- Some errors emerged saying password length requirement was not satisfied in my local testing.
  - I couldn't reproduce it... but the bottom line is that we can always increase security with a stronger password.

## ✔️ How to Test
(As usual, you might see unrelated and intermittent test failures)
```bash
make int-test PKG_NAME="linode/instance" PARALLEL=5
```

### DIFF
--- a/linode/backup/datasource_test.go
+++ b/linode/backup/datasource_test.go
@@ -34,7 +34,7 @@ func TestAccDataSourceInstanceBackups_basic(t *testing.T) {
 
 	resourceName := "data.linode_instance_backups.foobar"
 
-	rootPass := acctest.RandString(12)
+	rootPass := acctest.RandString(64)
 
 	var instance linodego.Instance
 	var snapshot *linodego.InstanceSnapshot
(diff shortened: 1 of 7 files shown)


ezilber-akamai (commented):
The test `TestAccResourceInstance_requestQuantity` failed locally for me with the error:

```
 resource_test.go:2134: Step 1/2 error: Error running pre-apply plan: exit status 1
        
        Error: Invalid resource type
        
          on terraform_plugin_test.tf line 46, in resource "linode_firewall" "e2e_test_firewall":
          46: resource "linode_firewall" "e2e_test_firewall" {
        
        The provider hashicorp/linode does not support resource type
        "linode_firewall".
```

This doesn't seem like an intermittent issue to me. Was this test passing on your machine?

zliang-akamai:
@ezilber-akamai Yes, the issue was that firewall resource has been migrated but that test case can only use the older SDKv2 testing provider. See https://github.com/linode/terraform-provider-linode/pull/1575 for details

ezilber-akamai:
> @ezilber-akamai Yes, the issue was that firewall resource has been migrated but that test case can only use the older SDKv2 testing provider. See #1575 for details

Thanks for clearing that up! That was the only failing test so everything looks good now!

ykim-1 (approved):
LGTM, nice work!

PR #1575, opened 2024-09-09
Skip `TestAccResourceInstance_requestQuantity`

zliang-akamai:
## 📝 Description

After firewall migration, [one instance test](https://github.com/linode/terraform-provider-linode/blob/b063cde4b91d6eead31c7ba446f25d6c71a8247b/linode/instance/resource_test.go#L2101-L2158) is failing because it's using [the older testing provider](https://github.com/linode/terraform-provider-linode/blob/b063cde4b91d6eead31c7ba446f25d6c71a8247b/linode/instance/resource_test.go#L2136) rather than [the providers factory](https://github.com/linode/terraform-provider-linode/blob/b063cde4b91d6eead31c7ba446f25d6c71a8247b/linode/instance/resource_test.go#L2176), and the firewall resource has been migrated and no longer available in the older provider.

I tried modifying the client after provider server creation in the factory, but it reported a nil pointer issue, probably meaning that the provider has not been initialized at that point.

```go
func GetProtoV5ProviderFactories(clientModifiers ...func(*linodego.Client)) map[string]func() (tfprotov5.ProviderServer, error) {
	return map[string]func() (tfprotov5.ProviderServer, error){
		"linode": func() (tfprotov5.ProviderServer, error) {
			ctx := context.Background()
			providers := []func() tfprotov5.ProviderServer{
				TestAccProviders["linode"].GRPCProvider,
				providerserver.NewProtocol5(
					TestAccFrameworkProvider,
				),
			}

			muxServer, err := tf5muxserver.NewMuxServer(ctx, providers...)
			if err != nil {
				return nil, err
			}

			server := muxServer.ProviderServer()

			for _, m := range clientModifiers {
				m(&TestAccProviders["linode"].Meta().(*helper.ProviderMeta).Client)
				m(TestAccFrameworkProvider.Meta.Client)
			}

			return server, nil
		},
	}
}
```

We might want to skip this test until we figuring out what's next.

## ✔️ How to Test

```bash
make int-test PKG_NAME="linode/instance" ARGS="-run TestAccResourceInstance_requestQuantity"
```


jriddle-linode (approved):
LGTM

PR #1578, opened 2024-09-10
Fix a NodeBalancer Configs Test Failure

zliang-akamai:
## 📝 Description

This is to fix one of the imtermitent test failures: https://github.com/linode/terraform-provider-linode/actions/runs/10795858504/job/29943481478#step:8:368

## ✔️ How to Test

```bash
make int-test PKG_NAME="linode/nbconfigs"
```

### DIFF
--- a/linode/nbconfigs/tmpl/data_basic.gotf
+++ b/linode/nbconfigs/tmpl/data_basic.gotf
@@ -3,6 +3,7 @@
 {{ template "nb_configs_data_base" . }}
 
 data "linode_nodebalancer_configs" "foo" {
+    depends_on = [ linode_nodebalancer_config.foo, linode_nodebalancer_config.bar ]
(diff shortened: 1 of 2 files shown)


linode/nbconfigs/tmpl/data_basic.gotf
```diff
@@ -3,6 +3,7 @@
 {{ template "nb_configs_data_base" . }}
 
 data "linode_nodebalancer_configs" "foo" {
+    depends_on = [ linode_nodebalancer_config.foo, linode_nodebalancer_config.bar ]
```
ykim-1, line 6: This will be a good addition to other tests that depends on a particular resource 👍

zliang-akamai: Thanks for the suggestion! I just added it to the data source filtering test case. Does it look good?

yec-akamai (approved):
Tested and worked well!

PR #1579, opened 2024-09-12
Add taints and labels support for embedded LKE pools

zliang-akamai:
## 📝 Description

This is to add taints and labels support to the embedded LKE node pools.

### Auto Testing

```bash
make int-test PKG_NAME="linode/lke" ARGS="-run TestAccResourceLKEClusterNodePoolTaintsLabels"
```

```bash
make int-test PKG_NAME="linode/lke" ARGS="-run TestAccDataSourceLKECluster_taints_labels"
```

### Manual Testing

```hcl
  label              = "my-cluster"
  k8s_version        = "1.30"
  region             = "us-mia"
  tags               = ["prod"]
  external_pool_tags = [local.external_pool_tag]

  pool {
    type  = "g6-standard-1"
    count = 1
  }
}
```


ykim-1 (approved):
LGTM, verified manually and tests work as expected

linode/lke/cluster.go
```diff
@@ -342,6 +359,20 @@ func matchPoolsWithSchema(pools []linodego.LKENodePool, declaredPools []interfac
 				continue
 			}
 
+			declaredTaints := expandNodePoolTaints(helper.ExpandObjectSet(declaredPool["taint"].(*schema.Set)))
+
+			// length comparison here is for handling the case of nil vs empty slice
+			if !reflect.DeepEqual(declaredTaints, apiPool.Taints) && len(declaredTaints) != len(apiPool.Taints) {
```
lgarber-akamai, line 365: Given order isn't guaranteed for taints, should we account for that in this comparison?

zliang-akamai: I think order of a list of object is guaranteed by the API? Otherwise we will have a lot more semantic equality checks to implement...

lgarber-akamai: Ahh okay, the piece I was missing is that the pool is also created using `(*schema.Set).List()`, and I also did confirm that the `taints` field is effectively treated as ordered in the API 👍 

The only case I can think of that might cause issues is during an import of an LKE cluster with the same taints in a different order, although I don't think this comparison would be reached since the import would pre-populated in state.

I still think ignoring order for this comparison would be good from a resiliency perspective, but I also think keeping it as-is is a valid approach 🙂

lgarber-akamai: One more thing:

I might be misunderstanding the logic here, but it seems like using `&&` in this comparison would cause any taints list same length to pass this check regardless of their contents.

zliang-akamai: Thanks @lgarber-akamai! I think these concerns has been addressed in the following up commits in this PR. Can you take a second look at it? Thanks!

lgarber-akamai (approved):
Thanks for addressing my feedback 🙏 
All tests are passing on my end, great work!

PR #1582, opened 2024-09-18
Move the workaround of removing resource when empty ID up

zliang-akamai:
## 📝 Description

Moving the workround for crossplane to the correct location.

## ✔️ How to Test

N/A


lgarber-akamai (approved):
Tests are all passing on my end, nice find!

ykim-1 (approved):

PR #1595, opened 2024-09-24
Framework Migration: `linode_nodebalancer_node` Resource

zliang-akamai:
## 📝 Description

This is to migrate `linode_nodebalancer_node` to be with Terraform Plugin Framework.

## ✔️ How to Test
```bash
make int-test PKG_NAME="linode/nbnode"
make int-test PKG_NAME="linode/nb"
make int-test PKG_NAME="linode/nbconfig"
```

```go
  count           = "2"
  label           = "web-${count.index + 1}"
  image           = "linode/ubuntu22.04"
  region          = "us-mia"
  type            = "g6-standard-1"
  authorized_keys = ["ssh-rsa AAAA...Gw== user@example.local"]
  root_pass       = "UNSAFE-PASSWORD!!"

  private_ip = true
}

  label                = "mynodebalancer"
  region               = "us-mia"
  client_conn_throttle = 20
}

resource "linode_nodebalancer_config" "foofig" {
  nodebalancer_id = linode_nodebalancer.foobar.id
  port            = 80
  protocol        = "http"
  check           = "http"
  check_path      = "/foo"
  check_attempts  = 3
  check_timeout   = 30
  stickiness      = "http_cookie"
  algorithm       = "source"
}

resource "linode_nodebalancer_node" "foonode" {
  count           = "2"
  nodebalancer_id = linode_nodebalancer.foobar.id
  config_id       = linode_nodebalancer_config.foofig.id
  address         = "${element(linode_instance.foo.*.private_ip_address, count.index)}:80"
  label           = "mynodebalancernode"
  weight          = 50

  lifecycle {
    // Tell Terraform to implicitly recreate the NodeBalancer node when
    // the target instance has been marked for recreation.
    // See: https://github.com/linode/terraform-provider-linode/issues/1224
    replace_triggered_by = [linode_instance.foo]
  }
}

```


linode/nbnode/framework_resource.go
```diff
@@ -0,0 +1,260 @@
...
+	})
+	node, err := client.CreateNodeBalancerNode(ctx, nodeBalancerID, configID, createOpts)
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to Create a Linode NodeBalancerNode", err.Error())
+		return
+	}
+
+	// Add resource to TF states earlier to prevent
+	// dangling resources (resources created but not managed by TF)
+	AddNodeResource(ctx, *node, resp, plan)
```
yec-akamai, line 68: Just curious, is preventing dangling resources in creation a new suggestion from Hashicorp? Looks like we didn't have this logic in other resources.

yec-akamai (commented):
Overall looks good, nice work!

linode/nbnode/framework_resource.go
zliang-akamai: Hi Ye, that's not a official suggestion, but I feel it would be safer to have the ID set into state as soon as they are available to prevent any possible errors returned before reaching `resp.State.Set(ctx, &plan)`.

I wrote a similar pattern in the disk resource: https://github.com/linode/terraform-provider-linode/blob/58fc6dee71f2d7a57712bf5f801866cf13e018af/linode/instancedisk/framework_resource.go#L124

jriddle-linode (approved):
LGTM and works locally.

yec-akamai (approved):

PR #1597, opened 2024-09-26
Terraform support for creation + assigning additional reserved IP to linode

AniJ98:
## 📝 Description

This PR introduces comprehensive functionality and tests for creating a linode with a reserved IP and adding additional reserved IP to a linode in the terraform provider. The changes are necessary to provide support for managing both of the aforementioned features through terraform. The changes include:

Implementation of the following operations:

- Creating a linode with a reserved IP address
- Adding additional reserved IP addresses to a linode

Test coverage:

- Resource test - Covers testing the functionality of creating a linode with a specific reserved IP address and assigning additional reserved IP address to an existing linode.

## ✔️ How to Test

1. Ensure you have a valid Linode API token. Set up your environment: 

      `export LINODE_TOKEN="your_token_here" `

2. To test creation of a linode with reserved IP address : 

    `make int-test PKG_NAME="linode/instance" ARGS="-run TestAccResourceInstance_withReservedIP"`

3. To test assigning additional reserved IP addresses to an existing linode:

      `make int-test PKG_NAME="linode/instancereservedipassignment" ARGS="-run TestAccInstanceIP_addReservedIP"`

4. Verify the test output for any failures or unexpected behavior.

Note:

Ensure you have proper permissions and sufficient quota in your Linode account to perform Reserved IP operations. Some tests may create and delete resources, so use a testing environment if possible.


linode/instanceip/framework_resource.go
```diff
@@ -54,13 +54,59 @@ func (r *Resource) Create(
...
 
 	client := r.Meta.Client
-	ip, err := client.AddInstanceIPAddress(ctx, linodeID, isPublic)
-	if err != nil {
-		resp.Diagnostics.AddError(
-			fmt.Sprintf("Failed to created instance (%d) IP", linodeID),
-			err.Error(),
-		)
-		return
+	var ip *linodego.InstanceIP
```
lgarber-akamai, line 57: Given reserved IPs use a separate set of endpoints (`networking/reserved/ips`), I wonder if it'd be worth creating a new `linode_instance_reserved_ip` resource entirely 🤔 

Given there is a lot of overlap, something like a `reserved` bool field might also suffice.

AniJ98: I already have another PR open for the new resource called `networkreservedips`. I kept these functionalities here to maintain consistency with the linodego library.

Yes I have added the reserved bool field in the model of the `networkreservedips` resource.

AniJ98: `networkreservedips` resource handles 4 other endpoints - Create, Fetch, List, Delete reserved IP addresses.

linode/instanceip/framework_resource.go
```diff
@@ -54,13 +54,59 @@ func (r *Resource) Create(
...
+	var err error
+
+	if !plan.Address.IsNull() && !plan.Address.IsUnknown() {
+		// Assign a reserved IP
+		createOpts := linodego.InstanceReserveIPOptions{
+			Type:    "ipv4",
+			Public:  isPublic,
+			Address: plan.Address.ValueString(),
+		}
+		_, err = client.AssignInstanceReservedIP(ctx, linodeID, createOpts)
```
lalenche, line 67: Please pass on the error to user in error message

linode/instanceip/framework_resource.go
```diff
@@ -54,13 +54,59 @@ func (r *Resource) Create(
...
+		if err != nil {
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Failed to assign reserved IP to instance (%d)", linodeID),
+				err.Error(),
+			)
+			return
+		}
+
+		// Fetch the IP information after assigning
```
lalenche, line 76: The code only needs to assign the IP, I think. Is there a need to read back and verify since you should be able to trust the return code of the first request?

linode/instanceip/framework_schema.go
```diff
@@ -41,6 +41,7 @@ var frameworkResourceSchema = schema.Schema{
 
 		"address": schema.StringAttribute{
 			Description: "The resulting IPv4 address.",
+			Optional:    true,
```
lalenche, line 44: Not sure why the address has become Optional.

lalenche: Also assuming 'reserved' will be added in a later PR? (If this is the response you get when listing instance IPs

linode/instance/resource.go
```diff
@@ -818,6 +825,17 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
...
+	if d.HasChange("ipv4") {
+		fmt.Println("Change detected")
+		oldIPv4, newIPv4 := d.GetChange("ipv4")
+		oldSet := oldIPv4.(*schema.Set)
+		newSet := newIPv4.(*schema.Set)
+		if len(newSet.Difference(oldSet).List()) > 0 {
+			// Log a warning that direct IP changes are not supported
+			return diag.Errorf("Changing IPv4 addresses after instance creation is not directly supported. The attempted change will be ignored.")
+		}
+	}
```
yec-akamai, line 837: It looks like the ipv4 addresses can't be updated, right? If so, I think we can simplify this part to be just returning the error when change detected: 
```suggestion
	if d.HasChange("ipv4") {
		// Log a warning that direct IP changes are not supported
		return diag.Errorf("Changing IPv4 addresses after instance creation is not directly supported. The attempted change will be ignored.")
	}
```

linode/instance/tmpl/template.go
```diff
@@ -732,3 +733,20 @@ func WithPG(t *testing.T, label, region, assignedGroup string, groups []string)
...
+			Label:    label,
+			PubKey:   pubKey,
+			Image:    acceptance.TestImageLatest,
+			Region:   region,
+			RootPass: rootPass,
+			IPv4:     []string{reservedIP},
+		})
+
+	// Add this debug logging
+	t.Logf("Generated Terraform config for WithReservedIP:\n%s", generatedConfig)
```
yec-akamai, line 749: We usually don't do debug logging for integration tests. Is there a specific reason that it's necessary?

linode/instanceip/framework_resource.go
yec-akamai: Since we have a separate resource to manage the reservedipds, I'm wondering if it might be better to not having them here to keep this instanceip resource staightforward.

yec-akamai (changes requested):

linode/instanceip/framework_resource.go
AniJ98: The reason I kept the functionality in the `instanceip` resource is because the linodego function to assign reserved IP to instances is within the `instance_ips.go` class of linodego. So I thought it would improve consistency. (The reason the function is in the `instance_ips.go` is because the endpoint matches the others in the class)

linode/instance/tmpl/template.go
AniJ98: Yeah I am going to go ahead and remove it

linode/instanceip/framework_schema.go
AniJ98: Yes the 'reserved' field is added in the other PR

linode/instanceip/framework_resource.go
AniJ98: Yeah this part of the code was just for confirmation. I will be removing it in the next commit.

linode/instance/resource.go
AniJ98: Yeah I have removed the part where I am storing the diff

linode/instanceip/framework_resource.go
lgarber-akamai: Sorry for the late response to this thread! 

I think I'm with @yec-akamai here, it might make more sense to exclude this logic from the `linode_instance_ip` resource since this resource specifically corresponds to a single non-reserved IP 🙂

linode/instancereservedip/framework_schema.go
```diff
@@ -0,0 +1,113 @@
...
+				boolplanmodifier.RequiresReplace(),
+			},
+		},
+
+		"address": schema.StringAttribute{
+			Description: "The resulting IPv4 address.",
+			Required:    true,
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.UseStateForUnknown(),
+			},
```
lgarber-akamai, line 47: Given this resource represents a reserved that isn't necessarily assigned to a Linode, would it make sense for it to be renamed to `linode_reserved_ip` with address & linode_ids fields being made computed? In that case existing reserved IPs could be imported using `terraform import`.

linode/instancereservedip/framework_schema.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+		"region": schema.StringAttribute{
+			Description: "The region this IP resides in.",
+			Computed:    true,
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.UseStateForUnknown(),
+			},
+		},
```
lgarber-akamai, line 74: Could this field be made optional given it's accepted when creating reserved IPs that aren't attached to a Linode?

linode/instancereservedip/framework_models.go
```diff
@@ -0,0 +1,114 @@
+package instancereservedip
```
lgarber-akamai, line 1: Sorry for all the confusion surrounding this PR! 

It looks like the changes in #1598 would make this resource implementation redundant so it can probably be dropped from this PR 👍

Correct me if I'm wrong, but I think the scope of this PR can probably be narrowed to just the changes to the `linode_instance` since we decided against making changes to `linode_instance_ip`.

AniJ98: No worries! The changes in #1598 handles the following operations of reserved IP -Create, Fetch, List. The changes in this PR is only restricted to creation of a linode with an existing reserved IP + assigning reserved IP to linodes. So I dont think there is an overlap here, but I maybe wrong.

linode/instanceip/framework_resource.go
```diff
@@ -153,7 +153,6 @@ func (r *Resource) Read(
 	if resp.Diagnostics.HasError() {
 		return
 	}
-
```
zliang-akamai, line 156: Can you restore this unrelated change?

linode/instanceip/tmpl/template.go
```diff
@@ -10,6 +10,7 @@ type TemplateData struct {
 	Label            string
 	ApplyImmediately bool
 	Region           string
+	Address          string
```
zliang-akamai, line 13: I don't think this is effective, right?

linode/instancereservedipassignment/framework_schema.go
```diff
@@ -0,0 +1,113 @@
...
+			Computed:    true,
+			PlanModifiers: []planmodifier.Int64{
+				int64planmodifier.UseStateForUnknown(),
+			},
+		},
+		"rdns": schema.StringAttribute{
+			Description: "The reverse DNS assigned to this address.",
+			Optional:    true,
+			Computed:    true,
+			PlanModifiers: []planmodifier.String{
```
yec-akamai, line 64: It looks like RDNS can be changed in Update. Should we not preserve its value from state if it leaves empty? User may intend to remove it.

yec-akamai (commented):
Thank you for addressing the comments. Just a few thing:

linode/instance/schema_resource.go
```diff
@@ -333,6 +333,7 @@ var resourceSchema = map[string]*schema.Schema{
 		Description: "This Linode's IPv4 Addresses. Each Linode is assigned a single public IPv4 address upon " +
 			"creation, and may get a single private IPv4 address if needed. You may need to open a support ticket " +
 			"to get additional IPv4 addresses.",
+		Optional: true,
```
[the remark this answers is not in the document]
AniJ98: Yes that makes sense. I will describe the when the ipv4 becomes an optional field

linode/instancereservedipassignment/framework_models.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/linode/linodego"
+
+	"github.com/linode/terraform-provider-linode/v2/linode/helper"
+	"github.com/linode/terraform-provider-linode/v2/linode/instancenetworking"
+)
+
+type IPVPCNAT1To1Model struct {
```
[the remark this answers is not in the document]
AniJ98: Thank you for catching that. I will remove it now

linode/instance/resource.go
```diff
@@ -821,6 +830,11 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
 		}
 	}
 
+	if d.HasChange("ipv4") {
+		// Log a warning that direct IP changes are not supported
+		return diag.Errorf("Changing IPv4 addresses after instance creation is not directly supported. The attempted change will be ignored.")
+	}
+
```
zliang-akamai, line 837: Would it be better to use [`ForceNew`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#forcenew) attribute to let TF recreated the resource when the field is updated?

linode/instancereservedipassignment/framework_resource.go
```diff
@@ -0,0 +1,293 @@
...
+) {
+	tflog.Debug(ctx, "Read linode_instance_reserved_ip")
+	var state InstanceIPModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if helper.FrameworkAttemptRemoveResourceForEmptyID(ctx, state.ID, resp) {
```
zliang-akamai, line 149:
```suggestion
	// TODO: cleanup when Crossplane fixes it
	if helper.FrameworkAttemptRemoveResourceForEmptyID(ctx, state.ID, resp) {
```

yec-akamai (commented):
Overall looks good now! Just one thing:

When I ran this test `TestAccResourceInstance_withReservedIP`, it failed with this error message: `no region found with the provided caps`. But I took a look at the test case, it uses a fixed region `us-east`. Are you aware of what's the issue could be?

linode/instance/tmpl/templates/instance_with_reserved_ip.gotf
```diff
@@ -0,0 +1,15 @@
...
+resource "linode_instance" "foobar" {
+    label  = "{{ .Label }}"
+    type   = "g6-nanode-1"
+    region = "{{ .Region }}"
+    image  = "{{ .Image }}"
+
+    root_pass = "{{ .RootPass }}"
+    authorized_keys = ["{{ .PubKey }}"]
+
+    ipv4 = [{{ range $index, $ip := .IPv4 }}{{ if $index }}, {{ end }}"{{ $ip }}"{{ end }}]
```
yec-akamai, line 12: Can you add the e2e test firewall to the instance for compliance:

```
    firewall_id = linode_firewall.e2e_test_firewall.id

```

linode/instance/resource_test.go
```diff
@@ -2893,3 +2893,37 @@ func checkComputeInstanceDisk(instance *linodego.Instance, label string, size in
...
+					resource.TestCheckResourceAttr(resourceName, "label", instanceName),
+					resource.TestCheckResourceAttr(resourceName, "ipv4.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "ipv4.0", reservedIP),
+				),
+			},
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"root_pass", "authorized_keys", "image", "migration_type", "resize_disk"},
```
ykim-akamai, line 2925:
```
ImportStateVerifyIgnore: []string{"root_pass", "authorized_keys", "image", "migration_type", "resize_disk", "firewall_id"},
```
To bypass firewall assertion

linode/instancereservedipassignment/framework_resource.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/linode/terraform-provider-linode/v2/linode/helper"
+)
+
+// For internal only: please refer to KB page for more information
+
+func NewResource() resource.Resource {
+	return &Resource{
+		BaseResource: helper.NewBaseResource(
+			helper.BaseResourceConfig{
+				Name:   "linode_reserved_ip_assign",
```
lgarber-akamai, line 21: Could this resource be renamed to `linode_reserved_ip_assignment` to maintain declarative naming?

AniJ98: I think it makes sense to keep the name "linode_reserves_ip_assign" since the endpoint also has "assign" in it. I could change the name of the folder from "instancereservedipassignment" to "instancereservedipassign"

AniJ98: On second thoughts, I agree that linode_reserved_ip_assignment makes more sense for our application. I will making that change and pushing it now

linode/instance/tmpl/templates/instance_with_reserved_ip.gotf
AniJ98: I have now added the firewall to the the linode_instance for compliance reasons. Thanks!

linode/instance/resource.go
AniJ98: Given that the update function is designed to modify existing resources, when a customer attempts to update a resource by adding an IPv4 field, it's more likely they intend to assign a reserved IPv4 address to the existing resource rather than create a new Linode instance with a reserved IP address. However, since this specific operation is not currently supported, we've decided to address this by displaying a warning message and preventing the operation from proceeding. This approach ensures that customers are informed about the limitation

linode/instancereservedipassignment/framework_models.go
```diff
@@ -0,0 +1,108 @@
...
+	RDNS             types.String `tfsdk:"rdns"`
+	Region           types.String `tfsdk:"region"`
+	SubnetMask       types.String `tfsdk:"subnet_mask"`
+	Type             types.String `tfsdk:"type"`
+	ApplyImmediately types.Bool   `tfsdk:"apply_immediately"`
+	IPVPCNAT1To1     types.List   `tfsdk:"vpc_nat_1_1"`
+	Reserved         types.Bool   `tfsdk:"reserved"`
+}
+
+func (m *InstanceIPModel) FlattenInstanceIP(
```
yec-akamai, line 31:
```suggestion
func (m *InstanceIPModel) flattenInstanceIP(
```
This function looks like only used within this package.

linode/instancereservedipassignment/framework_resource.go
```diff
@@ -0,0 +1,294 @@
...
+	linodeID := helper.FrameworkSafeInt64ToInt(
+		state.LinodeID.ValueInt64(),
+		&resp.Diagnostics,
+	)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	tflog.Debug(ctx, "client.DeleteInstanceIPAddress(...)")
+	if err := client.DeleteInstanceIPAddress(ctx, linodeID, address); err != nil {
```
yec-akamai, line 276: I'm a bit hesitate on the design of the Delete(...) function here. It looks like we deleted this ip address instead of unassign it from a linode, right? By using this assignment resource, I'll assume user expect to unassign an IP instead of delete it. Do we have an unassign reserved ip endpoint? If not, I'd prefer to not implement it here.

lalenche: DELETE https://api.linode.com/{apiVersion}/linode/instances/{linodeId}/ips/{address} is a valid way to delete an additional address on a Linode

[API documentation](https://techdocs.akamai.com/linode-api/reference/delete-linode-ip
)
[Our documentation on it](https://bits.linode.com/ReservedIP/tests/blob/main/how-to.md#7-how-to-unassign-an-additional-reserved-address-to-a-linode)

Is there another API that you are thinking of that may also do the same operation?

lgarber-akamai: @yec-akamai In practice it looks like `client.DeleteInstanceIPAddress(...)` unassigns reserved IPs from the target instance without actually deleting the reserved IP, but I can't find any documentation to back this up.

linode/instancereservedipassignment/framework_models.go
yec-akamai: You'll have to change them in the usage as well :)

linode/instancereservedipassignment/framework_resource.go
lalenche: Yeah, sadly no documentation exists for the reserved IP bit of it because it is not a publicly released feature, and our API spec itself does not mention it because we did not need to do any code change for it. So the API works like it always used to, and does not do anything about reserved IP feature.

linode/instance/resource.go
AniJ98: I agree. I will make the ipv4 attribute a ForceNew one so that a new linode will be created any time the ipv4 field is updated.

lgarber-akamai (approved):
Looks great and all tests are passing locally, thanks for the contribution!

As discussed privately, the tests will be updated and re-enabled as a part of #1598.

linode/instancereservedipassignment/resource_test.go
```diff
@@ -0,0 +1,58 @@
...
+const testInstanceIPResName = "linode_reserved_ip_assignment.test"
+
+var testRegion string
+
+func init() {
+	region, err := acceptance.GetRandomRegionWithCaps(nil, "core")
+	if err != nil {
+		log.Fatal(err)
+	}
+	region := "us-east"
```
yec-akamai, line 25: There is no new variables on left side of := you will have to just override the value:
```suggestion
	region = "us-east"
```

PR #1605, opened 2024-10-04
Fix Firewall Update

zliang-akamai:
## 📝 Description

- Correct a checking logic during update
- Add `UseStateForUnknown` to prevent unexpected removals of Linode and NodeBalancers from a firewall.

### Automated Testing
```bash
make int-test PKG_NAME="linode/firewall"
```

### Manual Testing
Verify any rule change won't detach the Linode from the firewall.

```hcl
resource "linode_firewall_device" "my_device" {
  firewall_id = linode_firewall.my_firewall.id
  entity_id = linode_instance.my_instance.id
}

resource "linode_firewall" "my_firewall" {
  label = "my_firewall"

  inbound {
    label    = "http"
    action = "DROP"
    protocol  = "TCP"
    ports     = "80"
    ipv4 = ["0.0.0.0/0"]
    ipv6 = ["::/0"]
  }

  inbound_policy = "DROP"
  outbound_policy = "ACCEPT"
}

resource "linode_instance" "my_instance" {
  label      = "my_instance"
  region     = "us-southeast"
  type       = "g6-standard-1"
}

```

### DIFF
--- a/linode/firewall/framework_models.go
+++ b/linode/firewall/framework_models.go
@@ -356,7 +356,7 @@ func (state *FirewallResourceModel) RulesAndPoliciesHaveChanges(
 func (state *FirewallResourceModel) LinodesOrNodeBalancersHaveChanges(
 	ctx context.Context, plan FirewallResourceModel,
 ) bool {
-	return !state.Linodes.Equal(plan.Linodes) || state.NodeBalancers.Equal(plan.NodeBalancers)
+	return !state.Linodes.Equal(plan.Linodes) || !state.NodeBalancers.Equal(plan.NodeBalancers)
 }
 
 func FlattenFirewallRules(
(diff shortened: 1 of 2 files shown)


lgarber-akamai:
@zliang-akamai Thanks for the quick turnaround on this! Could this PR be repointed to `main` so we can release it in a patch?

lgarber-akamai (approved):
I was able to reproduce the issue using the latest commit on `dev` and manually validated that it is resolved by this PR. Great work!

zliang-akamai:
@lgarber-akamai the base has been changed to `main`, thanks for the review!

lgarber-akamai:
Looks like the integration tests are required to merge to main but aren't run by default because this isn't the `dev` branch 👀

PR #1617, opened 2024-10-11
test: Add Cloud Firewall to LKE nodes after integration test execution

ykim-1:
## 📝 Description

Currently there isn't a CFW being applied to LKE nodes in integration test runs. Attaching FW to LKE cluster upon creation is currently not available which is the reason why only Calico rules were only being applied. This PR fixes that problem by going through the test account after test execution and attaching a default firewall to all remaining instances

## ✔️ How to Test

Tested on my forked - https://github.com/ykim-1/terraform-provider-linode/actions/runs/11351444139/job/31572142776 

## 📷 Preview

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -107,8 +107,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: [integration_tests]
     steps:
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
       - name: Install Linode CLI
-        run: pip install linode-cli
+        run: |
+          pip install linode-cli
 
       - name: Create Firewall and Attach to Instances
         run: |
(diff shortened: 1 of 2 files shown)


ezilber-akamai (approved):
Great addition. Nice work!

.github/workflows/integration_tests.yml
```diff
@@ -77,6 +77,60 @@ jobs:
...
+          mv calicoctl-linux-amd64 /usr/local/bin/calicoctl
+          mv kubectl /usr/local/bin/kubectl
+
+      - name: Apply Calico Rules to LKE # Only Running against Matrix USER 4 which includes LKE test suite
+        run: |
+          cd e2e_scripts/cloud_security_scripts/lke_calico_rules/ && ./lke_calico_rules_e2e.sh
+        env:
+          LINODE_TOKEN: ${{ secrets.LINODE_TOKEN_USER_4 }}
+
+  add-fw-lke-nodes:
```
lgarber-akamai, line 106: Would it make sense to add an `if: always()` here to also run this for failed integration test runs?

ykim-1: Yeah good catch

lgarber-akamai (approved):
Looks great, nice work!

PR #1624, opened 2024-10-17
fix: Instance disk nil pointer issue

yec-akamai:
## 📝 Description

When we try to refresh the latest disk status, it may fail and cause the refreshed disk id to be nil. Use a local variable to store the disk id value and prevent the nil pointer issue. 

## ✔️ How to Test

```
make PKG_NAME=linode/instancedisk int-test
```

### DIFF
--- a/linode/instancedisk/framework_resource.go
+++ b/linode/instancedisk/framework_resource.go
@@ -119,34 +119,35 @@ func (r *Resource) Create(
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ezilber-akamai (approved):
Thanks for fixing this! Tests are passing locally.

zliang-akamai (approved):
Thanks for fixing it!

PR #1639, opened 2024-10-31
Support upgrade attribute for DC migration

zliang-akamai:
## 📝 Description

This is to support `upgrade` attribute for DC migration for Linode instances.

### Automated Testing

```bash
make int-test PKG_NAME="linode/instance"
```

It's hard to simulate upgrading in integration testing because we don't have a newer type than g6.

### Manual Testing
I couldn't find a way to test it because I couldn't create a old Linode plan with a pending upgrade. Any suggestion?


go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/hashicorp/terraform-plugin-mux v0.17.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.34.0
 	github.com/hashicorp/terraform-plugin-testing v1.10.0
-	github.com/linode/linodego v1.42.0
+	github.com/linode/linodego v1.42.1-0.20241031190843-e69c3138ddcf
```
zliang-akamai, line 29: We will need to wait for next linodego release for this PR

PR #1658, opened 2024-11-12
Added support for updated behavior of linode API's (networkingip, rdns, instance)

AniJ98:
## 📝 Description

This PR implements functionality and tests for managing Reserved IP addresses in the Linodego client. These changes align with the updated Linode API, providing comprehensive support for Reserved IP operations. The changes include:

Implementation of core operations:

- Deleting a Linode while retaining the Reserved IP
- Retrieving IP information for Linodes, including Reserved IPs
- Listing all IP addresses, including Reserved IPs
- Allocating and assigning new Reserved IP addresses
- Assigning Reserved IPs to Linodes
- Fetching particular IP address, including Reserved IPs
- Converting between Ephemeral and Reserved IPs

Test Coverage:

- TestAccResourceInstance_deleteWithReservedIP - Verifies that deleting an instance created with a reserved IP does not delete the reserved IP.
- TestAccDataSourceInstanceNetworking_basicwithReseved - Confirms that the IP addresses of a Linode are returned with the updated reserved status.
- TestAccDataSourceNetworkingIP_list, TestAccDataSourceNetworkingIP_filterReserved - Tests the listing of IP addresses and the ability to filter them based on their reserved status.
- TestAccResourceNetworkingIP_reserved - Evaluates the allocation of reserved and ephemeral IP addresses to Linodes, as well as the creation of both types of IP addresses.
- TestAccResourceNetworkingIPsAssign - Tests the assignment of multiple reserved and ephemeral IP addresses to Linodes within a specific region.
- TestAccResourceNetworkingIP_reserveToUnreserve - Verifies the conversion of reserved IP addresses into ephemeral IP addresses
- TestAccDataSourceNetworkingIP_basic - Tests fetching a specific IP address, including its updated reserved status.

## ✔️ How to Test

1. Ensure you have a valid Linode API token. Set up your environment: 

   `export LINODE_TOKEN="your_token_here"`

2. To test the above mentioned functionalities individually : 

     `make int-test PKG_NAME="linode/networkingip" ARGS="-run TestAccResourceNetworkingIP_reserved"`

3. Verify the test output for any failures or unexpected behavior.

Note:

Ensure you have proper permissions and sufficient quota in your Linode account to perform Reserved IP operations. Some tests may create and delete resources, so use a testing environment if possible.


linode/rdns/framework_resource.go
```diff
@@ -39,236 +31,126 @@ type Resource struct {
...
-
-	ip, err := client.GetIPAddress(ctx, address)
-	if err != nil {
-		resp.Diagnostics.AddError(
-			"Failed to get the ip address associated with this RDNS",
-			err.Error(),
-		)
-		return
+	if !plan.Reserved.IsNull() {
+		reserved := plan.Reserved.ValueBool()
```
lalenche, line 51: As we discussed offline, it does not make sense to make `reserved` a property of the `linode_rdns` resource, because it is not related to RDNS. `reserved` is a property of an IP address and show maybe we need to create a new resource that let's user reserve or unreserve an address assigned to a Linode.

linode/instance/resource.go
```diff
@@ -830,6 +830,11 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta interface{
 		}
 	}
 
+	if d.HasChange("ipv4") {
+		// Log a warning that direct IP changes are not supported
+		return diag.Errorf("Changing IPv4 addresses after instance creation is not directly supported. The attempted change will be ignored.")
+	}
+
```
lgarber-akamai, line 837: Could this be removed since the `ipv4` field was made `ForceNew` in a previous PR?

linode/instanceip/tmpl/template.go
```diff
@@ -10,6 +10,7 @@ type TemplateData struct {
 	Label            string
 	ApplyImmediately bool
 	Region           string
+	Address          string
```
lgarber-akamai, line 13: It doesn't look like this field is referenced in any of the templates; would it make sense to drop it from this PR?

linode/networkipassignment/framework_resource.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/linode/linodego"
+	"github.com/linode/terraform-provider-linode/v2/linode/helper"
+)
+
+func NewResource() resource.Resource {
+	return &Resource{
+		BaseResource: helper.NewBaseResource(
+			helper.BaseResourceConfig{
+				Name:   "linode_networking_assign_ip",
```
lgarber-akamai, line 18: Could this resource be renamed to `linode_networking_ip_assignment` to match the declarative naming scheme?

linode/networkingip/framework_resource.go
```diff
@@ -0,0 +1,173 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/linode/linodego"
+	"github.com/linode/terraform-provider-linode/v2/linode/helper"
+)
+
+func NewResource() resource.Resource {
```
lgarber-akamai, line 14: Could pages in the `docs` directory be added for the new `linode_networking_ip`, `data.linode_networking_ips`, and `linode_network_ip_assignment` resources?

linode/rdns/framework_resource_schema.go
```diff
@@ -23,21 +23,24 @@ var frameworkResourceSchema = schema.Schema{
...
-			Required: true,
+			Optional: true,
 			Validators: []validator.String{
 				stringvalidator.LengthBetween(3, 254),
 			},
 		},
+		"reserved": schema.BoolAttribute{
+			Description: "Whether the IP address is reserved.",
+			Optional:    true,
+		},
```
lgarber-akamai, line 34: I've given this change a bit of thought and I think it might be worth excluding the `reserved` field from the `linode_rdns` resource since it seems a tiny bit out of scope and can be managed directly through the IP resources. 

I'd be interested to hear everyone's input on this!

linode/networkipassignment/framework_resource.go
```diff
@@ -0,0 +1,154 @@
+package networkipassignment
```
lgarber-akamai, line 1: Could this package be renamed to `networkingipassignment` to be consistent with the other package names?

linode/rdns/framework_resource.go
lgarber-akamai: > As we discussed offline, it does not make sense to make reserved a property of the linode_rdns resource

Agreed, I think it might make more sense for an IP's ephemeral vs reserved status to only be manageable through the IP resource(s) to reduce usage complexity. I'd definitely be interested to hear everyone else's thoughts though!

linode/instance/resource.go
AniJ98: Yeah makes sense! I will be removing it in the next commit

linode/instanceip/tmpl/template.go
AniJ98: Yes I just confirmed it can be dropped. I will remove it from the template file.

linode/networkipassignment/framework_resource.go
AniJ98: Yeah I agree that makes more sense. I will change the resource name too.

linode/framework_provider.go
```diff
@@ -54,6 +54,8 @@ import (
 	"github.com/linode/terraform-provider-linode/v2/linode/nbs"
 	"github.com/linode/terraform-provider-linode/v2/linode/nbtypes"
 	"github.com/linode/terraform-provider-linode/v2/linode/networkingip"
+	"github.com/linode/terraform-provider-linode/v2/linode/networkingips"
+	"github.com/linode/terraform-provider-linode/v2/linode/networkipassignment"
```
lgarber-akamai, line 58:
```suggestion
"github.com/linode/terraform-provider-linode/v2/linode/networkingipassignment"
```

linode/rdns/resource_test.go
```diff
@@ -206,6 +206,32 @@ func TestAccResourceRDNS_waitForAvailableWithTimeout(t *testing.T) {
...
+	resName := "linode_rdns.foobar"
+	linodeLabel := acctest.RandomWithPrefix("tf_test")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acceptance.PreCheck(t) },
+		ProtoV5ProviderFactories: acceptance.ProtoV5ProviderFactories,
+
+		Steps: []resource.TestStep{
+			{
+				Config: tmpl.UnreservedToReserved(t, linodeLabel, testRegion, true),
```
lgarber-akamai, line 221: This fixture seems to be missing 👀

linode/framework_provider.go
```diff
@@ -235,6 +237,8 @@ func (p *FrameworkProvider) Resources(ctx context.Context) []func() resource.Res
 		volume.NewResource,
 		vpc.NewResource,
 		vpcsubnet.NewResource,
+		networkingip.NewResource,
+		networkipassignment.NewResource,
```
lgarber-akamai, line 241:
```suggestion
		networkingipassignment.NewResource,
```

linode/networkingipassignment/resource_test.go
```diff
@@ -0,0 +1,108 @@
...
+		log.Fatal(err)
+	}
+
+	testRegion = region
+}
+
+func TestAccResourceNetworkingIPsAssign(t *testing.T) {
+	t.Parallel()
+
+	resourceName := "linode_networking_assign_ip.test"
```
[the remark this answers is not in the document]
AniJ98: Made this change now! Thanks!

linode/rdns/resource_test.go
AniJ98: This fixture doesnt belong in this resource. It has now been moved to linode_networking_ip resource.

linode/networkingipassignment/framework_resource_schema.go
```diff
@@ -0,0 +1,29 @@
...
+			ElementType: types.ObjectType{
+				AttrTypes: map[string]attr.Type{
+					"address":   types.StringType,
+					"linode_id": types.Int64Type,
+				},
+			},
+			Optional: true,
+		},
+	},
+}
```
lgarber-akamai, line 29: Same as https://github.com/linode/terraform-provider-linode/pull/1658/files#r1878492658

linode/networkingip/framework_resource_schema.go
```diff
@@ -0,0 +1,49 @@
...
+			Description: "The allocated IPv4 address.",
+			Computed:    true,
+			Optional:    true,
+		},
+		"type": schema.StringAttribute{
+			Description: "The type of IP address (ipv4).",
+			Computed:    true,
+			Optional:    true,
+		},
+	},
```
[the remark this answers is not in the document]
AniJ98: Yes I agree. I will be making the change in the next commit.

linode/networkingipassignment/framework_resource_schema.go
AniJ98: Made similar changes to this resource too

lgarber-akamai:
@AniJ98 Just a heads up, I'll remove the do-not-merge label once the linodego changes for this have been released 👍

linode/networkingip/framework_resource_schema.go
```diff
@@ -0,0 +1,68 @@
...
+		"address": schema.StringAttribute{
+			Description: "The allocated IPv4 address.",
+			Computed:    true,
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.UseStateForUnknown(),
+			},
+		},
+		"type": schema.StringAttribute{
+			Description: "The type of IP address (ipv4).",
+			Optional:    true,
```
lgarber-akamai, line 61: Could this be made either required or be given a default value of IPv4?

linode/networkingip/framework_resource.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			if lErr, ok := err.(*linodego.Error); (ok && lErr.Code != 404) || !ok {
+				resp.Diagnostics.AddError(
+					"Failed to Delete Reserved IP",
+					fmt.Sprintf(
+						"failed to delete reserved ip (%s): %s",
+						state.Address.ValueString(), err.Error(),
+					),
+				)
+			}
```
lgarber-akamai, line 175: Just a heads up, I noticed a 404 is suppressed here when using `v4` instead of `v4beta`. The 404 checking logic is necessary so there's not much we can do to remedy this without manually checking the API version on the client.

lgarber-akamai: Weirdly enough tainting a reserved IP seems to not properly destroy the IP even when the DELETE endpoint returns a 200 🤔

lgarber-akamai: Ohh I see, deleting a reserved IP seems to just convert it to ephemeral. In that case I think this logic should:

1. Always call `client.DeleteInstanceIPAddress(...)`
2. Only call `client.DeleteReservedIPAddress(...)` if the IP is a reserved IP

linode/networkingipassignment/framework_resource_schema.go
```diff
@@ -0,0 +1,38 @@
...
+			Description: "The ID of the IP assignment operation.",
+			PlanModifiers: []planmodifier.String{
+				stringplanmodifier.UseStateForUnknown(), // Use the state when ID is unknown.
+			},
+		},
+		"region": schema.StringAttribute{
+			Required:    true,
+			Description: "The region for the IP assignments.",
+		},
+		"assignments": schema.ListAttribute{
```
lgarber-akamai, line 25: Could we use a SetAttribute here to make arbitrary order changes not cause a diff?

lgarber-akamai, line 25: I think it might make more sense to make one instance of this resource represent one assignment, similar to the [linode_placement_group_assignment](https://github.com/linode/terraform-provider-linode/blob/dev/linode/placementgroupassignment/framework_resource_schema.go) resource. In that case we would be able to use RequiresReplace, which would eliminate the need for update logic and would prevent the resource's ID from changing during updates, which seems to be disallowed by TF in my testing.

linode/networkingip/framework_resource_schema.go
AniJ98: Maybe not required, but having a default value would definitely be useful to this resource

linode/networkingip/framework_resource.go
AniJ98: I agree. This was the conditional delete that we discussed earlier in the huddle. I will make the change.

linode/networkingipassignment/framework_resource_schema.go
AniJ98: I think having this resource support multiple assignments makes sense because the linode API for it works that way. I was wondering if we could just disallow a user from making any updates to this resource by throwing an error.

linode/networkingipassignment/framework_resource.go
```diff
@@ -0,0 +1,154 @@
...
+func (r *Resource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	tflog.Debug(ctx, "Delete linode_networking_assign_ip")
+	var state NetworkingIPModel
+
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// No need to do anything for delete as the IPs will be automatically unassigned when the Linodes are deleted
```
lgarber-akamai, line 153: Given this resource represents the relationship between a bunch of instances and their IPs, could we unassign all IP assignments that exist in the state during the deletion?

linode/networkingip/framework_resource.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	client := r.Meta.Client
+
+	ip, err := client.GetIPAddress(ctx, state.ID.ValueString())
+	if err != nil {
+		if lerr, ok := err.(*linodego.Error); ok && lerr.Code == 404 {
+			resp.State.RemoveResource(ctx)
+			return
+		}
```
lgarber-akamai, line 85: It looks like private ephemeral IPs are always hitting this condition (404) on post-creation reads, leading to the resource being recreated on the next apply. Do they need to be accessed through a different endpoint?

AniJ98: That is true. The GetIPAddress() method only works with public IP Addresses. The resource also only supports public IP's for now. So I guess it is ok to leave it as is for now?

linode/networkingipassignment/framework_resource.go
AniJ98: Yes that makes more sense. I will be making that changes now.

linode/networkingipassignment/framework_resource.go
```diff
@@ -0,0 +1,154 @@
...
+			"Error updating IP Assignments",
+			fmt.Sprintf("Could not update IP assignments: %s", err),
+		)
+		return
+	}
+
+	// Update the ID to reflect any changes in the number of assignments
+	plan.ID = types.StringValue(fmt.Sprintf("%s-%d", plan.Region.ValueString(), len(plan.Assignments)))
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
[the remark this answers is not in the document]
AniJ98: Yeah the implementation of the update is no longer required. I will be make the change now.

linode/networkingip/framework_resource.go
AniJ98: The behavior is correct. As confirmed in this [Slack thread](https://linode.slack.com/archives/CDXF1608G/p1734362729544919), there is no API available to fetch private IP addresses. To address this issue, we have two possible workarounds:

1. Use the /networking/ips API to list all addresses and manually filter out the specific private IP address.
2. Log a warning to indicate that the resource currently only supports public IPs, with the possibility of adding support for private IPs in the future.

linode/networkingip/framework_datasource.go
```diff
@@ -74,10 +70,29 @@ func (d *DataSource) Read(
...
-			"Unable to get IP Address: %s", err.Error(),
+			"Unable to list IP Addresses",
+			err.Error(),
+		)
+		return
+	}
+
+	var ip *linodego.InstanceIP
+	for _, candidate := range ips {
+		if candidate.Address == data.Address.ValueString() {
```
lalenche, line 85: It is possible that the same IP address is present but from different regions for a private address. Can the region is be checked as well?

AniJ98: We could additionally filter the results of the ListIPAddresses() based on the region. Will be making the change!

lalenche: FYI, the region is needed to disambiguate only if it is an RFC1918 address and there is a duplicate

ezilber-akamai:
> Tests are passing locally, but `TestAccResourceRDNS_unreservedToReserved` seems to be missing. Is this a test case you meant to include?

Fixed

ezilber-akamai (approved):
Looks good and tests are passing locally. Thanks for the contribution!

ykim-akamai (approved):
LGTM. Tested this pretty thoroughly. Nice work and thanks for your contribution!

Can you merge after fixing test highlighted in comment below?

ykim-akamai:
Here is integration test run on this PR - https://github.com/linode/terraform-provider-linode/actions/workflows/integration_tests_pr.yml

PR #1659, opened 2024-11-13
Allow white space characters in the label of an SSH key

zliang-akamai:
## 📝 Description

This is to allow white spaces in the label of an SSH key resource, as it's allowed by the API.

## ✔️ How to Test

```bash
make int-test PKG_NAME="linode/sshkey"
```


ykim-akamai (approved):
LGTM, tests pass locally

Issue #1618, opened 2024-10-15
Broken links in docs


thoward:
The page published at https://registry.terraform.io/providers/linode/linode/latest/docs has a number of broken links to https://www.terraform.io/docs/resources/object_storage_bucket.md and https://www.terraform.io/docs/resources/object_storage_object.md which should be linked to https://registry.terraform.io/providers/linode/linode/latest/docs/resources/object_storage_bucket and https://registry.terraform.io/providers/linode/linode/latest/docs/resources/object_storage_object respectively.

thoward:
Example of markdown source code where link is broken: https://github.com/linode/terraform-provider-linode/blob/dev/docs/index.md?plain=1#L87

I was going to try to edit this and submit a PR, but it's not clear how the docs publishing process works from this repo to Terraform's registry WRT to relative linking. I *think* you could just make the target a relative link to `index.md` e.g. `resources/object_storage_bucket.md` rather than `/docs/resources/object_storage_bucket.md`.

yec-akamai:
Fix in: https://github.com/linode/terraform-provider-linode/pull/1672

PR #1672, opened 2024-11-20
fix: Broken documentation links

ezilber-akamai:
## 📝 Description

Fixed broken docs links reported in #1618


lgarber-akamai (approved):
Looks great, thanks for fixing this!

zliang-akamai (approved):

Issue #628, opened 2022-04-29
[Bug]: Terraform linode_lke_cluster times out too soon, after 30 secs, and leaves orphaned nodes


blue928:
### Terraform Version

1.1.9

### Linode Provider Version

1.23.0

### Effected Terraform Resources

linode_lke_cluster
### Terraform Config Files

The following code follows this Linode tutorial https://www.linode.com/docs/guides/how-to-deploy-an-lke-cluster-using-terraform/
```
terraform {
  required_providers {
    linode = {
      source  = "linode/linode"
    }
    kubernetes = {
      source = "hashicorp/kubernetes"
    }
    helm = {
      source = "hashicorp/helm"
    }
    kubectl = {
      source = "gavinbunney/kubectl"
      #version = ">= 1.7.0"
    }
  }
}

# Configure the Linode Provider
provider "linode" {
  # token = "..."
  token = var.linode_token
}

resource "linode_lke_cluster" "peggyburkosky_prod" {
    label       = var.label
    k8s_version = var.k8s_version
    region      = var.region
    tags        = var.tags

    pool {
        type  = var.pool_type
        count = var.pool_count
    }
}

variable "linode_token" {
  default = ""
  sensitive = true
}

variable "label" {
  default = "peggyburkosky"
  type = string
}

variable "k8s_version" {
  default = "1.23"
  type = string 
}

variable "region" {
  default = "us-east"
  type = string
}

variable "tags" {
  default = ["prod"]
  type = list
}

variable "pool_type" {
  default = "g6-standard-1"
  type = string
}

variable "pool_count" {
  default = 1
  type = number
}

### Debug Output

_No response_

### Panic Output

_No response_

### Expected Behavior

The expected behavior is that the resource `linode_lke_cluster` will wait until the resource is configured before terminating. As well, expected behavior is that there should be a resource argument to specify waiting or at least stipulating a reasonable time period for cluster creation to complete. (30 secs is way too short.) Also, I'm not sure what the "label" argument is supposed to do, but it does not name the orphaned resource correctly. Instead, it's a random uuid generated by whereas the resource generated should be the label name (or we should be able to specify it regardless).

### Actual Behavior

[...]

PR #1673, opened 2024-11-21
doc: Remove Image Services Gen2 LA notice #628

yec-akamai:
## 📝 Description

Image Services Gen2 goes generally available 🎉


lgarber-akamai (approved):
🎉

jriddle-linode (approved):

PR #1678, opened 2024-11-27
Migrate obj to the framework

zliang-akamai:
## 📝 Description

This is to migrate the obj resource to be with TF framework.

## ✔️ How to Test
```bash
make int-test PKG_NAME="linode/obj" 
```


ykim-akamai (approved):
Quiet an update. All tests are passing locally, Nice work!

jriddle-linode (approved):
Looks good and everything is passing locally

PR #1679, opened 2024-11-27
Misc Fixes and Improvements

zliang-akamai:
## 📝 Description

- Use `r.Config.Name` as the SSoT for resource name in log in `linode_nodebalancer` resource.
- Use a simplified time comparison function in the schema of `linode_token` resource.
- Fix `AddNodeResource` function for `linode_nodebalancer_node` resource to add all required attributes for a `Read` function call (during `terraform refresh`).
- Fix an error message in `linode_nodebalancer_node` resource code.

## ✔️ How to Test
```bash
make int-test PKG_NAME="linode/token"
```

```bash
make int-test PKG_NAME="linode/nbnode"
```

```bash
make int-test PKG_NAME="linode/nb"
```

### DIFF
--- a/linode/nb/framework_resource.go
+++ b/linode/nb/framework_resource.go
@@ -38,7 +38,7 @@ func (r *Resource) Create(
 	req resource.CreateRequest,
 	resp *resource.CreateResponse,
 ) {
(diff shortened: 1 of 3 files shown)


yec-akamai (approved):
Thank you for the improvement!

PR #1688, opened 2024-12-06
Added support for migrations field for Placement Group Datasource

ezilber-akamai:
## 📝 Description

Added support for migrations field for Placement Group Datasource.

## ✔️ How to Test

`make PKG_NAME=linode/placementgroup ARGS="-run TestAccDataSourcePlacementGroup_basic" int-test`

Since there are no active migrations in the test, the field should not appear in the returned data.


linode/placementgroup/framework_datasource_schema.go
```diff
@@ -47,5 +55,22 @@ var DataSourceSchema = schema.Schema{
...
+						ElementType: pgMigrationInstanceObjectType,
+					},
+					"outbound": schema.ListAttribute{
+						Description: "The compute instances being migrated out of the placement group.",
+						Computed:    true,
+						ElementType: pgMigrationInstanceObjectType,
+					},
+				},
+			},
+		},
```
zliang-akamai, line 74: Since it not actually a list, can we use `SingleNestedBlock` rather than `ListNestedBlock` for it?

ezilber-akamai: Good point. Updated to use `SingleNestedBlock`. Thank you!

jriddle-linode (approved):
LGTM, working locally.

zliang-akamai (approved):

PR #1700, opened 2024-12-27
Deduplicate computed obj key regions; defer default value of `cors_enabled` to the API

zliang-akamai:
## 📝 Description

1. De-duplicate regions in object storage keys resource to prevent TF error
2. Defer determination of the default value of the `cors_enabled` attribute of the object storage bucket resource to the API due to the incoming change of the API default values. This might be a **breaking change** because it will change the plan value of the `cors_enabled` attribute from `true` to `known after apply`.

## ✔️ How to Test

```bash
make int-test PKG_NAME="linode/objkey"
```

```bash
make int-test PKG_NAME="linode/objbucket"
```


lgarber-akamai (approved):
Looks great, thank you for fixing this! I think the breaking change should be acceptable since the outcome would be more or less the same in most cases 👍

ezilber-akamai (approved):
Tests passing locally!

yec-akamai (approved):
Works well on my end!

PR #1712, opened 2025-01-10
Switch from daily dependencies updates to weekly

zliang-akamai:
This is to reduce frequency of PRs opened by @dependabot.


lgarber-akamai (approved):
This will make PR reviews a lot less tedious, thanks for doing this!

yec-akamai (approved):

PR #1713, opened 2025-01-13
bug(obj): Fix temporary keys with long bucket names

endocrimes:
## 📝 Description

When bucket names are >34 characters, the `createTempKeys` function begins to fail, as linode requires that object storage access keys are <= 50 chars. This change truncates bucket names to 34 characters for that label when required.

In the future, it may be desirable to reduce the label further and add a suffix that implies a name was truncated, but I haven't found an existing pattern for that in this provider.

## ✔️ How to Test

- Create a bucket with a long name (e.g `banana-global-loving-skunk-loki-chunks`) with Terraform and `obj_use_temp_keys` set.

I haven't added a test for this explicitly as the obj `resource_test` is mildly tricky to setup for first time use, but if this change is desirable I can add one.

### DIFF
--- a/linode/obj/helpers.go
+++ b/linode/obj/helpers.go
@@ -136,8 +136,16 @@ func createTempKeys(
 		tempBucketAccess.Region = regionOrCluster
 	}
 
+	// Bucket key labels are a maximum of 50 characters - if the bucket name is
+	// too long, then truncate it.
+	// We use 16 characters for `temp__{timestamp}`, so the maximum length of a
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
There might be a rare edge case where keys have duplicate labels but this shouldn't be a problem since duplicate key labels are allowed. Thank you for the contribution!

yec-akamai (approved):
Thank you for putting up the fix!

PR #1714, opened 2025-01-13
doc: Add doc example to grand explicit access to multiple buckets

yec-akamai:
## 📝 Description

Add an exmple to the object_storage_key doc for explaining how to grant a key access to multiple buckets.


ykim-akamai (approved):
LGTM

lgarber-akamai (approved):
Looks great! Always nice to provide more usage examples

PR #1738, opened 2025-01-23
Manually compute CRC32 sum in PutObject helper

lgarber-akamai:
## 📝 Description

This pull request manually computes the CRC32 sum of the object body for use in PutObjectInput struct. In theory this hash should be computed implicitly by the AWS SDK, but for some reason that functionality isn't working in the latest version.

## ✔️ How to Test

### Integration Testing

```
make PKG_NAME=linode/obj int-test
```


ezilber-akamai (approved):
Tests passing locally. Thanks for handling this!

ykim-akamai (approved):
Thanks for the quick fix

PR #1743, opened 2025-01-28
Adding Support for Object Storage Gen2 in Existing Resources and Data Sources

zliang-akamai:
## 📝 Description

This is to add support for object storage services gen2 in existing resources and data sources.

## ✔️ How to Test

```bash
make test-int PKG_NAME="objbucket"
```

```bash
make test-int PKG_NAME="objkey"
```

```bash
make test-int PKG_NAME="obj"
```

```hcl
provider "linode" {
}

resource "linode_object_storage_bucket" "foobar" {
  region        = "sg-sin-2"
  endpoint_type = "E2"
  endpoint      = "sg-sin-1.linodeobjects.com"
  label         = "zhiweitestsigr2bk"
}

resource "linode_object_storage_key" "foo" {
    label = "mykey"
}
```

### DIFF
--- a/docs/resources/object_storage_bucket.md
+++ b/docs/resources/object_storage_bucket.md
@@ -111,6 +111,10 @@ For example, `us-mia-1` cluster can be translated into `us-mia` region. Exactly 
   * configured by [`obj_secret_key`](../index.md#configuration-reference) in the provider configuration;
   * or, generated implicitly at apply-time if [`obj_use_temp_keys`](../index.md#configuration-reference) at provider-level is set.
 
+* `endpoint_type` - (Optional) The type of `s3_endpoint` available to the user in this region. See [Endpoint types](https://techdocs.akamai.com/cloud-computing/docs/object-storage#endpoint-type) for more information.
+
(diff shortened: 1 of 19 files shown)


lgarber-akamai (approved):
Tests are passing on my end, excellent work!

ykim-akamai (approved):
LGTM overall, there are still occasional intermittent failures due to API but not related to this PR. Nice work!

zliang-akamai:
Hey @ykim-akamai,

FYI:
The gen2 Object Storage clusters (e.g., `gb-lon-1`) doesn't support the `cluster` attribute. I added the filtering condition in the utility function to ensure it always select gen1 clusters, which support both `cluster` and `region` attribute.

The trade off is that we won't run any test against gen2 cluster, but API behaviors other than `cluster` vs `region` should be the same between gen1 and gen2.

ykim-akamai:
Thanks for the context and update @zliang-akamai . Tests seem to pass more consistently now beside the occasional 500 errors. Nice work!

PR #1746, opened 2025-01-29
Fix multiple miscellaneous issues

zliang-akamai:
## 📝 Description

Fix multiple issues found by CodeQL scan.

## ✔️ How to Test

```bash
make test-int PKG_NAME="firewall"
```

```bash
make test-int PKG_NAME="image"
```

```bash
make test-int PKG_NAME="objkey"
```

```bash
make test-int PKG_NAME="edns"
```

### DIFF
--- a/linode/rdns/framework_resource.go
+++ b/linode/rdns/framework_resource.go
@@ -181,13 +181,8 @@ func (r *Resource) Update(
 	defer cancel()
 
 	client := r.Meta.Client
-	resourceUpdated := false
 
 	if !state.RDNS.Equal(plan.RDNS) {
-		resourceUpdated = true
-	}
-
-	if resourceUpdated {
 		ip, err := updateIPAddress(
 			ctx,
 			client,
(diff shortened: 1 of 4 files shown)


linode/rdns/framework_resource.go
```diff
@@ -181,13 +181,9 @@ func (r *Resource) Update(
 	defer cancel()
 
 	client := r.Meta.Client
-
-	var updateOpts linodego.IPAddressUpdateOptions
-
 	resourceUpdated := false
 
 	if !state.RDNS.Equal(plan.RDNS) {
-		updateOpts.RDNS = plan.RDNS.ValueStringPointer()
```
yec-akamai, line 190: Is it necessary to remove updating RDNS? Looks like it's still an available function.

zliang-akamai: It's parsed to `updateIPAddress` function at line 194. This update options was never used and triggered the code scan alert.

ykim-akamai (approved):
LGTM, all tests passing

ezilber-akamai (approved):
All tests are passing locally. Nice work!

PR #1748, opened 2025-01-30
Fix `port` in mysqlv2 resource

ezilber-akamai:
## 📝 Description

Fixed the bug that was causing port to show up as 0 in the state after creating a mysqlv2 resource. Also pointed to latest linodego release.

## ✔️ How to Test
The following steps assume you have pulled down this PR locally.

### Unit Tests
`make test-unit`

### Integration Tests

### Manual Tests
1. In `dx-devenv`, run the following TF config:
```
 resource "linode_database_mysql_v2" "test" {
  region = "us-mia"
  label = "tf-test"
  type = "g6-nanode-1"
  engine_id = "mysql/8"

  allow_list = ["10.0.0.2/32", "10.0.0.3/32"]
  cluster_size = 3

  updates = {
    hour_of_day = 2
    day_of_week = 3
    frequency = "weekly"
    duration = 4
  }
}
```
2. Ensure that the DB creation was successful and verify that the DB shows up in Cloud Manager.
3. Look at the TF State and verify that the Port is not set to 0.
4. Clean up any resources created.


lgarber-akamai (approved):
Looks great, thanks for the quick fix on this!

Just double checking, do we have a TF follow-up ticket for the other changes made in TPT-3243?

ezilber-akamai:
> Looks great, thanks for the quick fix on this!
> 
> Just double checking, do we have a TF follow-up ticket for the other changes made in TPT-3243?

I don't believe so. In fact, I don't think we have TF follow-up tickets for any of the changes made in that epic (TPT-3247)

zliang-akamai (approved):

PR #1754, opened 2025-02-04
Fix issue causing DBs to get recreated with every update

ezilber-akamai:
## 📝 Description

Fixes a bug where DB Clusters would be destroyed and re-created even when making simple updates like label.

## ✔️ How to Test
Before the following test steps, make sure you pull down this PR locally.

### Integration Tests

### Manual Tests

```
resource "linode_database_mysql_v2" "test" {
  region = "us-mia"
  label = "tf-test"
  type = "g6-nanode-1"
  engine_id = "mysql/8"

  allow_list = ["10.0.0.2/32"]
  cluster_size = 3

  updates = {
    hour_of_day = 2
    day_of_week = 3
    frequency = "weekly"
    duration = 4
  }
}
```

2. Ensure that the DB Cluster was created successfully in Cloud Manager.
3. Make some changes to the `label` and `allow_list` in the TF config.
4. Reapply the config and make sure that the DB is not recreated.
5. Ensure that the updates went through successfully in Cloud Manager.
6. Repeat steps 1-5 with the following configuration:
```
resource "linode_database_postgresql_v2" "test" {
  region = "us-mia"
  label = "tf-test"
  type = "g6-nanode-1"
  engine_id = "postgresql/16"

  allow_list = ["10.0.0.2/32"]
  cluster_size = 3

  updates = {
    hour_of_day = 2
    day_of_week = 3
    frequency = "weekly"
    duration = 4
  }
}
```

7. Clean up all resources that you created.


lgarber-akamai (approved):

linode/helper/compare.go
```diff
@@ -10,6 +10,12 @@ import (
 )
 
 func CompareTimeStrings(t1, t2, timeFormat string) bool {
+	// time.Parse(...) does not support parsing empty string times,
+	// but this is a valid case in certain scenarios.
+	if t1 == "" && t2 == "" {
+		return true
+	}
+
```
zliang-akamai, line 18: Can the time string become empty strings? I feel we should return an error when they are empty strings..

zliang-akamai (approved):

linode/helper/compare.go
zliang-akamai: After discussion, we will fix this in a following up PR.

PR #1755, opened 2025-02-05
Update DB forked test cases

ykim-akamai:
## 📝 Description

## ✔️ How to Test

 `make test-int TEST_CASE="TestAccResource_fork" PKG_NAME="databasepostgresqlv2"`
 `make test-int TEST_CASE="TestAccResource_fork" PKG_NAME="databasemysqlv2"`

## 📷 Preview

### DIFF
--- a/linode/databasepostgresqlv2/resource_test.go
+++ b/linode/databasepostgresqlv2/resource_test.go
@@ -257,9 +257,10 @@ func TestAccResource_complex(t *testing.T) {
 				),
 			},
 			{
-				ResourceName:      resName,
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            resName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"updated"},
 			},
 		},
 	})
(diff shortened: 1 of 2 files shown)


linode/databasepostgresqlv2/resource_test.go
```diff
@@ -319,6 +320,11 @@ func TestAccResource_fork(t *testing.T) {
 					resource.TestCheckResourceAttr(resNameSource, "pending_updates.#", "0"),
 				),
 			},
+			{
+				ResourceName:      resNameSource,
+				ImportState:       true,
+				ImportStateVerify: true,
```
ezilber-akamai, line 326:
```suggestion
				ImportStateVerify: true,
				ImportStateVerifyIgnore: []string{"updated"},
```
Probably best to ignore `updated` here too just to be safe. Also, can you add the same thing in the `TestAccResource_complex` MySQL test? I was seeing some intermittent failures with this test yesterday that were caused by the `updated` field not quite matching up.

ezilber-akamai (approved):
LGTM other than a small change. Thanks for fixing this!

lgarber-akamai (approved):
All tests from both relevant test packages are passing on my end. Thanks for fixing this!

linode/databasepostgresqlv2/resource_test.go
ykim-akamai: Updated!

PR #1767, opened 2025-02-13
Bump golangci-lint -> v1.64.4

lgarber-akamai:
## 📝 Description

This pull request bumps golangci-lint to v1.64.4, which should resolve a compatibility issues between this repo's linter and go 1.24.


yec-akamai (approved):

tools/go.mod
```diff
@@ -1,191 +1,198 @@
 module github.com/linode/terraform-provider-linode/tools
 
-go 1.23
+go 1.23.0
+
+toolchain go1.23.2
```
yec-akamai, line 5: Do we still need to pin the toolchain version?

zliang-akamai (approved):

tools/go.mod
lgarber-akamai: Hmm, I'm not entirely sure. It looks like it was implicitly added by `go mod tidy` with go 1.23

PR #1773, opened 2025-02-19
fix: Replace update polling with status polling for DBaaS v2 updates

ezilber-akamai:
## 📝 Description

Since update events for DBaaS v2 are not always emitted properly, this PR switches to status polling for better consistency.

### Integration Tests


lgarber-akamai (approved):
Tests are passing on my end, thank you for fixing this!

zliang-akamai (approved):

Issue #1764, opened 2025-02-10
[Bug]: object storage broken with v2.34.0


Aransh:
### Terraform Version

Terraform v1.10.4 on darwin_arm64

### Linode Provider Version

version = "2.34.0"

### Effected Terraform Resources

linode_object_storage_bucket

### Terraform Config Files

```
data "linode_object_storage_cluster" "region" {
  id = "us-mia-2"
}

resource "linode_object_storage_bucket" "storage_bucket" {
  cluster = data.linode_object_storage_cluster.region.id
  label   = "aran-test"
  acl     = "private"

  lifecycle_rule {
    id      = "retention_rule"
    enabled = true
    expiration {
      days = 90
    }
  }

  access_key = "XXXX"
  secret_key = "XXXX"
}
```

### Debug Output

$ TF_LINODE=DEBUG LINODE_DEBUG=1 terraform apply
```
data.linode_object_storage_cluster.region: Reading...
data.linode_object_storage_cluster.region: Read complete after 1s [id=us-mia-2]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # linode_object_storage_bucket.storage_bucket will be created
  + resource "linode_object_storage_bucket" "storage_bucket" {
      + access_key    = (sensitive value)
      + acl           = "private"
      + cluster       = "us-mia-2"
      + cors_enabled  = (known after apply)
      + endpoint      = (known after apply)
      + endpoint_type = (known after apply)
      + hostname      = (known after apply)
      + id            = (known after apply)
      + label         = "aran-test"
      + region        = (known after apply)
      + s3_endpoint   = (known after apply)
      + secret_key    = (sensitive value)
      + versioning    = (known after apply)

      + lifecycle_rule {
          + enabled = true
          + id      = "retention_rule"

          + expiration {
              + days = 90
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
╷
│ Warning: Deprecated
│
│   with data.linode_object_storage_cluster.region,
│   on test.tf line 1, in data "linode_object_storage_cluster" "region":
│    1: data "linode_object_storage_cluster" "region" {
│
│ This data source has been deprecated because it relies on deprecated API endpoints. Going forward, `region` will be the preferred way to designate where
│ Object Storage resources should be created.
│
│ (and one more similar warning elsewhere)
╵
╷
│ Warning: Argument is deprecated
│
[...]

PR #1787, opened 2025-02-24
Correct the key for bucket label

zliang-akamai:
## 📝 Description

Fix: #1764

When s3_endpoint is not available from the API, we should compute the endpoint from other attribute. The current compute function was designed for the object resource, which has been migrated to framework and no longer requiring this compute function; so modifying the function to fit the bucket resource in this PR.

Some other cleanups are also included.

## ✔️ How to Test
```bash
make test-int PKG_NAME="objbucket"
```


lgarber-akamai (approved):
Great work and thanks for fixing this!

PR #1788, opened 2025-02-24
Fix multiple issues with LKE Enterprise

yec-akamai:
## 📝 Description

There were multiple issues block an enterprise lke from creation. This PR fixes the following problems:

- LKE enterprise doesn't support dashboard. Skipping getting a dashboard URL in read and create.
- The kubeconfig of an enterprise LKE currently takes a lot of time to be generated. Wait for it to be ready specifically after sending the creation request. Extend the retry timeout for an enterprise LKE to get available nodes to avoid context canceled before getting any of the result. 

## ✔️ How to Test

Integration test:
```
make PKG_NAME=lke TEST_CASE=TestAccResourceLKECluster_enterprise test-int
```

Manual test:
1. In a sandbox environment, pull this change to initialize terraform and use the following config to create a lke-e:
```
provider "linode" {
  api_version="v4beta"
}

    label       = "lke-e-test-tf"
    region      = "id-cgk"
    k8s_version = "v1.31.1+lke1"
    tags        = ["test"]
    tier = "enterprise"

    pool {
        type  = "g6-standard-1"
        count = 3
        tags  = ["test"]
    }
}
```
2. Run this config and observe the cluster is created successfully.


linode/lke/framework_datasource.go
```diff
@@ -69,6 +69,20 @@ func (r *DataSource) Read(
 		return
 	}
 
+	var dashboard *linodego.LKEClusterDashboard
+
+	// LKE Enterprise does not have a dashboard URL.
+	if cluster.Tier != TierEnterprise {
```
lgarber-akamai, line 75: optional: Would it make more sense to check if the tier is explicitly `standard` here to proof us for potential future tiers?

linode/lke/framework_models.go
```diff
@@ -192,7 +196,11 @@ func (data *LKEDataModel) parseLKEAttributes(
 	}
 	data.APIEndpoints = apiEndpoints
 
-	data.DashboardURL = types.StringValue(dashboard.URL)
+	if dashboard != nil {
+		data.DashboardURL = types.StringValue(dashboard.URL)
+	} else {
+		data.DashboardURL = types.StringNull()
+	}
```
zliang-akamai, line 203: Let's use `StringPointerValue` to replace the if...else...?

linode/lke/framework_models.go
```diff
@@ -179,7 +179,11 @@ func (data *LKEDataModel) parseLKEAttributes(
 	}
 	data.Pools = lkePools
 
-	data.Kubeconfig = types.StringValue(kubeconfig.KubeConfig)
+	if kubeconfig != nil {
+		data.Kubeconfig = types.StringValue(kubeconfig.KubeConfig)
+	} else {
+		data.Kubeconfig = types.StringNull()
+	}
```
zliang-akamai, line 186: Also here

linode/lke/framework_models.go
yec-akamai: dashboard.URL is not a string pointer, I'm afraid we can't use `StringPointerValue` here directly.

linode/lke/framework_models.go
yec-akamai: Same here that kubeconfig.KubeConfig is not a string pointer.

ykim-akamai (approved):
LGTM, test pass pretty consistently now 👍

linode/lke/framework_models.go
zliang-akamai: Oh, yep you are right

zliang-akamai (approved):
Works well and looks good to me!

PR #1794, opened 2025-02-26
Fix LKE control plane ACL test to match the updated API behavior

yec-akamai:
## 📝 Description

LKE control ACL addresses will be removed if set `enabled` to be false. Update test case to match this new behavior. 

## ✔️ How to Test

```
make PKG_NAME=lke TEST_CASE=TestAccResourceLKECluster_controlPlane test-int
```


ezilber-akamai (approved):
Thanks for the fix! Tests passing locally

PR #1800, opened 2025-03-03
Disallow addresses when ACL is set to disabled

zliang-akamai:
## 📝 Description

Disallow configuring addresses when ACL is disabled, because the addresses will be ignored by the API.

## ✔️ How to Test

Test case:
```bash
make test-int PKG_NAME="lke" TEST_CASE="TestAccResourceLKECluster_acl_disabled_addresses"
```

Verify an error returned for this config:

```hcl
    label       = "test-acl-addresses"     
    k8s_version = "1.32"           
    region      = "us-mia"     

    control_plane {      
        acl {
            enabled = false
            addresses {
                ipv4 = ["0.0.0.0/0"]
                ipv6 = ["2001:db8::/32"]
            }
        }
    }

    pool {
        type  = "g6-standard-1"
        count = 1
    }
}
```


ykim-akamai (approved):
LGTM, verified locally! Error like below shows when ACL is disabled
```
│ Error: addresses are not acceptable when ACL is disabled.
```

yec-akamai (commented):
Can we add some test cases for this scenario?

yec-akamai (approved):
Tests passed locally, nice work! Thanks for addressing the comments

PR #1804, opened 2025-03-06
Support Managed Database Suspension

lgarber-akamai:
## 📝 Description

This pull request adds support for suspending and resuming databases using the `linode_database_mysql_v2` and `linode_database_postgresql_v2` resources using a new `suspended` field. Additionally, a read-only version of this field is exposed under the corresponding data sources.

This change also resolves an issue that would prevent suspended databases from having their credentials and SSL configurations read, causing errors on Terraform plans.

## ✔️ How to Test

### Integration Testing

```
make test-int PKG_NAME=databasemysqlv2
make test-int PKG_NAME=databasepostgresqlv2
```

### Manual Testing

```terraform
resource "linode_database_postgresql_v2" "foobar" {
  label = "tf-db-suspension-test-postgresql"
  type = "g6-nanode-1"
  region = "us-mia"
  engine_id = "postgresql/16"

  suspended = true
}

resource "linode_database_mysql_v2" "foobar" {
  label = "tf-db-suspension-test-mysql"
  type = "g6-standard-1"
  region = "us-mia"
  engine_id = "mysql/8"

  suspended = true
}
```

2. Ensure the configuration applies successfully and the newly created databases are suspended in Cloud Manager.
3. Update the `suspended` attribute of both resources to be `false` and apply the configuration.
4. Ensure both databases are successfully resumed.


linode/databasepostgresqlv2/framework_resource_schema.go
```diff
@@ -82,9 +84,10 @@ var frameworkResourceSchema = schema.Schema{
 			},
 		},
 		"ca_cert": schema.StringAttribute{
-			Description: "The base64-encoded SSL CA certificate for the Managed Database.",
-			Computed:    true,
-			Sensitive:   true,
+			Description:   "The base64-encoded SSL CA certificate for the Managed Database.",
+			Computed:      true,
+			Sensitive:     true,
+			PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()},
```
lgarber-akamai, line 90: Corresponds to the workaround in framewor_models.go

linode/databasepostgresqlv2/framework_models.go
```diff
@@ -141,14 +148,34 @@ func (m *Model) Flatten(
...
+	m.Suspended = helper.KeepOrUpdateBool(m.Suspended, helper.DatabaseStatusIsSuspended(db.Status), preserveKnown)
 	m.Type = helper.KeepOrUpdateString(m.Type, db.Type, preserveKnown)
 	m.Updated = helper.KeepOrUpdateValue(m.Updated, timetypes.NewRFC3339TimePointerValue(db.Updated), preserveKnown)
 	m.Version = helper.KeepOrUpdateString(m.Version, db.Version, preserveKnown)
 
+	// SSL and credentials may be nil if the database is suspended
+	if ssl != nil {
+		m.CACert = helper.KeepOrUpdateString(m.CACert, string(ssl.CACertificate), preserveKnown)
+	} else {
+		// We always enable perserveKnown here because it will otherwise
```
lgarber-akamai, line 162: Any thoughts on workarounds for this? Unfortunately most of the clean solutions I found require newer TF and proto versions than our minimum supported version

ykim-akamai (approved):
LGTM, tests pass locally. Nice work!

PR #1816, opened 2025-03-12
Added additional examples to `lkeversions` datasource docs

ezilber-akamai:
## 📝 Description

Added an example of accessing the version id strings (i.e. `1.32`, `v1.31.1+lke1`) from the datasource to avoid confusion with the TF resource's id.


ykim-akamai (approved):

lgarber-akamai (approved):
Thanks for the quick turnaround on this!

PR #1823, opened 2025-03-24
Some more resource and data source names from config missing in the previous PR

zliang-akamai:
Some missed items supposed to be in #1801 

Adding them here


jriddle-linode (approved):
Looks much better!

yec-akamai (approved):
LGTM

PR #1827, opened 2025-03-26
Fix the k8s version for LKE-E test

yec-akamai:
## 📝 Description

Remove the fixed k8s version and grab it dynamically for lke-e test.

## ✔️ How to Test
```
make PKG_NAME=lke TEST_CASE=TestAccResourceLKECluster_enterprise test-int 
```

### DIFF
--- a/linode/lke/framework_resource_test.go
+++ b/linode/lke/framework_resource_test.go
@@ -23,10 +23,11 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Thanks for fixing this!

ezilber-akamai (approved):

PR #1833, opened 2025-04-03
Restore update polling and add resize polling for DBaaS V2 Resources

ezilber-akamai:
## 📝 Description

Replace status polling with update event and resize event polling for DBaaS V2 resources.

### Integration Tests


linode/databasemysqlv2/framework_resource.go
```diff
@@ -364,25 +365,41 @@ func (r *Resource) Update(
...
-		//		"Failed to create EventPoller for database",
-		//		err.Error(),
-		//	)
-		//	return
-		//}
+	if shouldUpdate || shouldResize {
+		var updatePoller *linodego.EventPoller = nil
+		var updatePollerErr error = nil
+		var resizePoller *linodego.EventPoller = nil
+		var resizePollerErr error = nil
```
lgarber-akamai, line 380: nitpick: The `= nil` can be excluded here since it's implied

linode/databasemysqlv2/framework_resource.go
```diff
@@ -364,25 +365,41 @@ func (r *Resource) Update(
...
-		//if err != nil {
-		//	resp.Diagnostics.AddError(
-		//		"Failed to create EventPoller for database",
-		//		err.Error(),
-		//	)
-		//	return
-		//}
+	if shouldUpdate || shouldResize {
+		var updatePoller *linodego.EventPoller = nil
+		var updatePollerErr error = nil
```
lgarber-akamai, line 378: nitpick: Given these run synchronously, we might be able to consolidate the error vars here, or use the existing `err` from scope

linode/databasemysqlv2/framework_resource.go
ezilber-akamai: Good point. Fixed!

linode/databasemysqlv2/framework_resource.go
ezilber-akamai: I consolidated the error vars. The existing `err` from scope wasn't accessible from within this `if` block, so it was simpler to use a new one.

PR #1849, opened 2025-04-18
Deprecate `ip_address` attribute of `linode_instance` resource

zliang-akamai:
## 📝 Description

Since we plan to remove this attribute in the next major version of the provider, add deprecate warning now.


linode/instance/schema_resource.go
```diff
@@ -317,6 +317,9 @@ var resourceSchema = map[string]*schema.Schema{
 		Description: "This Linode's Public IPv4 Address. If there are multiple public IPv4 addresses on this " +
 			"Instance, an arbitrary address will be used for this field.",
 		Computed: true,
+		Deprecated: "The `ip_address` attribute in linode_instance resource is deprecated. " +
```
lgarber-akamai, line 320: Could we also mark this as deprecated in the documentation?

ykim-akamai (approved):
LGTM

PR #1854, opened 2025-04-23
linode_lke_node_pool: Resolve LKE node pool autoscaler disable error when using explicit node_count

lgarber-akamai:
## 📝 Description

This pull request resolves an issue that caused the following validation error when attempting to disable the autoscaler of an existing LKE node pool:

```
[400] [autoscaler.min] Must be 1-500; [autoscaler.max] Must be 1-500
```

This error occurred because the PUT request body specifies a zero value for the `min` and `max` fields alongside `enabled: false`. This pull request resolves the issue by setting those values to the plan's node count if necessary.

## ✔️ How to Test

### Integration Testing

```
make PKG_NAME=lkenodepool TEST_ARGS="-run TestAccResourceNodePool_disableAutoscalingExplicitNodeCount" test-int
```

### Manual Testing

```terraform
  label = "tf-test-manual"
  region = "us-mia"
  k8s_version = "1.32"

  pool {
    count = 1
    type  = "g6-standard-1"
  }

  external_pool_tags = ["external_pool"]
}

resource "linode_lke_node_pool" "test" {
  cluster_id = linode_lke_cluster.test.id
  type = "g6-standard-1"
  tags = linode_lke_cluster.test.external_pool_tags
  node_count = 2

  autoscaler {
    min = 2
    max = 4
  }
}
```

2. Ensure the apply finishes successfully and all relevant resources have been created.
3. Comment out the autoscaler block:

```terraform
...

  # autoscaler {
  #   min = 2
  #   max = 4
  # }

...
```

4. Re-apply the configuration and ensure autoscaling has successfully been disabled on the externally managed node pool.

### DIFF
--- a/linode/lkenodepool/framework_models.go
+++ b/linode/lkenodepool/framework_models.go
@@ -178,15 +178,16 @@ func (pool *NodePoolModel) SetNodePoolUpdateOptions(
 ) bool {
 	var shouldUpdate bool
 
-	if !state.Count.Equal(pool.Count) {
-		p.Count = helper.FrameworkSafeInt64ToInt(
-			pool.Count.ValueInt64(),
-			diags,
-		)
-		if diags.HasError() {
-			return false
-		}
+	plannedCount := helper.FrameworkSafeInt64ToInt(
+		pool.Count.ValueInt64(),
+		diags,
+	)
+	if diags.HasError() {
+		return false
+	}
 
+	if !state.Count.Equal(pool.Count) {
+		p.Count = plannedCount
 		shouldUpdate = true
 	}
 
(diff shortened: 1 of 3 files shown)


yec-akamai (approved):
Tested and works well on my end, nice fix!

linode/lkenodepool/framework_models.go
```diff
@@ -283,8 +283,12 @@ func (pool *NodePoolModel) shouldUpdateLKENodePoolAutoscaler(
 		}
 	} else if len(state.Autoscaler) > 0 {
 		autoscaler.Enabled = false
-		autoscaler.Min = count
-		autoscaler.Max = count
+
+		// These values are discarded by the API when autoscaling is disabled,
+		// so we can just use the plan's count since it should meet the validation requirements.
+		planNodeCount := helper.FrameworkSafeInt64ToInt(pool.Count.ValueInt64(), diags)
+		autoscaler.Min = planNodeCount
```
zliang-akamai, line 290: Let's check the `diags` here after the function called.

zliang-akamai:
This change seems to be causing `TestAccResourceNodePool_enableAutoscaling` failing.

PR #1858, opened 2025-04-25
Add doc for node pool labels and taints

zliang-akamai:
## 📝 Description

This is to add the doc of labels and taints of LKE node pools.


yec-akamai (approved):

ykim-akamai (approved):
Thanks for updating these! Easy ones to miss

PR #1898, opened 2025-06-02
Add interface_id in various networking IP data sources

zliang-akamai:
## 📝 Description

Add interface_id to networking IP data sources.

With some statecheck migration in the test cases.

## ✔️ How to Test

```bash
make test-int PKG_NAME="networkingip"
```

```bash
make test-int PKG_NAME="networkingips"
```


ezilber-akamai (commented):
Got the following test failure locally:
```
=== NAME  TestAccDataSourceNetworkingIP_basic
    framework_datasource_test.go:39: Step 2/2 error: Post-apply refresh state check(s) failed:
        error checking value for attribute at path: data.linode_networking_ip.foobar.interface_id, err: expected non-nil value for NotNull check, got: <nil>
--- FAIL: TestAccDataSourceNetworkingIP_basic (88.65s)
```

zliang-akamai:
@ezilber-akamai Thanks! Just fixed

ezilber-akamai (approved):
Tests passing now. Nice work!

PR #1902, opened 2025-06-04
Add interfaces support in firewalls data source

zliang-akamai:
## 📝 Description

Add interfaces support in firewalls data source.

## ✔️ How to Test

Auto tests only pass on accounts that don't have default firewall setting.

```bash
make test-int PKG_NAME="firewalls"
```


lgarber-akamai:
> Auto tests only pass on accounts that don't have default firewall setting.

I take it we'll want to avoid configuring a default firewall on our CI accounts in that case?

zliang-akamai:
> > Auto tests only pass on accounts that don't have default firewall setting.
> 
> I take it we'll want to avoid configuring a default firewall on our CI accounts in that case?

Yes, at least before we figure out the way to make them work together.

ykim-akamai (approved):
LGTM

PR #1864, opened 2025-05-02
Add `interfaces_for_new_linodes` attribute in the account settings resource and data source

zliang-akamai:
## 📝 Description

This is to update the account settings resource and data source with new API spec from the enhanced Linode interfaces project.

Apparently, the `linode_account_settings` resource never worked as expected... fix it in this PR as well.

Test check functions have been migrated to be the state checks.

## ✔️ How to Test
```bash
ACC_OPT_IN_TESTS=TestAccResourceAccountSettings_basic,TestAccResourceAccountSettings_update,TestAccDataSourceLinodeAccountSettings_basic make test-int PKG_NAME="accountsettings"
```

PR #1903, opened 2025-06-04
Update VPC and account setting docs

zliang-akamai:
## 📝 Description

Add docs for https://github.com/linode/terraform-provider-linode/pull/1864 and https://github.com/linode/terraform-provider-linode/pull/1896


ykim-akamai (approved):
LGTM

PR #1921, opened 2025-06-10
Update runDiskOperation to prevent rebooting an instance without disk

lucaferocino:
## 📝 Description

This PR addresses the issue where the Linode Terraform provider attempts to reboot an instance after deleting a disk, even when no disks remain attached. Currently, this behavior results in errors because Linode instances cannot boot without any attached disks. The change adds a verification step to check if any disks are attached before initiating a reboot, thereby preventing unnecessary errors and ensuring smoother Terraform operations.

## ✔️ How to Test

1. Create a Linode instance with a single disk.
2. Use Terraform to delete the only disk attached to this instance.
3. Without this fix, Terraform attempts a reboot, causing an error since the instance has no disks.
4. With this fix, Terraform will skip the reboot if no disks are attached, logging an informative message instead.

```hcl
terraform {
  required_providers {
    linode = {
       source  = "linode/linode"
       version = "2.41.0"
    }
  }
}

provider "linode" {}

variable "ubuntu_version" {
  type    = string
  default = "22"
}

resource "linode_instance" "web" {
  label      = "complex_instance"
  tags       = ["foo"]
  region     = "us-central"
  type       = "g6-nanode-1"
  private_ip = true
}

resource "linode_instance_disk" "boot_disk" {
  label     = "boot1"
  linode_id = linode_instance.web.id

  size  = 5500
  image = "linode/ubuntu${var.ubuntu_version}.04"
}

resource "linode_instance_config" "boot_config" {
  label     = "boot_config"
  linode_id = linode_instance.web.id

  device {
    device_name = "sda"
    disk_id     = linode_instance_disk.boot_disk.id
  }

  root_device = "/dev/sda"
  kernel      = "linode/latest-64bit"
  booted      = true
}
```
With the sample terraform above, the following commands will reproduce the issue:
1. This will create a simple instance
```bash
terraform apply
```
2. This will change its image, hence forcing a disk replacement operation. Without these changes, this operation will fail because the provider will attempt to reboot the instance when there is no disk attached
```bash
terraform apply -var "ubuntu_version=24"
```

I added a test to demonstrate this behavior. To run it execute
```bash
make PKG_NAME="instanceconfig" TEST_CASE="TestAccResourceInstanceConfig_diskReplacement" test-int  
```


lucaferocino:
> But the two customize check functions in test case seem to be overkill and are more complicated than others. Would you mind to remove them?

Hi @zliang-akamai, I removed the custom checks from the test case as requested. Please do let me know if it looks better now. With the current implementation we no longer check that the disk was indeed replaced, but we do still test that the instance reboot behavior is correct after a disk replacement, which should be sufficient.

ezilber-akamai (approved):
Thank you for the contribution!

lucaferocino:
What are the timelines for this fix to be merged and a new version to be released? One of our stacks critically depends on this. Thank you!

PR #1953, opened 2025-07-01
Make config_id nullable for the VPC IPs data source

zliang-akamai:
## 📝 Description

Set the `config_id` to null when 0 is returned from linodego.

We don't want to make a breaking change at this time for the corresponding field in linodego so checking the zero value here as a workaround.

## ✔️ How to Test

```
make test-int PKG_NAME="vpcips"
```

Due to an potential API bug, the API filters are not available and may cause the test failure. We can temporary remove the filters in the test config template to make it pass locally, as long as there is not another Linode instance in you account with a VPC IP. 

```diff
diff --git a/linode/vpcips/tmpl/data_basic.gotf b/linode/vpcips/tmpl/data_basic.gotf
index a4d70cbc..dc226bee 100644
--- a/linode/vpcips/tmpl/data_basic.gotf
+++ b/linode/vpcips/tmpl/data_basic.gotf
@@ -65,10 +65,6 @@ resource "linode_instance_config" "barfoo" {
 
 data "linode_vpc_ips" "foobar" {
   depends_on = [linode_instance_config.foobar, linode_instance_config.barfoo]
-  filter {
-      name = "linode_id"
-      values = [linode_instance.foobar.id, linode_instance.barfoo.id]
-    }
 }
 
 data "linode_vpc_ips" "barfoo" {
```


ezilber-akamai (approved):
Test passed locally after removing the filter as you suggested. Do we want to remove it in this PR as well to avoid breaking the tests?

zliang-akamai:
> The change works well, i wonder if we should remove the filter from the test for now until the API issue get fixed?

The filter is required for accounts running multiple tests like our test accounts. I will be following up with the issue and modify the test if it's still not fixed before project branch merged.

PR #1975, opened 2025-07-15
fix: Removed unused variables from `instanceconfig` tests

ezilber-akamai:
## 📝 Description

Removed unused variables from `instanceconfig` tests

## ✔️ How to Test

`make test-int`


yec-akamai (approved):
Thanks for the fix!

PR #1976, opened 2025-07-15
Remove rDNS check for reserved IP test

zliang-akamai:
## 📝 Description

This is a computed field and the value is determined by the API. Removing for reserved IP test to allow it to pass.

## ✔️ How to Test

```bash
make test-int PKG_NAME="networkingips" TEST_CASE="TestAccDataSourceNetworkingIP_filterReserved"
```


ezilber-akamai (approved):
Thanks for fixing this!

PR #1993, opened 2025-07-28
Updated Database `engine_config` tests

ezilber-akamai:
## 📝 Description

Updated DB engine config test suite to check values or not null rather than checking for exact values to reduce test flakiness.

## ✔️ How to Test

The following steps assume you have pulled down this PR locally.

### Integration Tests
`make test-int TEST_SUITE="databasemysqlconfig"`
`make test-int TEST_SUITE="databasepostgresqlconfig"`


yec-akamai (approved):
Tests passed locally, nice work!

PR #2002, opened 2025-07-30
Add nested object update helper

zliang-akamai:
## 📝 Description

Add a generic helper for updating a nested object.

### Integration test

```bash
make test-int PKG_NAME="firewallsettings"
```

### Unit tests

```bash
make test-unit PKG_NAME="helper"
```

```bash
make test-unit PKG_NAME="firewallsettings"
```


lgarber-akamai (approved):
This is much cleaner. Great work!

PR #2044, opened 2025-08-20
Require `Maintenance Policy` capability in related tests

lgarber-akamai:
## 📝 Description

This pull request updates the `TestAccResourceInstance_updateMaintenancePolicy` and `TestAccDataSourceInstances_basic` integration tests to use a region with the `Maintenance Policy` capability, which should resolve the test failures caused by the limited rollout.

CI Integration Test Run (failures unrelated): https://github.com/linode/terraform-provider-linode/actions/runs/17135246549

## ✔️ How to Test

### Integration Testing

```
```

### DIFF
--- a/linode/maintenancepolicies/framework_datasource_test.go
+++ b/linode/maintenancepolicies/framework_datasource_test.go
@@ -3,6 +3,7 @@
 package maintenancepolicies_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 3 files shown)


linode/maintenancepolicies/framework_datasource_test.go
```diff
@@ -48,7 +49,13 @@ func TestAccDataSourceMaintenancePolicies_basic(t *testing.T) {
...
 						resourceName,
 						tfjsonpath.New("maintenance_policies").AtSliceIndex(0).AtMapKey("notification_period_sec"),
-						knownvalue.Int64Exact(300),
+						knownvalue.Int64Func(func(value int64) error {
+							if value <= 0 {
+								return fmt.Errorf("expected non-zero notification_period_sec")
+							}
+
+							return nil
+						}),
```
lgarber-akamai, line 58: I noticed this assertion failing in the CI because the API-returned value for `notification_period_sec` changed at some point. Since its value isn't guaranteed, I decided to just check for > 0 here.

PR #2058, opened 2025-08-27
Change all dependency update intervals to be monthly

zliang-akamai:
## 📝 Description

[The previous PR](https://github.com/linode/terraform-provider-linode/pull/2043) trying to change the update interval of AWS dependency update intervals [has failed](https://github.com/linode/terraform-provider-linode/runs/49020643306) because the dependabot doesn't support configuring the interval for a specific dependencies.

Considering change all intervals to be monthly for solving it. The security updates is still immediately applied and won't be affected by this config change, because it's operating separately.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,47 +9,33 @@ updates:
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     ignore:
       - dependency-name: "github.com/linode/terraform-provider-linode/linode"
-      - dependency-name: "github.com/aws/aws-sdk-go-v2*"
-  - package-ecosystem: "gomod"
-    directory: "/"
-    schedule:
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Seems like a sensible change to me, thank you for the quick fix!

yec-akamai (commented):
I wonder if it makes more sense to change them to be bi-weekly. Because currently the terraform release cadence is bi-weekly, to align with that we can make sure when cutting the release the dependencies are updated.

zliang-akamai:
> I wonder if it makes more sense to change them to be bi-weekly. Because currently the terraform release cadence is bi-weekly, to align with that we can make sure when cutting the release the dependencies are updated.

@yec-akamai Unfortunately, this option is not provided by dependabot...

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#interval

yec-akamai (approved):

PR #2078, opened 2025-09-02
UseStateForUnknown on database sensitive values only if not null

lgarber-akamai:
## 📝 Description

This pull request adds a new `UseStateForUnknownIf` plan modifier, which allows us to use the state value of a database's sensitive fields only if their values are not null. 

This resolves the following errors that surfaced in CI and are reproducible locally, while still allowing users to configure databases :

```
Error: Provider produced inconsistent result after apply
        
        When applying changes to linode_database_postgresql_v2.foobar, provider
        "provider[\"registry.terraform.io/hashicorp/linode\"]" produced an unexpected
        new value: .root_password: inconsistent values for sensitive attribute.
```

## ✔️ How to Test

### Integration Testing

```
make test-int PKG_NAME=databasemysqlv2
make test-int PKG_NAME=databasepostgresqlv2
```

### Manual Testing

```terraform
resource "linode_database_postgresql_v2" "test" {
  label = "foobar2"
  region = "us-mia"
  type = "g6-nanode-1"
  engine_id = "postgresql/17"
  suspended = true
}

output "test" {
  value = linode_database_postgresql_v2.test
  sensitive = true
}
```

3. Set `suspended = false` and re-apply the configuration.
4. Ensure no errors are raised.

### DIFF
--- a/linode/helper/stringplanmodifiers/caseinsensitiveset.go
+++ b/linode/helper/stringplanmodifiers/caseinsensitiveset.go
@@ -1,4 +1,4 @@
-package planmodifiers
+package stringplanmodifiers
 
 import (
 	"context"
--- a/linode/helper/stringplanmodifiers/usestateforunknownif.go
+++ b/linode/helper/stringplanmodifiers/usestateforunknownif.go
@@ -1,0 +1,68 @@
+package stringplanmodifiers
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+// UseStateForUnknownIfNotNull is a convenience wrapper to only use the state value
+// in place of unknown values in plans if its value is not null.
+func UseStateForUnknownIfNotNull() planmodifier.String {
+	return UseStateForUnknownIf(
+		func(ctx context.Context, request planmodifier.StringRequest) bool {
+			return !request.StateValue.IsNull()
+		},
+	)
+}
+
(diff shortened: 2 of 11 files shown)


linode/helper/stringplanmodifiers/usestateforunknownif.go
```diff
@@ -0,0 +1,68 @@
...
+
+type planModifierCondition func(context.Context, planmodifier.StringRequest) bool
+
+// UseStateForUnknownIf returns a plan modifier that will only use the state value
+// in place of an unknown value if the given condition is met.
+func UseStateForUnknownIf(condition planModifierCondition) planmodifier.String {
+	return useStateForUnknownIfModifier{
+		conditionFunc: condition,
+	}
+}
```
lgarber-akamai, line 27: I wanted to make this type-agnostic but couldn't figure out a good way to do it. For the time being I renamed the package to keep it type-specific.

If anyone has any ideas some feedback would be much appreciated! We can revisit this in the future if not.

linode/helper/stringplanmodifiers/caseinsensitiveset.go
zliang-akamai, line 1: I think this modifier can go to another directory maybe named as `setplanmodifiers`?

lgarber-akamai: Oops, good call!

yec-akamai (approved):
Tests passed locally!

PR #2087, opened 2025-09-19
Implement linode interface resource

zliang-akamai:
## 📝 Description

Implement Linode interface resource

## ✔️ How to Test

```bash
make PKG_NAME="linodeinterface" test-int
```

```hcl
    label = "test-interface"
    region = "us-mia"
    type = "g6-nanode-1"
    interface_generation = "linode"
}

resource "linode_interface" "test" {
    
    public = {
        ipv4 = {
            addresses = [
                {
                    address = "auto"
                    primary = true
                }
            ]
        }
    }
}
```

### DIFF
--- a/docs/resources/interface.md
+++ b/docs/resources/interface.md
@@ -6,9 +6,9 @@ description: |-
 
 # linode\_interface
 
-Provides a Linode Interface resource. This can be used to create, modify, and delete network interfaces for Linode instances. Interfaces allow you to configure public, VLAN, and VPC networking for your Linode instances.
+Provides a Linode Interface resource that can be used to create, modify, and delete network interfaces for Linode instances. Interfaces allow you to configure public, VLAN, and VPC networking for your Linode instances.
 
-This resource is for Linode interfaces only, and if you are interested in deploying a Linode instance with legacy config interface, you can checkout the doc for `linode_instance_config` resource for details.
+This resource is specifically for Linode interfaces. If you are interested in deploying a Linode instance with a legacy config interface, please refer to the `linode_instance_config` resource documentation for details.
 
 This resource is designed to work with explicitly defined disk and config resources for the Linode instance. See the [Complete Example with Linode](#complete-example-with-linode) section below for details.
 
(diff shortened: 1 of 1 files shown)


linode/helper/conversion.go
```diff
@@ -110,6 +111,38 @@ func FrameworkSafeInt64PointerToIntPointer(number *int64, diags *diag.Diagnostic
 	return linodego.Pointer(result)
 }
 
+func FrameworkSafeInt64ValueToIntDoublePointerWithUnknownToNil(v types.Int64, diags *diag.Diagnostics) **int {
```
zliang-akamai, line 114: I really don't know how to shorten naming of these functions. Let me know if you have any idea..

linode/linodeinterface/framework_vpc_models.go
```diff
@@ -0,0 +1,198 @@
+package linodeinterface
```
lgarber-akamai, line 1: The sub-models being split into separate files is much more readable!

lgarber-akamai (approved):
This is super readable and works well in my local testing. Excellent work!

linode/linodeinterface/framework_resource_schema.go
```diff
@@ -0,0 +1,395 @@
...
+		"address": schema.StringAttribute{
+			Optional: true,
+			Computed: true,
+			Default:  stringdefault.StaticString("auto"),
+		},
+		"primary": schema.BoolAttribute{
+			Optional: true,
+		},
+	},
+}
```
lgarber-akamai, line 32: Is it possible that the optional-only sub-models will cause issues when importing? We do this with other resources so I'd imagine it's probably fine if their new configuration matches.

linode/linodeinterface/framework_resource.go
```diff
@@ -0,0 +1,248 @@
...
+		ctx, req, resp,
+		[]helper.ImportableID{
+			{
+				Name:          "id",
+				TypeConverter: helper.IDTypeConverterString,
+			},
+			{
+				Name:          "linode_id",
+				TypeConverter: helper.IDTypeConverterInt64,
+			},
```
lgarber-akamai: Sorry for the post-approval feedback! I think the order of these needs to be swapped to match other resources (parent_id,id)

linode/linodeinterface/framework_resource_schema.go
zliang-akamai: I don't see any problem when testing this. I think it's okay except we would have to skip some attributes in the import verification checks. BTW I for got adding the import verification checks... which I will do later today

ezilber-akamai (approved):
Great work!

PR #2097, opened 2025-09-30
project: VPC Dual Stack

lgarber-akamai:


yec-akamai (approved):

zliang-akamai:
Can we migrate the check functions to statecheck?

lgarber-akamai:
> Can we migrate the check functions to statecheck?

Good call! I'll do this real quick

PR #2123, opened 2025-10-07
Fix instance resource boot status waiting

zliang-akamai:
## 📝 Description

Correct the behavior of the waiting of desired status.

Easier diff view from VS Code:
image

## ✔️ How to Test
```bash
make test-int PKG_NAME=instance PARALLEL=8
```

An error like `Error: timed-out waiting for Linode instance 1234567 to reach status running: Error waiting for Instance 85130399 status running: context deadline exceeded` is relevant to the change, and others might be intermittent or irrelevant issues.

### DIFF
--- a/linode/instance/resource.go
+++ b/linode/instance/resource.go
@@ -452,38 +452,31 @@ func createResource(ctx context.Context, d *schema.ResourceData, meta interface{
 
 	targetStatus := linodego.InstanceRunning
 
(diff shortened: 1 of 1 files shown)


ezilber-akamai (approved):

lgarber-akamai (approved):
Tests are all passing on my end. Thanks for the quick fix!

PR #2127, opened 2025-10-09
Switch linode/ubuntu18.04 to linode/ubuntu24.04 in tests

lgarber-akamai:
## 📝 Description

This pull request bumps all uses of the image `linode/ubuntu18.04` to `linode/ubuntu24.04`, which is necessary because the image for Ubuntu 18.04 is no longer offered.

## ✔️ How to Test

### Integration Testing

```
make test-int PKG_NAME=nbnode
make test-int PKG_NAME=stackscript
make test-int PKG_NAME=stackscripts
```

### DIFF
--- a/linode/instance/resource_test.go
+++ b/linode/instance/resource_test.go
@@ -2982,10 +2982,19 @@ func TestAccResourceInstance_configInterfaceVPCIPv6(t *testing.T) {
 				},
 			},
 			{
-				ResourceName:            resName,
-				ImportState:             true,
-				ImportStateVerify:       true,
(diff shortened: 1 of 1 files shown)


linode/stackscripts/tmpl/data_substring.gotf
```diff
@@ -4,7 +4,7 @@ resource "linode_stackscript" "stackscript" {
     label = "{{.Label}}"
     script = <<EOF
 {{.Script}}EOF
-	images = ["linode/ubuntu18.04", "linode/ubuntu16.04lts"]
+	images = ["linode/ubuntu24.04", "linode/ubuntu16.04lts"]
```
zliang-akamai: Should we also update the 16?

PR #2169, opened 2025-11-13
Add FirewallID to NodePool

rammanoj:
## 📝 Description

- Add `firewall_id` to NodePool.

## ✔️ How to Test

- Run make build. It generates a terraform-provider-linode.
- Move the binary into ~/.terraform.d/plugins/local/lab/linode/99.0.0/darwin_arm64
- Set LINODE_API_VERSION to v4beta.
- Create 2 firewalls (firewall-1 and firewall-2)
- Create a LKE-E Cluster and use the below config to create a nodePool with a firewall-1 ID.
```
terraform {
  required_providers {
    linode = {
      source  = "local/lab/linode"
      version = "~> 99.0.0"
    }
  }
}

provider "linode" {
  token = "<token>"
}

# New node pool managed as a separate resource
resource "linode_lke_node_pool" "gpu_pool" {
  cluster_id = "<cluster-id>"
  type       = "g6-dedicated-2" # example; use any valid Linode type

  node_count  = 4
  label       = "tmp-c"
  firewall_id = 3476976
}
```
- After applying, ensure nodes are attached to firewall-1
- Change the firewall ID to firewall-2 and apply. Ensure that the change is reflected in cloud-manager UI 
- Use the below config to perform same set of operation but using the `linode_lke_cluster` resource
```
terraform {
  required_providers {
    linode = {
      source  = "local/lab/linode"
      version = "~> 99.0.0"
    }
  }
}

provider "linode" {
  token = "<token>"
}

  label       = "my-cluster"
  k8s_version = "v1.31.9+lke7"
  region      = "us-ord"
  tags        = ["prod"]
  tier        = "enterprise"
  # update_strategy = "on_recycle"

  pool {
    type        = "g6-dedicated-2"
    count       = 3
    label       = "test-pool"
    firewall_id = 3476976
    # tags        = ["hello", "world"]
  }

  pool {
    type        = "g6-dedicated-2"
    count       = 3
  }
}
```

## 📷 Preview

### DIFF
--- a/linode/helper/boolplanmodifiers/usestateforunknownif.go
+++ b/linode/helper/boolplanmodifiers/usestateforunknownif.go
@@ -3,6 +3,7 @@ package boolplanmodifiers
 import (
 	"context"
 
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 )
 
@@ -10,24 +11,33 @@ import (
 // in place of unknown values in plans if its value is not null.
 func UseStateForUnknownIfNotNull() planmodifier.Bool {
 	return UseStateForUnknownIf(
-		func(ctx context.Context, request planmodifier.BoolRequest) bool {
-			return !request.StateValue.IsNull()
+		func(ctx context.Context, request planmodifier.BoolRequest, resp *UseStateForUnknownIfFuncResponse) {
+			resp.UseState = !request.StateValue.IsNull()
 		},
 	)
 }
 
-type planModifierCondition func(context.Context, planmodifier.BoolRequest) bool
+type UseStateForUnknownIfFunc func(context.Context, planmodifier.BoolRequest, *UseStateForUnknownIfFuncResponse)
+
+type UseStateForUnknownIfFuncResponse struct {
+	// Diagnostics report errors or warnings related to this logic. An empty
+	// or unset slice indicates success, with no warnings or errors generated.
+	Diagnostics diag.Diagnostics
+
+	// UseState should be enabled if conditions are met
+	UseState bool
+}
 
 // UseStateForUnknownIf returns a plan modifier that will only use the state value
 // in place of an unknown value if the given condition is met.
-func UseStateForUnknownIf(condition planModifierCondition) planmodifier.Bool {
+func UseStateForUnknownIf(condition UseStateForUnknownIfFunc) planmodifier.Bool {
 	return useStateForUnknownIfModifier{
 		conditionFunc: condition,
 	}
 }
 
 type useStateForUnknownIfModifier struct {
-	conditionFunc planModifierCondition
+	conditionFunc UseStateForUnknownIfFunc
 }
 
 func (m useStateForUnknownIfModifier) Description(_ context.Context) string {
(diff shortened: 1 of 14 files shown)


docs/resources/lke_cluster.md
```diff
@@ -187,6 +205,8 @@ The following arguments are supported in the `pool` specification block:
 
 * `label` - (Optional) A label for the Node Pool. If not provided, it defaults to empty string.
 
+* `firewall_id` - (Optional) The ID of the firewall to associate with this node pool. If not provided, default firewall will be associated.
```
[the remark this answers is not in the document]
rammanoj: incorrect suggestion

docs/resources/lke_node_pool.md
```diff
@@ -104,6 +115,8 @@ The following arguments are supported:
 
 * `label` - (Optional) A label for the Node Pool. If not provided, it defaults to empty string.
 
+* `firewall_id` - (Optional) The ID of the firewall to associate with this node pool. If not provided, default firewall will be associated.
```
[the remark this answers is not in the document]
rammanoj: incorrect suggestion

ezilber-akamai (commented):
In testing locally, updating the `firewall_id` from one firewall to another does not seem to propagate the changes down to the nodes in the pool. It does update the `firewall_id` of the pool itself. Perhaps some additional logic is needed in the update to propagate the change down? It's also possible that this is an API issue.

rammanoj:
> In testing locally, updating the `firewall_id` from one firewall to another does not seem to propagate the changes down to the nodes in the pool. It does update the `firewall_id` of the pool itself. Perhaps some additional logic is needed in the update to propagate the change down? It's also possible that this is an API issue.

Right now, the firewall_id on the nodes are updated once the nodes recycle. However, this would be modified to perform in-line update in future releases.

ezilber-akamai:
> > In testing locally, updating the `firewall_id` from one firewall to another does not seem to propagate the changes down to the nodes in the pool. It does update the `firewall_id` of the pool itself. Perhaps some additional logic is needed in the update to propagate the change down? It's also possible that this is an API issue.
> 
> Right now, the firewall_id on the nodes are updated once the nodes recycle. However, this would be modified to perform in-line update in future releases.

I tried manually recycling the nodes after updating the firewall id of the node pool and still, the newly recycled node pointed at the original firewall id. Any idea what could be going on here?

docs/resources/lke_cluster.md
```diff
@@ -118,6 +118,24 @@ resource "linode_lke_cluster" "my-cluster" {
...
+    label       = "my-cluster"
+    k8s_version = "1.32"
+    region      = "us-central"
+    tags        = ["prod"]
+
+    pool {
+        type  = "g6-standard-2"
+        count = 2
+        label = "db-pool"
+        firewall_id = 12345 # id of the firewall to associate with this node pool
```
ezilber-akamai:
```suggestion
        firewall_id = 12345
```

nit: Can you remove the comment here? We don't typically add comments to the example usage blocks.

linode/lke/framework_resource_test.go
```diff
@@ -392,6 +392,7 @@ func TestAccResourceLKECluster_poolUpdates(t *testing.T) {
 				{
 					Config: tmpl.LabelledPools(t, clusterName, k8sVersionLatest, testRegion, "test"),
 					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr(resourceClusterName, "pool.0.firewall_id", "12345"),
```
ezilber-akamai: This assertion fails when running the test. It gets a firewall_id of 0 when I run the test locally. This test is not on an enterprise cluster though so should this assertion be here?

rammanoj: I don't think this is needed as well. Probably left behind by accident. Removed it.

PR #2171, opened 2025-11-14
Add api_version=v4beta validation for linode_lke_cluster tier field

### LINKED ISSUE #1945: [Bug]: Adding tier on LKE forces replacement
### Terraform Version

v1.7.5 on darwin_arm64

### Linode Provider Version

2.41.0

### Effected Terraform Resources

Adding the `tier` on an existing `linode_lke_cluster` resource forces a replacement, even if the tier matches.

I have an existing LKE cluster (created before LKE-E release). I'm adding tier = 'standard' and it forces a replacement of the whole cluster.   

### Terraform Config Files

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Expected Behavior

Tier verified and imported into the state

### Actual Behavior

Resource force-replaced

### Steps to Reproduce

Create `linode_lke_cluster` without tier, then add tier = 'standard'.

lgarber-akamai:
## 📝 Description

This pull request adds validation to prevent an issue where attempting to set the linode_lke_cluster `tier` field without setting api_version to `v4beta` would cause the plan to propose the cluster be recreated.

This is necessary because the `tier` field is **not** populated when using `v4`, which causes even `tier = "standard"` to diff.

Resolves #1945 

## ✔️ How to Test

### Integration Testing

```
make test-int PKG_NAME=lke
```

### Manual Testing

1. In a terraform-provider-linode sandbox environment (e.g. dx-devenv), attempt to plan the following configuration:

```
provider "linode" {
  api_version = "v4"
}

  label = "tf-manual-test"
  region = "us-mia"
  tier = "standard"
  k8s_version = "1.34"

  pool {
    type = "g6-standard-1"
    count = 1
  }
}
```

2. Expect the following error:

```
Error: tier: The api_version provider argument must be set to 'v4beta' to use this field.
```

3. Remove the set api_version to `v4beta` and apply.
4. Ensure the apply finishes successfully.
5. Revert the api_version to `v4`, remove the `tier = "standard"`, and reapply.
6. Ensure the apply finishes successfully.

### DIFF
--- a/linode/databasemysqlv2/resource_test.go
+++ b/linode/databasemysqlv2/resource_test.go
@@ -937,7 +937,7 @@ func TestAccResource_noPendingUpdatesRegression(t *testing.T) {
 		PreCheck: func() { acceptance.PreCheck(t) },
 		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
 			"linode": func() (tfprotov6.ProviderServer, error) {
-				return acceptance.ProtoV6CustomProviderFactories["linode"](overriddenProvider)
+				return acceptance.ProtoV6CustomProviderFactories["linode"](overriddenProvider, nil)
 			},
 		},
 		CheckDestroy: acceptance.CheckMySQLDatabaseV2Destroy,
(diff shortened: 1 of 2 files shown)


linode/lke/framework_resource_test.go
```diff
@@ -820,3 +821,75 @@ func TestAccResourceLKECluster_acl_disabled_addresses(t *testing.T) {
...
+							}
+
+							for _, resourceEntry := range req.State.Values.RootModule.Resources {
+								if resourceEntry.Type != "linode_lke_cluster" || resourceEntry.Name != "test" {
+									continue
+								}
+
+								tier := resourceEntry.AttributeValues["tier"].(string)
+								if tier != "" {
+									log.Printf(
```
lgarber-akamai, line 898: This doesn't use `t.Logf(...)` because we don't use verbose mode in CI (to my knowledge).

zliang-akamai (approved):
LGTM!

PR #2198, opened 2025-12-17
Resource lock

zliang-akamai:
## 📝 Description

Add resource lock support for the TF provider.

## ✔️ How to Test
```bash
make test-int TEST_SUITE=lock
```

```bash
make test-int TEST_SUITE=locks
```

### DIFF
--- a/linode/locks/tmpl/data_basic.gotf
+++ b/linode/locks/tmpl/data_basic.gotf
@@ -12,6 +12,8 @@ resource "linode_lock" "test" {
   lock_type   = "{{ .LockType }}"
 }
 
-data "linode_locks" "test" {}
+data "linode_locks" "test" {
+  depends_on = [ linode_lock.test ]
+}
 
 {{ end }}


linode/locks/framework_datasource_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+	"github.com/linode/terraform-provider-linode/v3/linode/acceptance"
+	"github.com/linode/terraform-provider-linode/v3/linode/locks/tmpl"
+)
+
+const (
+	testRegion        = "us-ord"
```
lgarber-akamai: nitpick: Would it make sense to dynamically resolve the region here?

linode/locks/framework_datasource_schema.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/linode/terraform-provider-linode/v3/linode/helper"
+	"github.com/linode/terraform-provider-linode/v3/linode/helper/frameworkfilter"
+)
+
+var filterConfig = frameworkfilter.Config{
+	"id":        {APIFilterable: true, TypeFunc: helper.FilterTypeInt},
+	"lock_type": {APIFilterable: true, TypeFunc: helper.FilterTypeString},
```
lgarber-akamai, line 11: Would it make sense to support local filters on entity_id and entity_type too?

linode/locks/framework_datasource_test.go
```diff
@@ -0,0 +1,71 @@
...
+	testLocksDataName = "data.linode_locks.test"
+)
+
+func TestAccDataSourceLocks_basic(t *testing.T) {
```
lgarber-akamai, line 34: Seeing the following error on this test:
```
=== NAME  TestAccDataSourceLocks_basic
    framework_datasource_test.go:26: Step 1/1 error: Post-apply refresh state check(s) failed:
        path not found: SliceStep index locks.0 is out of range with slice length 0
        path not found: SliceStep index locks.0 is out of range with slice length 0
        path not found: SliceStep index locks.0 is out of range with slice length 0
        path not found: SliceStep index locks.0 is out of range with slice length 0
        path not found: SliceStep index locks.0 is out of range with slice length 0
        path not found: SliceStep index locks.0 is out of range with slice length 0
--- FAIL: TestAccDataSourceLocks_basic (51.73s)
```

I'm guessing the locks data source needs an explicit `depends_on = [...]` in the test config. Otherwise this looks great!

linode/locks/framework_datasource_schema.go
zliang-akamai: For the record, per our private discussion, entity_id and other fields can't be directly added here without a change in the filtering framework due to structure misalignment among linodego, API filtering key, and TF attribute. We are planning to support it in a future PR.

PR #2219, opened 2026-01-05
Add `locks` attribute to instance resource and data source

zliang-akamai:
## 📝 Description

Add support for locks on instance resource and data source.

## ✔️ How to Test
```bash
TF_ACC=1 go test -v -tags instance ./linode/instance/... -run TestAccResourceInstance_withLock
```

```bash
TF_ACC=1 go test -v -tags instance ./linode/instance/... -run TestAccDataSourceInstances_withLock
```

## State Checks Incompatibility

Due to the conflict between `RefreshState` and `ConfigStateChecks` which causes:

```
=== RUN   TestAccResourceInstance_withLock
    resource_test.go:3511: Test validation error: TestStep 2/2 validation error: TestStep ConfigStateChecks must only be specified with Config, ConfigDirectory or ConfigFile
--- FAIL: TestAccResourceInstance_withLock (0.00s)
```

I can't use the newer state checks for these tests. This issue has been reported to the upstream TF testing framework developers: https://github.com/hashicorp/terraform-plugin-testing/issues/590

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,22 +20,3 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
-
-  # extended-support/v2
-  - package-ecosystem: "gomod"
-    target-branch: extended-support/v2
-    directory: "/"
-    schedule:
-      interval: "monthly"
-    ignore:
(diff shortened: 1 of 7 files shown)


jriddle-linode (approved):
LGTM works locally!

PR #2220, opened 2026-01-05
Add `host_standby` attribute and deprecate `host_secondary` attribute

zliang-akamai:
## 📝 Description

Deprecate old attribute name as API renamed the field.

## ✔️ How to Test

```bash
make PKG_NAME="databasepostgresqlv2" TEST_CASE="TestAccResource_complex" test-int
```

```
make PKG_NAME="databasemysqlv2" TEST_CASE="TestAccResource_complex" test-int
```


vshanthe (approved):
Tests passing LGTM

PR #2235, opened 2026-01-16
Validate obj regions in obj bucket resource

zliang-akamai:
## 📝 Description

Add region validations to object storage bucket resource to prevent clusters being used as regions.

## ✔️ How to Test

```bash
make PKG_NAME="objbucket" TEST_CASE="TestAccResourceBucket_invalid_region" test-int
```


yec-akamai (approved):
Tests passed locally!

PR #2238, opened 2026-01-28
fix: Instance `image` diff during Instance import

ezilber-akamai:
## 📝 Description

Fixed an import/migration diff where linode_instance_disk.image was missing from state, causing unnecessary updates.

## ✔️ How to Test

### Integration Testing:

`make test-int TEST_SUITE="instancedisk"`

### Manual Testing:

1. In a sandbox environment (i.e. `dx-devenv`), run `make init` followed by `make apply` for the following TF Configuration:
```hcl
provider "linode" {}

  label  = "tf-disk-migration-test"
  region = "us-east"
  type   = "g6-standard-1"

  image = "linode/ubuntu22.04"
}
```
2. Ensure the instance is created successfully.
3. Run the following CLI commands and note the Disk and Config IDs:
    - `linode-cli linodes disks-list <LINODE_ID>`
    - `linode-cli linodes configs-list <LINODE_ID>`
4. Update the TF Configuration to the following:
```hcl
provider "linode" {}

  label  = "tf-disk-migration-test"
  region = "us-east"
  type   = "g6-standard-1"

  image = "linode/ubuntu22.04"
}

  label = "linode-root"
  size  = 5120

  image = linode_instance.test.image
}

resource "linode_instance_config" "cfg" {
  label     = "My Ubuntu 22.04 LTS Disk Profile"

  device {
    device_name = "sda"
    disk_id     = linode_instance_disk.boot.id
  }
}
```
5. Run the following commands and ensure they run successfully: 
    - `terraform import linode_instance_disk.boot <LINODE_ID>,<DISK_ID>`
    - `terraform import linode_instance_config.cfg <LINODE_ID>,<CONFIG_ID>`
6. Run `make plan` and ensure that no `image` diff is detected.
7. Clean up all dangling resources created for this test.

### DIFF
--- a/linode/databasemysqlv2/framework_datasource_schema.go
+++ b/linode/databasemysqlv2/framework_datasource_schema.go
@@ -72,7 +72,12 @@ var frameworkDatasourceSchema = schema.Schema{
 			Computed:    true,
 		},
 		"host_secondary": schema.StringAttribute{
-			Description: "The secondary/private host for the Managed Database.",
+			Description:        "The secondary/private host for the Managed Database.",
+			Computed:           true,
+			DeprecationMessage: "Use host_standby instead.",
+		},
+		"host_standby": schema.StringAttribute{
+			Description: "The standby host for the Managed Database.",
 			Computed:    true,
 		},
 		"members": schema.MapAttribute{
--- a/linode/databasemysqlv2/framework_models.go
+++ b/linode/databasemysqlv2/framework_models.go
@@ -37,6 +37,7 @@ type Model struct {
 	EngineID      types.String      `tfsdk:"engine_id"`
 	HostPrimary   types.String      `tfsdk:"host_primary"`
 	HostSecondary types.String      `tfsdk:"host_secondary"`
+	HostStandby   types.String      `tfsdk:"host_standby"`
 	Label         types.String      `tfsdk:"label"`
 	Members       types.Map         `tfsdk:"members"`
 	Platform      types.String      `tfsdk:"platform"`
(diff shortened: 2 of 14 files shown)


linode/instancedisk/framework_resource.go
```diff
@@ -207,7 +207,16 @@ func (r *Resource) Read(
 		return
 	}
 
+	// Only populate Image if state.Image is null and the parent instance has an image set
+	if state.Image.IsNull() {
+		instance, err := client.GetInstance(ctx, linodeID)
+		if err == nil && instance.Image != "" {
+			state.Image = types.StringValue(instance.Image)
+		}
+	}
```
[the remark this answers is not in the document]
ezilber-akamai: Fixed.

linode/instancedisk/framework_resource.go
```diff
@@ -207,7 +207,21 @@ func (r *Resource) Read(
...
+		instance, err := client.GetInstance(ctx, linodeID)
+		if err != nil {
+			tflog.Debug(ctx, "Failed to fetch parent instance for disk image fallback", map[string]any{
+				"linode_id": linodeID,
+				"error":     err.Error(),
+			})
+		} else if instance.Image != "" {
+			state.Image = types.StringValue(instance.Image)
+		}
+	}
```
zliang-akamai: Can we wrap this as a function of ResourceModel struct to make the CRUD functions shorter?

zliang-akamai (approved):

linode/instancedisk/framework_resource_schema.go
```diff
@@ -85,6 +85,7 @@ var frameworkResourceSchema = schema.Schema{
 			Optional:    true,
 			PlanModifiers: []planmodifier.String{
 				stringplanmodifier.RequiresReplace(),
+				stringplanmodifier.UseStateForUnknown(),
```
zliang-akamai, line 89: I think this is not needed because non-computed attribute will never be unknown.

ezilber-akamai: Good catch! This got left over by accident.

dawiddzhafarov: Since it won't be unknown, which scenarios does this code:
```
if plan.Image.IsUnknown() {
	plan.Image = types.StringNull()
}
``` 
cover in Create and Update methods? Won't the value be known at time of executing those methods?

ezilber-akamai: I changed the field to being Optional + Computed so these checks cover the case where the user does not specify an Image before importing a disk.

linode/instancedisk/framework_resource_schema.go
```diff
@@ -83,6 +83,7 @@ var frameworkResourceSchema = schema.Schema{
 		"image": schema.StringAttribute{
 			Description: "An Image ID to deploy the Linode Disk from.",
 			Optional:    true,
+			Computed:    true,
 			PlanModifiers: []planmodifier.String{
 				stringplanmodifier.RequiresReplace(),
 			},
```
zliang-akamai, line 90: Actually, can we add the UseStateForUnknown back? So the user won't see "known after apply" during `terraform plan`.

ezilber-akamai: Oh yeah good point, meant to do that and forgot 😆

PR #1843, opened 2025-04-11
Add TF setup action in the test workflow

zliang-akamai:
## 📝 Description

Add the Terraform setup action prior to the test run as a workaround for [a known TF test issue](https://github.com/hashicorp/terraform-plugin-testing/issues/429) that causes intermittent errors.

## ✔️ How to Test
See the integration test run result
https://github.com/linode/terraform-provider-linode/actions/runs/14396032679

The `text file busy` intermittent error should not appear anymore.

PR #2241, opened 2026-01-30
Add Terraform setup step to manual triggering integration tests workflow

zliang-akamai:
## 📝 Description

Similar to https://github.com/linode/terraform-provider-linode/pull/1843, but for the manually triggering workflow.


.github/workflows/integration_tests_pr.yml
```diff
@@ -57,6 +57,8 @@ jobs:
           fetch-depth: 0
           submodules: 'recursive'
 
+      - uses: hashicorp/setup-terraform@v3
```
mgwoj:
```suggestion
      - uses: hashicorp/setup-terraform@v4
```
Probably newer version could be used

PR #2246, opened 2026-02-03
TPT-4110 Add datasources for region vpc availability and regions vpc availability

dawiddzhafarov:
## 📝 Description

This PR adds two new datasources: `linode_region_vpc_availability` and `linode_regions_vpc_availability`. This is the result of adding coverage for those endpoints in [linodego](https://github.com/linode/linodego/pull/880), which must be merged first.

`linode_region_vpc_availability` requires providing `id` attribute, equivalent to the Region ID. Returns `available` (bool) and `available_ipv6_prefix_lengths` (list) attributes.

`linode_regions_vpc_availability` does not require any attributes as input. Returns the same data as above (along `id` attribute) for all the regions, inside the `regions_vpc_availability` attribute.

## ✔️ How to Test

To run integration tests:

* `make PKG_NAME="regionvpcavailability" TEST_CASE="TestAccDataSourceRegionVPCAvailability_basic" test-int`
* `make PKG_NAME="regionsvpcavailability" TEST_CASE="TestAccDataSourceRegionsVPCAvailability_basic" test-int`

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -49,7 +49,7 @@ jobs:
         with:
           go-version: '1.x'
 
-      - uses: hashicorp/setup-terraform@v3
+      - uses: hashicorp/setup-terraform@v4
 
       - name: Install Dependencies
         run: make deps


linode/regionsvpcavailability/datasource_test.go
```diff
@@ -0,0 +1,32 @@
...
+		ProtoV6ProviderFactories: acceptance.ProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: tmpl.DataBasic(t),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(resourceName, "id"),
+					acceptance.CheckResourceAttrGreaterThan(resourceName, "regions_vpc_availability.#", 0),
+					resource.TestCheckResourceAttrSet(resourceName, "regions_vpc_availability.0.available"),
+					resource.TestCheckResourceAttrSet(resourceName, "regions_vpc_availability.0.id"),
+					resource.TestCheckResourceAttr(resourceName, "regions_vpc_availability.0.available_ipv6_prefix_lengths.#", "0"),
```
mgwoj: It would be good to check this array populated

linode/regionvpcavailability/datasource_test.go
```diff
@@ -0,0 +1,36 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acceptance.PreCheck(t) },
+		ProtoV6ProviderFactories: acceptance.ProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: tmpl.DataBasic(t, regionID),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(resourceName, "id"),
+					resource.TestCheckResourceAttrSet(resourceName, "available"),
+					resource.TestCheckResourceAttr(resourceName, "available_ipv6_prefix_lengths.#", "0"),
```
mgwoj: It would be good to check if this array is populated

dawiddzhafarov: Updated test data and checks to the latest values from api.

linode/regionsvpcavailability/datasource_test.go
dawiddzhafarov: Updated test data and checks to the latest values from api.

PR #2253, opened 2026-02-06
Support resetting `root_pass` of `linode_instance` without forcing recreation

ezilber-akamai:
## 📝 Description

Make `root_pass` updatable in the `linode_instance` resource by supporting the [reset a linode's root password endpoint](https://techdocs.akamai.com/linode-api/reference/post-reset-linode-password).

## ✔️ How to Test

`make test-int PKG_NAME="instance" TEST_CASE="TestAccResourceInstance_resetRootPass"`

### DIFF
--- a/linode/instance/resource.go
+++ b/linode/instance/resource.go
@@ -843,8 +843,17 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
 
 		// Power off if not already offline
 		if settledStatus != linodego.InstanceOffline {
-			if err := ShutdownInstanceForVPCInterfaceUpdate(
-				ctx, &client, skipImplicitReboots, id, helper.GetDeadlineSeconds(ctx, d),
+			offlineReason := "offline-required update"
+			if pendingRootPassReset && pendingInterfaceUpdate {
+				offlineReason = "root password reset and VPC interface update"
+			} else if pendingRootPassReset {
+				offlineReason = "root password reset"
+			} else if pendingInterfaceUpdate {
+				offlineReason = "VPC interface update"
+			}
+
+			if err := ShutdownInstanceForOfflineOperation(
+				ctx, &client, skipImplicitReboots, id, helper.GetDeadlineSeconds(ctx, d), offlineReason,
 			); err != nil {
 				return diag.FromErr(err)
 			}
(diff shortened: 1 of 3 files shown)


linode/instance/resource.go
```diff
@@ -582,6 +582,54 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+				0,
+				helper.GetDeadlineSeconds(ctx, d),
+			); err != nil {
+				return diag.Errorf("failed to boot instance after root_pass update: %s", err)
+			}
+		}
+
+		instance, err = client.GetInstance(ctx, id)
+		if err != nil {
+			return diag.Errorf("Error fetching data about the current linode: %s", err)
```
zliang-akamai: Everything looks good! But can we wrap it into a function?

ezilber-akamai: Sure!

linode/instance/schema_resource.go
```diff
@@ -464,7 +464,6 @@ var resourceSchema = map[string]*schema.Schema{
 		Description: "The password that will be initially assigned to the 'root' user account.",
 		Sensitive:   true,
 		Optional:    true,
-		ForceNew:    true,
```
mgwoj, line 467: I do not see a check in the `TestAccResourceInstance_resetRootPass` that the instance wasn't recreated. Should it be added?

ezilber-akamai: I think removing `ForceNew` should take care of that. It may still be recreated as a part of other updates but it shouldn't be when only the root password is updated.

linode/instance/resource.go
```diff
@@ -582,6 +582,23 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
+			d,
+			client,
+			instance,
+			skipImplicitReboots,
+		)
+
+		if diags != nil {
+			return diags
+		}
+	}
```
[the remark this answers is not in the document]
zliang-akamai: I think this is a valid suggestion to consider. Is there a way to combine multiple reboot?

zliang-akamai (approved):

linode/instance/resource.go
```diff
@@ -776,58 +781,106 @@ func updateResource(ctx context.Context, d *schema.ResourceData, meta any) diag.
...
 
-		if powerOffRequired {
+		// Power off if not already offline
+		if settledStatus != linodego.InstanceOffline {
 			if err := ShutdownInstanceForVPCInterfaceUpdate(
 				ctx, &client, skipImplicitReboots, id, helper.GetDeadlineSeconds(ctx, d),
 			); err != nil {
 				return diag.FromErr(err)
 			}
 		}
```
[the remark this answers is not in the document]
ezilber-akamai: Fixed

PR #2257, opened 2026-02-11
TPT-4163: Remove build tags

psnoch-akamai:
## 📝 Description

TPT-4163 Verify and remove build tags on terraform test files


tools/tools.go
```diff
@@ -1,4 +1,3 @@
-//go:build tools
 // +build tools
```
psnoch-akamai: Do we need the second line or should I remove it as well?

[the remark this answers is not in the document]
mgwoj: Please follow this recommendation

mgwoj: IMHO restore first one and remove the second one

mgwoj (changes requested):

mgwoj (changes requested):
Please resolve merge conflict and follow suggestion related to running unit tests

PR #2263, opened 2026-02-16
Deprecate `xen` from SDKs and Tools

mgwoj:
## 📝 Description

The `xen` field has been removed from the kernel API response. We can deprecate this field from SDKs and tools. https://www.linode.com/docs/api/linode-instances/#kernel-view 

## ✔️ How to Test

No changes to the tests.

### DIFF
--- a/docs/data-sources/kernel.md
+++ b/docs/data-sources/kernel.md
@@ -41,4 +41,4 @@ The Linode Kernel resource exports the following attributes:
 
 * `version` - Linux Kernel version
 
-* `xen` - (Deprecated) If this Kernel is suitable for Xen Linodes.
+* `xen` - (Deprecated) This field is no longer populated by the Linode API and will not contain a meaningful value; it is kept only for backward compatibility and should not be relied upon.


linode/kernel/framework_models.go
```diff
@@ -17,10 +17,11 @@ type DataSourceModel struct {
 	Label        types.String      `tfsdk:"label"`
 	PVOPS        types.Bool        `tfsdk:"pvops"`
 	Version      types.String      `tfsdk:"version"`
-	XEN          types.Bool        `tfsdk:"xen"`
+	// XEN is deprecated. Deprecated: this field is no longer populated by the API, but it is included here for backward compatibility with existing implementations.
+	XEN types.Bool `tfsdk:"xen"`
 }
 
-func (data *DataSourceModel) ParseKernel(ctx context.Context, kernel *linodego.LinodeKernel) {
+func (data *DataSourceModel) ParseKernel(_ context.Context, kernel *linodego.LinodeKernel) {
```
dawiddzhafarov, line 24: Maybe remove `context.Context` entirely?

yec-akamai: SGTM. Just need to make sure all the places using this function are updated. It's possible not only used in this data source

zliang-akamai (approved):

PR #2275, opened 2026-02-28
Set APIFilterable to false for image type attribute to workaround API bug

### LINKED ISSUE #2178: [Feature]: Allow to filter images by "type" field
### Description

Hello,

When the tenant uses:
1/ automation 
2/ short-lived resource

We end up with a lot of images created from deleted instances.
Which result to many pages of images -- we have about 20 right now -- and behind the scene triggers HTTP/429 response and often lead to 2+ minutes to retrieve a few images.

Could you add the support of "type" field in the provided so we could set it to manual and improve performance?

Thanks.

### New or Affected Terraform Resources

linode_images

### Potential Terraform Configuration

```HCL
data "linode_images" "aggr_images" {

  filter {
    name = "tags"
    values = ["aggregator"]
  }

  filter {
    name = "is_public"
    values = ["false"]
  }

  filter {
    name = "type"
    values = ["manual"] # We don't need image created automatically from a deleted compute instance
  }

}
```

zliang-akamai:
## 📝 Description

Set APIFilterable to false for image type attribute to workaround API bug

## ✔️ How to Test

Verify this works:

```hcl
data "linode_images" "aggr_images" {
  filter {
    name   = "type"
    values = ["manual"] # We don't need image created automatically from a deleted compute instance
  }

}

output "images" {
  value = data.linode_images.aggr_images.images
}
```

### DIFF
--- a/linode/images/framework_schema_datasource.go
+++ b/linode/images/framework_schema_datasource.go
@@ -11,11 +11,12 @@ var filterConfig = frameworkfilter.Config{
 	"is_public":  {APIFilterable: true, TypeFunc: frameworkfilter.FilterTypeBool},
 	"label":      {APIFilterable: true, TypeFunc: frameworkfilter.FilterTypeString},
 	"size":       {APIFilterable: true, TypeFunc: frameworkfilter.FilterTypeInt},
-	"type":       {APIFilterable: true, TypeFunc: frameworkfilter.FilterTypeString},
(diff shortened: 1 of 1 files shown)


dawiddzhafarov (commented):
LGTM and verified it works as expected

dawiddzhafarov (approved):

psnoch-akamai (approved):

lgarber-akamai (approved):
Tests are passing on my end, thanks for fixing this!

PR #2288, opened 2026-03-03
Add VPC dual stack support to `instance_networking` data source

zliang-akamai:
## 📝 Description

Add VPC dual stack support to `instance_networking` data source.

## ✔️ How to Test

```bash
make PKG_NAME="instancenetworking" TEST_CASE="TestAccDataSourceInstanceNetworking_vpcDualStack" test-int
```


linode/instancenetworking/framework_datasource_test.go
```diff
@@ -113,3 +116,39 @@ func TestAccDataSourceInstanceNetworking_basicwithReseved(t *testing.T) {
...
+					// IPv6 VPC IPs are present
+					statecheck.ExpectKnownValue(testInstanceNetworkResName,
+						tfjsonpath.New("ipv6").AtSliceIndex(0).AtMapKey("vpc"),
+						knownvalue.NotNull()),
+					statecheck.ExpectKnownValue(testInstanceNetworkResName,
+						tfjsonpath.New("ipv6").AtSliceIndex(0).AtMapKey("vpc").AtSliceIndex(0).AtMapKey("ipv6_range"),
+						knownvalue.NotNull()),
+					statecheck.ExpectKnownValue(testInstanceNetworkResName,
+						tfjsonpath.New("ipv6").AtSliceIndex(0).AtMapKey("vpc").AtSliceIndex(0).AtMapKey("ipv6_addresses"),
+						knownvalue.NotNull()),
```
mgwoj, line 149: is it possible to check for exact values?

linode/instancenetworking/framework_datasource_test.go
```diff
@@ -113,3 +116,39 @@ func TestAccDataSourceInstanceNetworking_basicwithReseved(t *testing.T) {
 }
 
 // TODO (Linode Interfaces): Add test for new interface_id field.
```
mgwoj, line 118: is it still needed?

linode/instancenetworking/framework_datasource_test.go
zliang-akamai: I don't think so because the IPv6 address is auto assigned

linode/instancenetworking/framework_datasource_test.go
zliang-akamai: Probably.. let me add it in another PR

PR #2289, opened 2026-03-03
Fix diff when updating lke node pool resource

dawiddzhafarov:
## 📝 Description

PR fixes diff output when updating other attributes than `nodes` and `disk_encryption`, which were marked to be deleted and changed, respectively.

## ✔️ How to Test

1. Apply this config
```
resource "linode_lke_cluster" "nodepool_test_cluster" {
  label       = "test-nodepool-cluster"
  region      = "nl-ams"
  k8s_version = "1.33"
  tags        = ["nodepool_test_cluster"]
  external_pool_tags  = ["external"]

  pool {
      type  = "g6-standard-1"
      count = 1
  }
}

resource "linode_lke_node_pool" "foobar" {
    cluster_id = linode_lke_cluster.nodepool_test_cluster.id
    autoscaler {
        min = 1
        max = 3
    }
    label = "test-nodepool"
    type  = "g6-standard-1"
}
```
2. Update `label` in `linode_lke_node_pool` resource:
```
label = "test-nodepool-update"
```
3. Run plan/apply and verify only the requested changes are visible. It should similar to this:
```
# linode_lke_node_pool.foobar will be updated in-place
~ resource "linode_lke_node_pool" "foobar" {
        id              = "123"
      ~ label           = "ddzh-test-nodepool" -> "ddzh-test-nodepool-update"
        tags            = []

        # (1 unchanged block hidden)
    }
```


linode/lkenodepool/framework_resource_schema.go
```diff
@@ -71,6 +72,9 @@ var resourceSchema = schema.Schema{
 		"disk_encryption": schema.StringAttribute{
 			Description: "The disk encryption policy for nodes in this pool.",
 			Computed:    true,
+			PlanModifiers: []planmodifier.String{
```
mgwoj, line 75: No additional tests needed?

mgwoj (approved):

linode/lkenodepool/framework_resource_schema.go
dawiddzhafarov: I don't think they're needed, as they are very low-level comparing to current integration tests. Manual check and existing tests passing should be enough in my opinion.

PR #2292, opened 2026-03-04
Enhance linode_lock resource to support additional entity types

zliang-akamai:
## 📝 Description

Add more supported entity types for the lock resource.

## ✔️ How to Test
```bash
make PKG_NAME="lock" test-int
```

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -49,7 +49,7 @@ jobs:
         with:
           go-version: '1.x'
 
-      - uses: hashicorp/setup-terraform@v3
+      - uses: hashicorp/setup-terraform@v4
 
       - name: Install Dependencies
         run: make deps
--- a/docs/data-sources/monitor_alert_definition.md
+++ b/docs/data-sources/monitor_alert_definition.md
@@ -1,0 +1,98 @@
+---
+page_title: "Linode: linode_monitor_alert_definition"
+description: |-
+Get Monitor Alert Definition.
+---
+
+# linode\_monitor\_alert\_definition
+
+Retrieves Monitor Alert Definitions.
+For more information, see the [Linode APIv4 docs](https://techdocs.akamai.com/linode-api/reference/get-alert-definition). (**Note: v4beta only.**)
+
+## Example Usage
+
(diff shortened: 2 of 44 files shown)


linode/lock/framework_resource_test.go
```diff
@@ -98,6 +111,47 @@ func TestAccResourceLock_withSubresources(t *testing.T) {
...
+					statecheck.ExpectKnownValue(lockName,
+						tfjsonpath.New("entity_url"), knownvalue.NotNull(),
+					),
+				},
+			},
+			{
+				ResourceName:      lockName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
dawiddzhafarov, line 150: Don't know if it's worth the effort but we could have a step that tries to delete the locked resource and expect an error if that's doable. What do you think?

linode/lock/tmpl/nodebalancer_basic.gotf
```diff
@@ -0,0 +1,14 @@
+{{ define "lock_nodebalancer_basic" }}
```
dawiddzhafarov, line 1: General question: We have tests for linode and nodebalncer, should we have for the rest with assumption of adding the delete step that would result in an error? Or to combine some resources (linode+volume) and test the locks at the same time to limit the number of test cases?
Please don't implement it yet, just let me know what you think.

linode/lock/framework_resource_test.go
zliang-akamai: As discussed in DM, Terraform put the lock and locked resource into the same test config for testing so it's not easy to be tested for the deletion error.

linode/lock/framework_resource_test.go
```diff
@@ -27,6 +33,13 @@ func init() {
 	}
 
 	testRegion = region
+
+	nodeBalancerRegion, err := acceptance.GetRandomRegionWithCaps([]string{"nodebalancers"}, "core")
```
mawilk90: One small thing @zliang-akamai : 
Shouldn't we use capabilities (`linodego.CapabilityNodeBalancers`) instead of hardcoded values?

PR #2310, opened 2026-03-31
TPT-4329: Deprecate vpcs for nodebalancer create in all projects and support backend vpcs

dawiddzhafarov:
## 📝 Description

This PR deprecates `vpcs` in favor of `backend_vpcs` attribute.

## ✔️ How to Test

```
make TEST_SUITE="nb" test-int
```


linode/nb/tmpl/template.go
mawilk90: Does it make sense to change it to `BackendVPC` right away as we already use `backend_vpcs` in `nodebalancer_vpc` template and we have a dedicated `VPCDeprecated` function? If so then to rename the template into backend_vpc.gotf ?

linode/nb/tmpl/backend_vpc.gotf
```diff
@@ -0,0 +1,41 @@
...
+resource "linode_nodebalancer" "test" {
+    label = "{{ .Label }}"
+    region = "{{ .Region }}"
+    firewall_id = linode_firewall.e2e_test_firewall.id
+
+    backend_vpcs = [
+        {
+            subnet_id = linode_vpc_subnet.test.id
+            ipv4_range = "10.0.0.4/30"
+            ipv6_range = "auto"
```
zliang-akamai: `ipv6_range` is not defined in `frameworkResourceSchemaVPCs`. Terraform doesn't validate the nested object, which is a known issue https://github.com/hashicorp/terraform/issues/33570

mawilk90 (approved):

linode/nb/resource_test.go
```diff
@@ -364,7 +409,7 @@ func TestAccResourceNodeBalancer_frontendVPC(t *testing.T) {
 				ResourceName:            resName,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"created", "updated", "firewall_id"}, // Ignore strict comparison for these attributes
+				ImportStateVerifyIgnore: []string{"created", "updated", "firewall_id", "backend_vpcs"}, // Ignore strict comparison for these attributes
```
zliang-akamai, line 412: It can't handle `backend_vpcs` in the import?

zliang-akamai (approved):
Overall looks good!

linode/nb/resource_test.go
zliang-akamai: In my local test run, this doesn't seem to be an issue

zliang-akamai:
```suggestion
				ImportStateVerifyIgnore: []string{"created", "updated", "firewall_id"}, // Ignore strict comparison for these attributes
```

zliang-akamai:
We may also need this to fix state inconsistent issues which can happen sometimes:
```diff
diff --git a/linode/nb/framework_models.go b/linode/nb/framework_models.go
index 7058c6eb..d5960f7a 100644
--- a/linode/nb/framework_models.go
+++ b/linode/nb/framework_models.go
@@ -114,7 +114,9 @@ func (data *NodeBalancerModel) Flatten(
                return diags
        }
 
-       data.Firewalls = *fws
+       if fws != nil {
+               data.Firewalls = helper.KeepOrUpdateValue(data.Firewalls, *fws, preserveKnown)
+       }
 
        // Filter out backend VPC configs, matching `vpcs` attribute.
        backendVPCConfigs := filterVPCConfigsByPurpose(vpcConfigs, linodego.NodeBalancerVPCConfigPurposeBackend)
```

Or alternatively we can remove the `listplanmodifier.UseStateForUnknown()` plan modifier on the `firewalls` attribute.

linode/nb/resource_test.go
dawiddzhafarov: When removed, Im getting: 
```
NAME  TestAccResourceNodeBalancer_frontendVPC
    resource_test.go:351: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        -       "backend_vpcs.#":            "1",
        -       "backend_vpcs.0.%":          "3",
        -       "backend_vpcs.0.ipv4_range": "10.0.0.4/30",
        -       "backend_vpcs.0.subnet_id":  "573694",
          }
--- PASS: TestAccDataSourceNodeBalancer_frontendVPC (61.75s)
```

The address might be changed by the API so we don't know what the user has put in originally.

yec-akamai (approved):

zliang-akamai (approved):

linode/nb/framework_models.go
```diff
@@ -165,7 +170,12 @@ func (data *NodeBalancerModel) Flatten(
 
 	// TODO: Make use of new KeepOrUpdate helpers once Linode Interfaces has been merged.
 	// In the meantime, enabling preserveKnown will break the diff logic for computed fields.
-	data.VPCs = helper.KeepOrUpdateValue(data.VPCs, backendVPCs, false)
+	if !data.BackendVPCs.IsNull() {
+		data.BackendVPCs = helper.KeepOrUpdateValue(data.BackendVPCs, backendVPCs, false)
+	}
+	if !data.VPCs.IsNull() {
+		data.VPCs = helper.KeepOrUpdateValue(data.VPCs, backendVPCs, false)
+	}
```
zliang-akamai, line 178: Why did we need the if checks here?

mawilk90: Hi @zliang-akamai , I am not 100% sure of this, but let me try to explain as Dawid is not in the project anymore:
As mentioned in lines 34-35 VPCs attribute is deprecated in favor of BackendVPCs. However for now these both attributes work and their behavior is the same so the user can choose any of them to get the same result - I guess it may be a reason of the IF here

zliang-akamai (changes requested):
I am working on some changes for this

zliang-akamai (approved):

zliang-akamai:
Let's move to this PR, together with other changes: https://github.com/linode/terraform-provider-linode/pull/2422

PR #2322, opened 2026-04-01
TPT-4293: terraform: Implement linode_reserved_ip_types data source

mgwoj:
## 📝 Description

Implement a new Terraform data source linode_reserved_ip_types that returns pricing and type information for reserved IPv4 addresses. This endpoint is publicly accessible and does not require authentication.

## ✔️ How to Test

```
make test-int PKG_NAME="reservediptypes" TEST_CASE="TestAccDataSourceReservedIPTypes_basic"
```


linode/reservediptypes/framework_datasource_schema.go
```diff
@@ -0,0 +1,52 @@
...
+			NestedObject: schema.NestedAttributeObject{
+				Attributes: map[string]schema.Attribute{
+					"id": schema.StringAttribute{
+						Description: "The unique ID assigned to this Reserved IP Type.",
+						Required:    true,
+					},
+					"label": schema.StringAttribute{
+						Description: "The Reserved IP Type's label.",
+						Computed:    true,
+						Optional:    true,
```
dawiddzhafarov:
```suggestion
```

linode/reservediptypes/framework_models.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// RegionPriceModel is the tfsdk-tagged struct for a region-specific price entry.
+type RegionPriceModel struct {
+	ID      types.String  `tfsdk:"id"`
+	Hourly  types.Float64 `tfsdk:"hourly"`
+	Monthly types.Float64 `tfsdk:"monthly"`
+}
+
+type DataSourceModel struct {
```
dawiddzhafarov: Does it have to be exported for now?

linode/reservediptypes/tmpl/data_basic.gotf
```diff
@@ -0,0 +1,10 @@
+{{ define "reserved_ip_types_data_basic" }}
```
dawiddzhafarov, line 1: Imo the `basic` case should not have filtering and we should expect a list in the state, and introduce the `filter` case that would narrow down the results from the `basic` output.
Probably could be done as another step in existing testing function?

dawiddzhafarov (commented):
Also, documentation is missing for this new DS

linode/reservediptypes/framework_models.go
mgwoj: No, I will fix it

linode/reservediptypes/framework_datasource_schema.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/linode/terraform-provider-linode/v3/linode/helper"
+	"github.com/linode/terraform-provider-linode/v3/linode/helper/frameworkfilter"
+)
+
+var filterConfig = frameworkfilter.Config{
+	"label": {APIFilterable: true, TypeFunc: frameworkfilter.FilterTypeString},
+}
```
lgarber-akamai, line 12: I'm seeing this error when running the tests: 

```
 Error: Non-Filterable Field
        
          with data.linode_reserved_ip_types.foobar,
          on terraform_plugin_test.tf line 13, in data "linode_reserved_ip_types" "foobar":
          13: data "linode_reserved_ip_types" "foobar" {
        
        Field "id" is not filterable.
        Filterable Fields: label
```

Look like we just need to add "id" (APIFilterable: false) to filterConfig here. Otherwise implementation looks good!

mgwoj: right, I have corrected that

linode/reservediptypes/tmpl/data_filter.gotf
```diff
@@ -0,0 +1,10 @@
+{{ define "reserved_ip_types_data_filter" }}
+
+data "linode_reserved_ip_types" "foobar" {
+    filter {
+        name   = "id"
+        values = ["reserved-ip"]
+    }
+}
```
mawilk90, line 8: When filtering by `label` is applied, then `data` contains all the pricing values returned by API. However, when `ID` is used (as above), it returns no values + `TestAccDataSourceReservedIPTypes_basic` fails on Step 2.

Image

mawilk90: Ok, now I see what is incorrect: please update the template and test to use `reserved-ipv4` instead of `reserved-ip`

mawilk90 (approved):

lgarber-akamai (approved):
Great work!

PR #2323, opened 2026-04-02
TPT-4294: terraform: Update linode_instance resource to support reserved IP assignment at creation

mgwoj:
## 📝 Description

Extend the linode_instance resource to support assigning a pre-reserved IPv4 address to the instance at creation time. Currently, reserved IP assignment after creation is handled by linode_instance_reserved_ip_assignment. This story adds support for specifying a reserved IP at the point of POST /linode/instances.

## ✔️ How to Test

```
make TEST_SUITE="instance" test-int
```


dawiddzhafarov:
Sorry but I don't quite understand those changes - what changed that it enables using reserved IP in this resource?
To me, the source code is exactly the same, the difference is that some part of it was extracted to a func and tests were added for this func so there are no changes in the behavior of the resource - or am I missing something?

mgwoj:
> Sorry but I don't quite understand those changes - what changed that it enables using reserved IP in this resource? To me, the source code is exactly the same, the difference is that some part of it was extracted to a func and tests were added for this func so there are no changes in the behavior of the resource - or am I missing something?

It is only cleanup for the functionality which was already introduced earlier.

dawiddzhafarov:
Okay, so the PR title and commit message are a bit misleading for me

dawiddzhafarov (approved):

mawilk90 (approved):

lgarber-akamai (approved):
Looks great, nice work!

PR #2324, opened 2026-04-02
TPT-3701: RBAC IAM Changes

jriddle-linode:
## 📝 Description

Adds support for the following endpoints:
- `/entities`
- `/iam/users/{username}/role-permissions`

Adds the `io_ready` field to Volume related endpoint

## ✔️ How to Test

Needs the `iam_beta` tag and `ACC_OPT_IN_TESTS` set.
```
export ACC_OPT_IN_TESTS="TestAccDataSourceIAMEntities_basic"
make PKG_NAME="iamentities" test-int
```
```
export ACC_OPT_IN_TESTS="TestAccDataSourceIAMUser_basic,TestAccResourceIAMUser_Create,TestAccResourceIAMUser_Update,TestAccResourceIAMUser_UpdateAccount"
make PKG_NAME="iamuser" test-int
```
```
make PKG_NAME="volume" test-int
```


docs/data-sources/iam_entities.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+*While the filtering system is in place currently this endpoint does not have filter support for any field.*
+
+* [`filter`](#filter) - (Optional) A set of filters used to select Linode users that meet certain requirements.
+
+* `order_by` - (Optional) The attribute to order the results by. See the [Filterable Fields section](#filterable-fields) for a list of valid fields.
```
dawiddzhafarov, line 25: Is the "See the.." part applicable to this field (`order_by`)? I feel like it should be appended to description for `filter` arg

linode/iamentities/framework_datasource.go
```diff
@@ -0,0 +1,72 @@
...
+	client := d.Meta.Client
+	var data ListEntitiesModel
+	diags := resp.Diagnostics
+
+	diags.Append(req.Config.Get(ctx, &data)...)
+	if diags.HasError() {
+		return
+	}
+
+	id, diag := filterConfig.GenerateID(data.Filters)
```
dawiddzhafarov, line 43: Since there are no filters currently, will the ID be correctly generated?

linode/iamentities/framework_datasource_test.go
```diff
@@ -0,0 +1,81 @@
...
+	// entities doesn't support filtering at the moment so i'm only
+	// checking type incase test account has multiple volumes
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acceptance.PreCheck(t) },
+		ProtoV6ProviderFactories: acceptance.ProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: tmpl.DataBasic(t, label, testRegion),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(resName, "entities.0.id"),
```
dawiddzhafarov, line 75: Should we use `statecheck` pkg for new code?

linode/iamentities/tmpl/basic.gotf
dawiddzhafarov, line 1: Is this file needed?

linode/iamentities/tmpl/update.gotf
dawiddzhafarov, line 1: Is this file used in those tests?

linode/iamentities/tmpl/data_basic.gotf
```diff
@@ -0,0 +1,16 @@
+{{ define "iam_entities_data_basic" }}
+
+resource "linode_volume" "test_volume" {
+    label = "{{.VolumeLabel}}"
+    region = "{{.Region}}"
+    tags = ["tf_test"]
+}
+
+data "linode_iam_entities" "test_iam_entities" {
+    filter {
```
dawiddzhafarov, line 10: Since filters don't work now, maybe the basic case shouldn't filter? And once it works we could add a test case for that?

linode/iamuser/framework_datasource.go
```diff
@@ -0,0 +1,74 @@
...
+
+	client := d.Meta.Client
+	var data IAMUserDataSourceModel
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if data.Username.ValueString() == "" {
```
dawiddzhafarov: Can we validate that on schema level? (i.e., `stringvalidator.LenghtAtLeast(1)`

linode/iamuser/framework_datasource.go
```diff
@@ -0,0 +1,74 @@
...
+	perms, err := client.GetUserRolePermissions(ctx, data.Username.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Error listing permissions: %s", data.Username.ValueString()),
+			err.Error(),
+		)
+		return
+	}
+
+	if perms == nil {
```
dawiddzhafarov, line 53: I wonder if having no permissions should be treated as an error?
In the end, it shows lack of something and that does not mean something went wrong, so maybe empty list would be still ok
What do you think?

linode/iamuser/framework_resource_test.go
```diff
@@ -0,0 +1,147 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resName, "account_access.0", "account_event_viewer"),
+					resource.TestCheckResourceAttr(resName, "entity_access.0.type", "volume"),
+					resource.TestCheckResourceAttr(resName, "entity_access.0.roles.0", "volume_admin"),
+				),
+			},
+			{
+				Config: tmpl.Update(t, volumeName, testRegion, username, email, "volume_viewer", true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resName, "account_access.0", "account_event_viewer"),
```
dawiddzhafarov, line 140: Is it possible to have a test case that updates account_access? (or use this attribute at all?)

linode/volume/framework_datasource_test.go
```diff
@@ -32,6 +32,7 @@ func TestAccDataSourceVolume_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "linode_id", "0"),
 					resource.TestCheckResourceAttrSet(resourceName, "created"),
 					resource.TestCheckResourceAttrSet(resourceName, "updated"),
+					resource.TestCheckResourceAttrSet(resourceName, "io_ready"),
```
dawiddzhafarov, line 35: Should there be a similar check added to the resource tests if it exists?

ezilber-akamai (commented):
Seeing this failure consistently when running tests locally:
```
=== NAME  TestAccResourceVolume_update
    framework_resource_test.go:115: Step 3/3 error: The non-refresh plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
        Terraform will perform the following actions:
        
          # linode_volume.foobar will be updated in-place
          ~ resource "linode_volume" "foobar" {
                id              = "14916204"
              ~ io_ready        = false -> (known after apply)
                tags            = [
                    "tf_test",
                    "tf_test_2",
                ]
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
```

docs/data-sources/iam_entities.md
jriddle-linode: Perhaps, however i took this from our other list modules so we might should update all of them as a different PR.

linode/iamentities/framework_datasource.go
jriddle-linode: From how i understand the code its just to generate a somewhat random ID for the datasource, i dont think the ID is used for anything where it can be correct or incorrect.

linode/iamentities/tmpl/basic.gotf
jriddle-linode: Whoops thanks!

linode/iamentities/tmpl/update.gotf
jriddle-linode: Ditto

linode/iamentities/tmpl/data_basic.gotf
jriddle-linode: When i ran it originally the resource step to create the volume wasn't running before the datasource step to pull it until i made the datasource step dependent on the resource.

linode/iamuser/framework_datasource.go
jriddle-linode: Because of the way IAM works perms can never be empty. At minimum it has to have some kind of Account Access. A brand new user is either created with 4 perms if its restricted or 1 perm if its not restricted.

docs/resources/iam_user.md
```diff
@@ -0,0 +1,51 @@
+---
+page_title: "Linode: linode_iam_user"
+description: |-
+  Manages IAM details about a Linode user.
+---
+
+# Resource: linode\_iam\_user
+
+Manages IAM information about a Linode user
+For more information, see the [Linode APIv4 docs](https://techdocs.akamai.com/linode-api/reference/get-iam-users-role-permissions).
```
mawilk90: Is the link correct? Shouldn't be: https://techdocs.akamai.com/linode-api/reference/put-iam-users-role-permissions ?

dawiddzhafarov (approved):

mawilk90 (approved):

PR #2341, opened 2026-05-02
TPT-4399: Make `nat_1_1` attribute as non-computed

zliang-akamai:
## 📝 Description

Making `nat_1_1` attribute as non-computed to trigger diff when the attribute is removed from the `.tf` config files.

## ✔️ How to Test
```bash
make test-int PKG_NAME=instanceconfig
```

```bash
```


mgwoj:
Should we add test exposing this scenario?

mgwoj (approved):

jriddle-linode (approved):

PR #2372, opened 2026-06-03
fix: Removed plan validation for auth fields in favor of apply validation

ezilber-akamai:
## 📝 Description

Removed init/plan phase validation for `root_pass`, `authorized_users`, and `authorized_keys` fields being set in favor of apply phase validation.

## ✔️ How to Test

`make test-int PKG_NAME="instance" TEST_CASE="TestAccResourceInstance"`
`make test-int PKG_NAME="instance" TEST_CASE="TestAccDataSourceInstance"`

### DIFF
--- a/linode/instance/helpers.go
+++ b/linode/instance/helpers.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/sha3"
 	"encoding/base64"
-	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):
Thanks for the quick fix!

yec-akamai (approved):

PR #2389, opened 2026-06-24
fix: Add retry logic for transient disk creation errors

cheeseburgermotivated:
## Problem

Disk creation can fail in development environments with two manifestations of the same underlying issue:

1. HTTP 400 with "Linode busy." error when attempting to create a disk immediately after instance creation
2. **[002] network errors** (e.g., "[002] unexpected EOF", "[002] failed to decode response body") when the API connection is reset during the busy state

Both errors indicate the instance is still provisioning and not yet ready to accept disk operations. These are transient infrastructure state issues that should be automatically retried rather than failing the Terraform operation.

## Root Cause

When a Linode instance is in a provisioning state, the API backend may:

• Return a clean 400 error with "Linode busy." message, or  
• Experience connection resets/timeouts, manifesting as **[002]-prefixed network errors**

The same condition produces different error types depending on backend load and timing. The provider previously did not retry either error type for disk creation operations.

## Solution

Added `InstanceDiskCreateBusyRetry()` to the global retry conditions in `linode/helper/retry.go`. This function:

• Retries HTTP 400 responses containing "Linode busy." on disk creation  
• Retries **[002] network errors** (including "[002] unexpected EOF" and "[002] failed to decode response body") on disk creation  
• Only applies to the disk creation endpoint pattern: `linode/instances/{id}/disks`  
• Uses response body checking for 400s to avoid retrying validation errors  
• Verifies request URL and HTTP method for [002] errors to scope retries appropriately

**Note:** Only errors with the `[002]` error code prefix are retried. Plain EOF errors without `[002]` are not retried as they may indicate different failure modes unrelated to instance provisioning state.

The retry is applied globally via the existing `ApplyAllRetryConditions()` mechanism, similar to how 500 errors are handled for other resources.

## Implementation Details

The retry condition function checks both error types:

1. For **[002] network errors**: Verifies the error message contains "[002]" prefix with "EOF" or "failed to decode response body" keywords, confirms HTTP POST method, and validates the request URL matches the disk creation endpoint pattern
2. For "Linode busy." errors: Checks for HTTP 400 status, verifies the endpoint pattern, and searches the response body for the specific error message

This dual approach handles the same transient state issue regardless of how it manifests at the HTTP layer.

## Testing

• Added comprehensive unit tests in `linode/helper/retry_test.go`:
  ◦ Validates retry on "Linode busy." 400 responses  
  ◦ Validates retry on [002] network errors with various formats  
  ◦ Ensures retries only occur on disk creation endpoints  
  ◦ Verifies no retries on other endpoints or error types  
• All existing unit tests pass  
• Manual testing confirms retries work in development environments  

## Related Patterns

This follows the existing pattern used for LKE cluster operations, which also retry EOF errors during provisioning. The LKE code has a comment noting: "Sometimes the K8S API will raise an EOF error if polling immediately after a cluster is created. We should retry accordingly."

Fixes intermittent disk creation failures in development ecosystems where instances are frequently created and modified in rapid succession.


linode/helper/retry.go
```diff
@@ -75,6 +76,56 @@ func OBJBucketDelete500Retry() func(response *resty.Response, err error) bool {
...
+	}
+
+	return func(response *resty.Response, err error) bool {
+		// Check if this is an EOF error on disk creation
+		if err != nil && strings.Contains(err.Error(), "EOF") {
+			// We need to verify this is a disk creation request
+			if response != nil && response.Request != nil {
+				requestURL, urlErr := url.ParseRequestURI(response.Request.URL)
+				if urlErr == nil && diskCreatePath.MatchString(requestURL.Path) {
+					log.Printf("[DEBUG] Retrying disk creation due to EOF error: %s", err.Error())
```
yec-akamai: nit: Can you use tflog.debug() to print the message?

yec-akamai (commented):
Works on my end, just minor things:

linode/helper/retry.go
cheeseburgermotivated: This has been done at 8161e0075851020564452f3cd0204a193d1324d3

lgarber-akamai (approved):
Seems sensible to me, should be good after merge conflicts are resolved. Thanks for the contribution!

linode/helper/retry.go
```diff
@@ -74,6 +77,113 @@ func OBJBucketDelete500Retry() func(response *http.Response, err error) bool {
...
+			// We need to verify this is a disk creation request
+			if response != nil && response.Request != nil && response.Request.URL != nil {
+				if response.Request.Method != http.MethodPost {
+					return false
+				}
+				if diskCreatePath.MatchString(response.Request.URL.Path) {
+					tflog.Debug(ctx, "Retrying disk creation due to [002] network error", map[string]any{
+						"error": err.Error(),
+						"path":  response.Request.URL.Path,
+					})
```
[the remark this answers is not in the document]
cheeseburgermotivated: no, this is scope creep, i don't want a ballooning PR, nobody else is even doing nil response handling here.

lgarber-akamai (approved):
Seems pretty stable in my local testing. Thanks again for the contribution!

yec-akamai (approved):
Thank you for the work!

PR #2392, opened 2026-06-26
TPT-4498: Migrate Terraform to use linodego v2

ezilber-akamai:
## 📝 Description

Address issues caused by breaking changes introduced in linodego v2.

## ✔️ How to Test

`make test-unit`
`make test-int`


psnoch-akamai (approved):

linode/objbucket/schema_resource.go
```diff
@@ -18,23 +18,12 @@ var resourceSchema = map[string]*schema.Schema{
...
-		Description:  "The region of the Linode Object Storage Bucket.",
-		Optional:     true,
-		Computed:     true,
-		ForceNew:     true,
-		ExactlyOneOf: []string{"region", "cluster"},
+		Type:        schema.TypeString,
+		Description: "The region of the Linode Object Storage Bucket.",
+		Optional:    true,
+		Computed:    true,
+		ForceNew:    true,
```
lgarber-akamai, line 25: Would it make sense to make this required now that `cluster` isn't available? Not sure if this will cause any weird SDKv2 issues so otherwise it seems fine to keep as-is 👍

linode/objkey/framework_resource_schema.go
```diff
@@ -81,33 +79,10 @@ var frameworkResourceSchema = schema.Schema{
...
-							),
-						},
-						PlanModifiers: []planmodifier.String{
-							stringplanmodifier.UseStateForUnknown(),
-						},
-					},
 					"region": schema.StringAttribute{
 						Description: "The region where the bucket resides.",
 						Optional:    true,
 						Computed:    true,
```
lgarber-akamai: Same as my other comment

lgarber-akamai (approved):
Works well in my local testing!

PR #2415, opened 2026-07-17
TPT-4126: Project NB Front-End IP in VPC-MTC + 40Gbps NB-MTC

mawilk90:
## 📝 Description

This PR merges support for NB Front-End IP + 40Gbps features into the dev branch for release.

Note:
1. This [PR](https://github.com/linode/linodego/pull/978) needs to be merged first and then **replace** statement should be removed from `go.mod`
2. Needs to be checked/merged first: [PR](https://github.com/linode/terraform-provider-linode/pull/2310)

## ✔️ How to Test

Run integration tests:

```
make PKG_NAME="nb" test-int
```
```
make PKG_NAME="nbconfig" test-int
```
```
make PKG_NAME="nbconfigs" test-int
```
```
make PKG_NAME="nbnode" test-int
```
```
make PKG_NAME="nbtypes" test-int 
```
```
make PKG_NAME="nbs" test-int   
```
```
make PKG_NAME="nbvpc" test-int 
```
```
make PKG_NAME="nbvpcs" test-int 
```


linode/nb/framework_resource.go
```diff
@@ -90,6 +91,16 @@ func (r *Resource) Create(
 		createOpts.VPCs = vpcs
 	}
 
+	if !data.FrontendVPCs.IsNull() {
```
lgarber-akamai, line 126: Apparently WriteOnly fields may only be available in req.Config and may not be populated in req.Plan. Not sure if this matters in practice but might be worth looking into: https://developer.hashicorp.com/terraform/plugin/framework/resources/write-only-arguments#general-concepts

linode/nbnode/tmpl/template.go
```diff
@@ -56,12 +57,13 @@ func Updates(t testing.TB, nodebalancer, region string, rootPass string) string
...
 					Label:  nodebalancer,
 					Region: region,
+					Type:   nbType,
 				},
 			},
 		})
 }
 
-func VPC(t testing.TB, nodebalancer, region string, _ string) string {
+func VPC(t testing.TB, nodebalancer, region, nbType string, _ string) string {
```
lgarber-akamai: Looks like the order of params might need to be swapped here to make `nbType` last 👍

zliang-akamai:
Do we need to get this in? https://github.com/linode/terraform-provider-linode/pull/2310

linode/nb/framework_resource.go
mawilk90: Thanks @lgarber-akamai for the comment. I think you're right and current implementation causes that `frontend_vpcs .ipv4_range` and `ipv6_range` to always be null, regardless of whether they're provided in .tf or not. The reason is that WriteOnly arguments are not accissible via `req.Plan`, `req.Config` should be used instead. I will push a fix tomorrow morning

mawilk90: If not mistaken, `req.Config` should be also used for `vpcs.ipv4_range_auto_assign`.

mawilk90:
> Do we need to get this in? #2310

Hi @zliang-akamai, I think we still need it, because it's still possible to use both `vpcs` and `backend_vpcs` specifying backend VPCs for nodebalancer.

linode/nb/framework_resource.go
zliang-akamai, line 1: Ideally, we should never use config values in CRUD functions because they may be differ than the plan

zliang-akamai: Can we undo the config usage and get my model flatten work in? That might resolve some issues related to the plan and unknown.
https://github.com/linode/terraform-provider-linode/pull/2422

zliang-akamai: We can't use WriteOnly for new stuff neither... It's not compatible with older Terraform CLI.

PR #2438, opened 2026-08-04
TPT-3516: Fixed issue with obj bucket `lifecycle_rule`

ezilber-akamai:
## 📝 Description

Fixed an issue where `lifecycle_rule` was incorrectly persisted to state after a failed apply when linode_object_storage_bucket was configured without access_key and secret_key, preventing stale state and subsequent plan failures.

## ✔️ How to Test
`make test-int PKG_NAME="objbucket"`


psnoch-akamai (approved):
LGTM, test works well

linode/objbucket/resource.go
```diff
@@ -183,7 +183,21 @@ func createResource(
...
+		// state and failures on subsequent plans.
+		if err := d.Set("lifecycle_rule", nil); err != nil {
+			diags = append(diags, diag.FromErr(err)...)
+		}
+		if err := d.Set("versioning", nil); err != nil {
+			_ = d.Set("versioning", false)
+			diags = append(diags, diag.FromErr(err)...)
+		}
+	}
+	return diags
```
zliang-akamai: I think this only fixes create, not update?

zliang-akamai: Can we make the read to fail soft when obj key is not available so it won't interrupt destroy?

Also, I think we can revert the change back to original value from the update function, like:
```go
func updateResource(
	ctx context.Context, d *schema.ResourceData, meta any,
) diag.Diagnostics {
	// when failed to get obj keys
	revertS3ManagedFields(d, diags)
	// ...
}



func revertS3ManagedFields(d *schema.ResourceData, diags diag.Diagnostics) diag.Diagnostics {
	oldLifecycle, _ := d.GetChange("lifecycle_rule")
	oldVersioning, _ := d.GetChange("versioning")

	if err := d.Set("lifecycle_rule", oldLifecycle); err != nil {
		diags = append(diags, diag.FromErr(err)...)
	}
	if err := d.Set("versioning", oldVersioning); err != nil {
		diags = append(diags, diag.FromErr(err)...)
	}
	return diags
}
```

zliang-akamai (approved):

linode/objbucket/resource.go
zliang-akamai: The current implementation also works, so this is actually optional

ezilber-akamai: I think your suggestion is cleaner, just pushed it up!

zliang-akamai (approved):

PR #2441, opened 2026-08-10
TPT-4610: Add platform attribute to linode_databases data source

zliang-akamai:
## 📝 Description

Adds the computed `platform` attribute to the `linode_databases` data source so Managed Database listings expose the back-end platform returned by the API.

Also:
- Flattens and unit-tests the new `platform` field on database models
- Migrates the databases data source integration test to `statecheck` / `knownvalue` and asserts `platform`
- Fixes the long-running test skip message to reference the correct `RUN_LONG_TESTS` env var
- Documents `platform` in the data source docs

## ✔️ How to Test

```bash
make PKG_NAME="databases" test-unit
RUN_LONG_TESTS=true make TEST_SUITE="databases" test-int
```


mawilk90 (approved):

lgarber-akamai (approved):
Looks great, nice work!

PR #2444, opened 2026-08-10
TPT-2261: Support resolving linode_sshkey data source by ID

zliang-akamai:
## 📝 Description

The `linode_sshkey` data source previously only supported lookup by `label`. This change adds support for resolving an SSH key by `id` as well, consistent with other single-resource data sources.

- Exactly one of `label` or `id` is required (plan-time `ExactlyOneOf`)
- Both attributes are Optional+Computed so the unset selector is filled from the API response
- Existing label-based configs continue to work
- `id` now exports the real numeric API ID as a string (instead of a JSON blob)
- Nested list attributes for `linode_sshkeys` were moved into that package
- Docs and acceptance tests cover label lookup, ID lookup, and not-found cases

### Unit Testing

```bash
make PKG_NAME=sshkey test-unit
make PKG_NAME=sshkeys test-unit
```

### Integration Testing

```bash
make PKG_NAME=sshkey test-int
make PKG_NAME=sshkeys test-int
```


psnoch-akamai (approved):

lgarber-akamai (approved):
Looks great, nice work!

PR #2452, opened 2026-08-19
TPT-4456: Add Linode test with RDMA interfaces

yec-akamai:
## 📝 Description

Add integration test to Linode RDMA interfaces.

## ✔️ How to Test

Unit test:
```
make test-unit PKG_NAME=instance
```

Integration test:

Please export the region to enable the test: `export LINODE_RDMA_REGION=the-region`
```
make test-int PKG_NAME=instance TEST_CASE=TestAccResourceInstance_linodeInterfacesRDMAVPC
```


linode/instance/tmpl/templates/interfaces_rdma_vpc.gotf
```diff
@@ -0,0 +1,70 @@
...
+
+resource "linode_vpc_subnet" "rdma" {
+    vpc_id = linode_vpc.rdma.id
+    label  = "{{.Label}}-rdma-subnet"
+    ipv4   = "10.0.0.0/24"
+}
+
+resource "linode_instance" "foobar" {
+    label     = "{{.Label}}"
+    type      = "g3-gpu-rtxpro6000-blackwell-rdma-8"
```
mawilk90, line 45: Shouldn't we store this value in env vars similar to region?

yec-akamai: I think it's fine since the access is well gated by the customer tag, and the plan should be communicated to user. The env var for region mostly works as a flag to enable the test.

ezilber-akamai (approved):
LGTM and passing locally. Nice work!

zliang-akamai (approved):

PR #2464, opened 2026-09-01
TPT-4456: Project RDMA Linode Interfaces

yec-akamai:
## 📝 Description

Merge project Linode RDMA interfaces into dev.

## ✔️ How to Test

Unit test:
```
make test-unit PKG_NAME=instance
```

Integration test:

Please export the region to enable the test: `export LINODE_RDMA_REGION=the-region`
```
make test-int PKG_NAME=instance TEST_CASE=TestAccResourceInstance_linodeInterfacesRDMAVPC
```


linode/linodeinterface/framework_rdma_vpc_models.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx context.Context,
+	ipv4 linodego.RDMAVPCInterfaceIPv4,
+	preserveKnown bool,
+	diags *diag.Diagnostics,
+) {
+	addresses := helper.MapSlice(
+		ipv4.Addresses,
+		func(addr linodego.RDMAVPCInterfaceIPv4Address) RDMAVPCIPv4AddressAttrModel {
+			return RDMAVPCIPv4AddressAttrModel{
+				Address: types.StringValue(addr.Address),
```
lgarber-akamai: Copilot picked up on a risk here: 
Is `auto` a valid value here and would it cause a perpetual diff if the user configured it with auto?

linode/linodeinterface/framework_resource_schema.go
```diff
@@ -409,6 +412,77 @@ var vpcInterfaceSchema = schema.SingleNestedAttribute{
...
+				"Must contain exactly one element if specified.",
+			Optional:     true,
+			Computed:     true,
+			NestedObject: configuredRDMAVPCInterfaceIPv4Address,
+			PlanModifiers: []planmodifier.List{
+				listplanmodifier.UseStateForUnknown(),
+			},
+			Validators: []validator.List{
+				listvalidator.NoNullValues(),
+				listvalidator.SizeAtMost(1),
```
lgarber-akamai: Can we use `listvalidator.SizeBetween(1, 1)` here if we're requiring exactly one address if specified? Not sure if that will cause a validation error when unspecified, so feel free to ignore if so

ezilber-akamai (approved):

linode/linodeinterface/framework_resource_schema.go
yec-akamai: Updated!

linode/linodeinterface/framework_rdma_vpc_models.go
yec-akamai: It's true that changing from an explicit ip address to `auto` will trigger an issue. I've added a semantic check to this with a custom type. Thanks for catching it!

lgarber-akamai (approved):
Looks great, nice work!

